Issue #23, opened 2013-10-14
Check for ServiceControl Updates


dannycohen:
As Opie, I would like to be notified when a new version of ServiceControl is available, and what are the enhancements provided by it.

1. When ServicePulse starts, check for the latest released version of ServiceControl by calling: http://particular.net/api/products/ServiceControl/Current
   * Note: ServiceControl version API is not yet implemented (see https://github.com/Particular/Website.Backend/issues/133) 
* Compare the version field number of the current release to the ServiceControl version installed. 
* If the current version is newer than the currently installed version, display the following dialog:
```
A newer version of ServiceControl [NewVersionNumber] is available for download at: http://particular.net/downloads.
```
```
For more information on version [NewVersionNumber] see https://github.com/Particular/ServiceControl/releases/tag/[NewVersionNumber] 
```
   * Note: for an example of a release info, see: https://github.com/Particular/NServiceBus/releases/tag/4.0.5
* Repeat the check for updates and the notification every 24 hours (since ServicePulse users tend to leave it running for long periods; note that this is just an in-memory task, and there is no need to persist this timed task; if the user closes ServicePulse, then the check and notification will be performed again on ServicePulse restart)

**Important Notes:**

   * The URI calls to check the latest release  must not delay the loading time of ServicePulse. (It must be done as a background async operation)
   * If there is no internet connection or the Particular.net site is not accessible for any reason, do not display an error message
   * The message dialog must be modeless (it must not interfere with the user's work; it is intended only to draw the user's attention; it can be implemented as a toast popup that can be hidden via a click as stated above). 
* This issue requires the addition of ServiceControl version API support in the Particular.net backend API (see https://github.com/Particular/Website.Backend/issues/133)
* This issue requires the addition of a version parameter in ServiceControl version API (see https://github.com/Particular/ServiceControl/issues/72)
* If ServicePulse and ServiceControl both have a new version, display their indicators together (see #22)

petersgiles:
this was solved in https://github.com/Particular/ServicePulse/issues/22

PR #175, opened 2015-07-02
Added Get new version! directive to dashboard

petersgiles:
added "Get new version!" notification
that looks like Team city footer on ServicePulse Dashboard


gbiellem:
@weralabaj  Do you want to review this since you're plying in the SP space?

gbiellem:
This would address #22 and #23

weralabaj:
@TheBaconPhoenix @gbiellem Should we also take into account the fact that new SP version needs to be compatible with the particular SC version to work? We have exactly that concern in [FeatureDev #272](https://github.com/Particular/FeatureDevelopment/issues/272). In that case the new SP functionality requires SC upgrade too.

Imho the simplest solution would be to upgrade all tools (SC, SP, SI) at once, not only SP. Any concerns regarding that?

weralabaj:
@TheBaconPhoenix Added a few comments, mostly questions.

gbiellem:
> Should we also take into account the fact that new SP version needs to be compatible with the particular SC version to work?

:+1: Makes sense to re-use if you're working on feature switching stuff in SP

> Imho the simplest solution would be to upgrade all tools (SC, SP, SI) at once, not only SP. Any concerns regarding that?

I'm assuming because feature switching is a pain :)

weralabaj:
@gbiellem 
> :+1: Makes sense to re-use if you're working on feature switching stuff in SP
Not really, we have hardcoded the version number that supports what we need :( "Advanced versioning" afair is supposed to be more like "is this feature supported in the sc version I currently have" rather than forcing SP to know what SC version it needs. But that was out of scope for 272.

TheBaconPhoenix:
@weralabaj @gbiellem 

> I'm not sure what 'M' stands for in restrict?
> I think we don't need 'A', as it's used only as an element.
> I'm not an Angular ninja so just a question really - why should that part be a separate directive? Seems 
>  that now we have directives only for heavily reused elements such as time display, that doesn't seem 
>  like a re-usable component.

M restrictions allow you to write your directive like so  
Here is a good reference  http://www.angularjshub.com/examples/customdirectives/restrict/
Just habit on my part I guess as I didn't want to be prescriptive about we use it. We can talk about how we want to reuse directives and settle on one. :smiley:

semverUtils is a library that @johnsimons sent me that I hacked a bit to get it to work. I think you're right about some of the logic you mentioned being superfluous and I think I can rewrite it as a generic service that suits our needs specifically. Including the notion 'is newer' vs 'is different'.

petersgiles:
@weralabaj are you happy with my latest revisions? Merge?

weralabaj:
@petergiles I'd only move the feature from directives directory into a "feature" directory, similar to what is done for failedMessages. Looks ok for me.

PR #177, opened 2015-07-16
Grouping of failed messages by exception type

petersgiles:
fix button icons
add retry group
align and make equal sized buttons


wolfbyte:
Updated the contents to make it more "Announcy". Here is the original content (because there's no history on issue bodies):

Implemented a cleaner failed message management interface. 
- Included a failed message grouping by failure type. 
- Improved consistency of button icons, aligned and made equal sized buttons and tooltips. 
- Added retry message group that will retry all messages in the same failure group
- De-cluttered the failed messages page by adding a tabbed interface to navigate between message groups and individual messages

weralabaj:
@wolfbyte do we explain somewhere (e.g. in doco) how messages are grouped? If so it would be awesome to include a link to that content in an announcement.

PR #189, opened 2015-09-10
basic state machine for animation complete

petersgiles:
Part of https://github.com/Particular/PlatformDevelopment/issues/401


petersgiles:
@johnsimons this is the workflow state diagram

basic state diagram - uml state diagram

petersgiles:
@johnsimons do you want to have a look,

I've dropped 'noisy' info notifications and added slide animations to the lists

colin-higgins:
@petersgiles 
The animation seems to work pretty well. Lots of nice cleanup in this PR too.
It seems like the commits could be squashed for a cleaner history though.
cc / @johnsimons

petersgiles:
@colin-higgins 

Thanks, I'm trying to get things a lot more consistent throughout the code base.

I understand you have a bit of angular experience under your belt I'd like some help

- Dealing with all the $scope references throughout the controllers I'd like to isolate them better
- And I think it would be better if the signalr stuff was better encapsulated maybe we could use $rootscope events to break the controller dependencies on 

Advice welcome

colin-higgins:
I have found that relying on events on the scope hierarchy can get nasty really fast.
Services are singletons though and you can do a really nice pub sub pattern with them too. 
Maybe we can spike something on it soon?

petersgiles:
@colin-higgins I'd really like that

PR #193, opened 2015-09-24
Refactor streamService to not rely on events. Remove apply call with $jquery.

colin-higgins:
The $apply call was necessary because we were using a global reference to jQuery instead of injecting and giving angular context. 

I opted to use an internal dictionary per messageType for the subscribers. The subscribe method returns an unsubscribe method, removing the need for $scope references.

### DIFF
--- a/src/ServicePulse.Host/app/js/app.js
+++ b/src/ServicePulse.Host/app/js/app.js
@@ -23,7 +23,6 @@
             $rootScope.$log = $log;
         }]);
 
-    angular.module('sc')
-    .value('$jquery', $);
+    angular.module('sc').value('$jquery', $);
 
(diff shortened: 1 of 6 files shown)


colin-higgins:
The other option is injecting $window and referencing $window.jQuery. That's a catch all though. What do you think of the refactor overall?

petersgiles:
I would change cleanupMethods to activeSubscriptions. 

Aside from that everything is an improvement...neatens things up nicely

colin-higgins:
Okay, I will change that tomorrow and merge? Or if you want to today that works too. Cheers!

petersgiles:
do it tomorrow

Issue #194, opened 2015-10-19
Error Groups Title


cjoergensen:
The title column in the Error group view could be optimized for longer exception messages, its kinda hard to tell the different exceptions apart:

![image](https://cloud.githubusercontent.com/assets/8087733/10572522/88a994e2-7649-11e5-818c-c792ea14510b.png)

The overview could be better if the column was compose over multiple lines:

[NServiceBus.ExceptionInfo.ExceptionType]
[MessageType]
[NServiceBus.ExceptionInfo.Message - without stacktrace]

Ex.:

![image](https://cloud.githubusercontent.com/assets/8087733/10572724/148be7b6-764b-11e5-8c28-d55ca77fb641.png)

petersgiles:
@cjoergensen I'll talk it over with the design guys

petersgiles:
@cjoergensen we've made an adjustment to show the whole title field. You should see this appear in the next release of Service Pulse.

cjoergensen:
@petersgiles Cool

PR #196, opened 2015-10-25
Updated ui-bootstrap and removed ellipses from group titles

petersgiles:
related to https://github.com/Particular/ServicePulse/issues/194


petersgiles:
@johnsimons here is the title fix please review

johnsimons:
:+1:

Issue #200, opened 2015-10-29
ServicePulse fails to retry messages


Bishbulb:
I have been experiencing issues where ServicePulse is not retrying messages. This was occurring with both ServicePulse 1.2.1/ServiceControl 1.6.3 and now ServicePulse 1.3.0/ServiceControl 1.7.1.

The steps to reproduce are simple:
- Select an error group
- Select a message
- Click "Retry 1 message"

Actual: The page redirects back to the list of Error Groups (no confirmation/warning/error messages are seen). From looking at the journal queue, I can clearly see that no attempts were made to retry the message. The error message remains visible in ServicePulse with no indication that a retry was attempted.

Expected: The message should have been retried and should have appeared in the journal queue.

### HTTP Request Headers
POST /api//errors/retry HTTP/1.1
Host: localhost:33333
Connection: keep-alive
Content-Length: 40
Accept: application/json, text/plain, */*
Origin: http://localhost:9090
User-Agent: Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.101 Safari/537.36
Content-Type: application/json;charset=UTF-8
Referer: http://localhost:9090/
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.8

### HTTP Request Body
["d50331f8-4bfa-24de-5e7a-cf9969195f5c"]

### HTTP Response Headers
HTTP/1.1 202 Accepted
Cache-Control: private, max-age=0
Transfer-Encoding: chunked
Content-Type: text/html
Server: Microsoft-HTTPAPI/2.0
X-Particular-Version: 1.7.1
Access-Control-Expose-Headers: ETag, Last-Modified, Link, Total-Count, X-Particular-Version
Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept
Access-Control-Allow-Methods: POST, GET, PUT, DELETE, OPTIONS, PATCH
Access-Control-Allow-Origin: *
Date: Thu, 29 Oct 2015 16:24:05 GMT

**The request URL looks suspicious to me -- api//errors/retry has a double forward slash, maybe that is causing the problem?**

petersgiles:
Thanks for reporting this, I'll try and get a hotfix out ASAP

petersgiles:
I have written a patch so you should need to worry about that backslash. I'll try and get it out today.

PR #201, opened 2015-10-30
If ServiceControl Api Uri is configured with or without a trailing slash the application will now behave properly

petersgiles:
relates to https://github.com/Particular/ServicePulse/issues/200

If there is a trailing slash on the end of the ServiceControl api uri in the config file ServicePulse would compensate for it inappropriately. Now if the config has a trailing slash or not ServicePulse will be able to handle it.

### DIFF
--- a/src/ServicePulse.Host/ServicePulse.Host.csproj
+++ b/src/ServicePulse.Host/ServicePulse.Host.csproj
@@ -253,6 +253,7 @@
     <None Include="app\css\font-awesome\fonts\fontawesome-webfont.woff" />
     <None Include="app\css\font-awesome\fonts\fontawesome-webfont.woff2" />
     <None Include="app\css\font-awesome\fonts\FontAwesome.otf" />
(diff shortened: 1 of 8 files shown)


petersgiles:
@johnsimons @colin-higgins please review

johnsimons:
@petersgiles at first it scared me the ammount of code changed to fix a missing backspash.
But I think i see the approach u took and cover all the basis.
So :+1: to merge

Just for future, is there a way to test this ?

petersgiles:
@johnsimons we need to look at a test framework perhaps karma but I've never run one up in anything other than gulp. Dunno how it would all work with team city etc.

colin-higgins:
@petersgiles @johnsimons 

I've set up javascript tests in TeamCity before, I can get that up for us soon. We'd use a headless browser like PhantomJS to run the tests. PhantomJS is just an executable that you feed some arguments too, easy to add as a step in TeamCity.

PR #205, opened 2015-11-02
WIP: QUnit & Jasmine run by Resharper

petersgiles:
Can't seem to get angular.js to play nice with the tests @colin-higgins do you have any experience with this?

Look at angular_tests.js in the test project.

I keep getting a `can't find variable: angular` message


colin-higgins:
It looks like window.module is causing the test to crash. After looking into it and trying angular.mock.module instead, I would have to assume this is caused by having these tests in a class library instead of a web project. It seems like this is breaking how jasmine typically loads the references.

I wrote the tests below to verify what is loading:

```
    //fails
    it('angular.mock.module is loaded!', function () {
        expect(typeof (angular.mock.module) == typeof (undefined)).toEqual(false);
    });

    //fails
    it('window.module is loaded!', function () {
        expect(typeof (module) == typeof (undefined)).toEqual(false);
    });

    //passes
    it('angular is loaded!', function () {
        expect(typeof (angular) == typeof (undefined)).toEqual(false);
    });

    //passes
    it('angular.mock is loaded!', function () {
        expect(typeof (angular.mock) == typeof (undefined)).toEqual(false);
    });

```

petersgiles:
@colin-higgins I just created the file and dragged the relevant .js files into it. VisualStudio does the rest.

petersgiles:
Ok I got this working

Your 'is angular there' tests really helped. the paths were wrong in _references.js VisualStudio short changed me a `../`I think that there might be more issues ahead for us there though. We might need to make them relative to the solution somehow.

Second thing was I changed the setting to use Jasmine 2.0 instead of 1.3. Again your observation that things weren't in the right context led me there

This is how I have my resharper set up

capture

Issue #215, opened 2015-11-10
Unable to disable endpoints in ServicePulse 1.3.4


petersgiles:
From Google Groups https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/M_zkgTKy15g/TL1WuQ-YBQAJ

> I'm running version 1.3.4 of ServicePulse and am not able to disable a couple of endpoints I'm not using (which, of course, causes my dashboard to indicate an endpoint error when I don't run them).  When I click the Disable button next to each of these endpoints on the Configuration page, the button changes to text that says "Update Failed:". 

> Is there a way to get these endpoints to be disabled?  If not, I'd be happy if I could even just have them no longer appear as endpoints at all (something I've tried to do, but with no success).

> Thanks,

> David Madrian

petersgiles:
I've been able to recreate similar behaviour. I should have a hotfix ready soon.

Issue #214, opened 2015-11-10
Custom checks shows as failed on dashboard even the check is ok


andreasohlund:
Note the eventlog item that the check is ok

<img width="1577" alt="sp-dashboard" src="https://cloud.githubusercontent.com/assets/125028/11063652/738f300a-87b8-11e5-8fe3-700995afd1ab.PNG">

and looking at the custom checks page its listed as ok as well

<img width="742" alt="sp-custom checls" src="https://cloud.githubusercontent.com/assets/125028/11063662/7eaa5794-87b8-11e5-8fa2-a7e8d5e76eb9.PNG">

### Versions

SC 1.7.4
SP 1.3.4

mauroservienti:
The issue is broader, it affects in the same way also the `Endpoints` indicator.

andreasohlund:
Looking at the response from http://localhost:33333/api/customchecks?status=unresolved
  

```
[
{
id: "df0acd20-2ca4-4eb8-1632-e3a141509bc6",
custom_check_id: "RetryRateCheck",
category: "MyCategory",
status: "pass",
reported_at: "2015-11-10T19:38:21.7255372Z",
originating_endpoint: {
name: "RetriesCustomCheckSample",
host_id: "fd9c16f5-8b4e-af7d-cfcb-6887a11366d4",
host: "ANDREAS2015"
}
}
]

```
It seems like there might be an issue with the filtering since as you can see the status is `pass`

andreasohlund:
Looking at https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl/CustomChecks/CustomChecksModule.cs#L69 is seems like SC is expecting `fail` to be passed instead of `unresolved`

@petersgiles @colin-higgins ideas?

andreasohlund:
@petersgiles this commit seems to be the culprit https://github.com/Particular/ServicePulse/commit/b709cf940c6ab72aaed9b4b88cebbb3138dfb519#diff-2e8e4aea7512a76953509cb663ed9ce2

@mauroservienti please review and see if this is the issue with endpoints as well

petersgiles:
I have a fix for this now

PR #217, opened 2015-11-11
Hotfix-1.3.5

petersgiles:
@andreasohlund @johnsimons I think I have this right...I did something weird with git and my pull request has more junk in it than I would like.

This fixes these 
- https://github.com/Particular/ServicePulse/issues/214
- https://github.com/Particular/ServicePulse/issues/215
- https://github.com/Particular/ServicePulse/issues/218
- https://github.com/Particular/ServicePulse/issues/220
- https://github.com/Particular/ServicePulse/issues/221
- https://github.com/Particular/ServicePulse/issues/222

If OK I'll release early tomorrow

I have put in a few bits an pieces to smoke-test some scenarios around endpoint and heartbeats. I'd like some help removing them from the build on the TeamCity because they really only are of use during development


andreasohlund:
Looks like you branch the hotfix branch of develop? (could that be why all the extra commits are here)

What commits should be included in the hotfix?

petersgiles:
@johnsimons yes for some odd reason in my tests it could come through either way. This was what was causing the endpoint update issue.

petersgiles:
@andreasohlund I took the hotfix from master then realized the smoketest stuff I wanted to include was in develop. so i rebased the hotfix to develop. which only half worked so I merged develop into hotfix-1.3.5. In a nutshell it's all related to improving the quality of the app and solving those bugs

PR #227, opened 2015-11-18
Fixed error with signalr calls not broadcasting properly

petersgiles:
Also added a basic signalr listener page to watch message traffic from
ServiceControl


petersgiles:
@johnsimons please review

cc/ @colin-higgins

Issue #228, opened 2015-11-18
ServicePulse commandline arguments are not validated


ramonsmits:
ServicePulse command-line arguments are not validated. They are case sensitive and having an argument in incorrect casing or having a typo just ignores it.

petersgiles:
@gbiellem maybe we should look at this http://www.nuget.org/packages/NDesk.Options/ I've used it before and it has a nice api for handling command-line args

gbiellem:
Not sure it's necessary to add a new package, SP is already using the same lib for handling param sets that SC uses and it works pretty well.  I think all that's needed is to ensure it's actually set up correctly

So i think all that's needed is

- Removing case sensitivity ( should be an easy fix.)
- Validate that the provided SC and SP URLs are valid URIs.  
- Ensure the param sets work as advertised
- Update the help text to ensure that users know what the param sets do

Validation of the servicename should be unnecessary as the underlying API will throw if that's a problem.

petersgiles:
ok that sounds fine

gbiellem:
@petersgiles I just realized that the package you proposes above is actually the one we are already using :)

petersgiles:
hahaha...well great minds think alike. from memory that one lets you ignore case, or at least compensate for it pretty easily

PR #232, opened 2015-11-27
Add command line validation

gbiellem:
related to #228


gbiellem:
@petersgiles I think this is ready for review

petersgiles:
So you had to override OptionSet to do Caseless stuff. Nice work. The tests seem to cover everything I can think of. Its so nice to have unit tests in ServicePulse

gbiellem:
> So you had to override OptionSet to do Caseless stuff.

I did consider modifying optionset directly but this way it should be more upgrade friendly.

petersgiles:
Yes, I preferred what you did.

PR #244, opened 2016-01-14
Alerts User to Upgrade ServiceControl

petersgiles:
@johnsimons @mikeminutillo please review

### DIFF
--- a/src/ServicePulse.Host/app/js/directives/ui.particular.productVersion.tpl.html
+++ b/src/ServicePulse.Host/app/js/directives/ui.particular.productVersion.tpl.html
@@ -6,9 +6,15 @@
             <li><a href="#" target="_blank">ServicePulse</a> <span class="greyNote"> <span class="vWord">Version</span> {{version}}</span></li>
         </ul>
 
-        <span ng-show="newversion == true">
+        <span ng-show="newspversion == true">
             <span class="newVersionLink">
-                <a href="{{newversionlink}}" title="Download ServicePulse" target="_blank">Get new version!</a>
+                <a href="{{newspversionlink}}" title="Download ServicePulse" target="_blank">Upgrade ServicePulse!</a>
+            </span>
+        </span>
+        
+        <span ng-show="newscversion == true">
+            <span class="newVersionLink">
+                <a href="{{newscversionlink}}" title="Download ServiceControl" target="_blank">Upgrade ServiceControl!</a>
             </span>
         </span>
 
--- a/src/ServicePulse.Host/app/js/services/services.platform-update.js
+++ b/src/ServicePulse.Host/app/js/services/services.platform-update.js
@@ -8,24 +8,26 @@
 
 
             var serviceProductUrls = [
-                { product:'SP', url: '//platformupdate.particular.net/servicepulse.txt' },
+                { product: 'SP', url: '//platformupdate.particular.net/servicepulse.txt' },
                 { product: 'SC', url: '//platformupdate.particular.net/servicecontrol.txt' }
             ];
 
-            return $q.all(serviceProductUrls.map(function (item) {
-                return $http({
-                    method: 'GET',
-                    url: item.url
+            return $q.all(serviceProductUrls.map(function(item) {
+                    return $http({
+                        method: 'GET',
+                        url: item.url
+                    });
+                }))
+                .then(function(results) {
+                    var resultObj = {};
+                    results.forEach(function(val, i) {
+                        resultObj[serviceProductUrls[i].product] = val.data;
+                    });
+                    return resultObj;
+                }, function () {
+                    // swallow errors
+                    return {};
                 });
-            }))
-          .then(function (results) {
-              var resultObj = {};
-              results.forEach(function (val, i) {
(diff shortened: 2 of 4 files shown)


gbiellem:
@petersgiles  Does this version swallow the error if it can't read from the feeds? The old one did and I don't think this one does.

petersgiles:
I can't remember why we needed it to, I'll fix it fix it fix it

gbiellem:
We need it for corporate network with no internet access... Sounds like a place we worked. :)

petersgiles:
Yes, I remember. It swallows errors now

src/ServicePulse.Host/app/js/directives/ui.particular.productVersion.tpl.html
```diff
@@ -6,9 +6,15 @@
             <li><a href="#" target="_blank">ServicePulse</a> <span class="greyNote"> <span class="vWord">Version</span> {{version}}</span></li>
         </ul>
 
-        <span ng-show="newversion == true">
+        <span ng-show="newspversion == true">
```
johnsimons: Should it be === ?

src/ServicePulse.Host/app/js/directives/ui.particular.productVersion.tpl.html
```diff
@@ -6,9 +6,15 @@
...
             <span class="newVersionLink">
-                <a href="{{newversionlink}}" title="Download ServicePulse" target="_blank">Get new version!</a>
+                <a href="{{newspversionlink}}" title="Download ServicePulse" target="_blank">Upgrade ServicePulse!</a>
+            </span>
+        </span>
+        
+        <span ng-show="newscversion == true">
```
johnsimons: see above

johnsimons:
So this check is only done once (in theory) right ? Because we are a SPA.

src/ServicePulse.Host/app/js/directives/ui.particular.productVersion.tpl.html
petersgiles: no that's fine i'm only comparing values not values and type

petersgiles:
the check is done whenever we go to the dashboard page the browser will cache the call to aws

src/ServicePulse.Host/app/js/directives/ui.particular.productVersion.tpl.html
petersgiles: "==" means equals, whereas "===" means identically equal.

petersgiles:
@gbiellem you may have also noticed that i took the aws url out of app.constants it's our business no one should need to configure it

gbiellem:
@petersgiles  I did notice. And then I jumped on docs to make sure it didn't need a doco change over :) 

The only other thing to check is the installer,  It does some trickery on upgrade.  It tries to parse an existing version of app.constants.js and apply whatever the currently configured SC url is to the new app.constants.js. 

So since the file has changed we should double check that this still works

petersgiles:
Yep, I do remember we do a regex kind of search for the version property and the same for the control url. The pulse_url wasn't included. 

Except the was some compensating version migration stuff that might have it

src/ServicePulse.Host/app/js/services/services.platform-update.js
```diff
@@ -2,30 +2,41 @@
...
+                { product: 'SC', url: '//platformupdate.particular.net/servicecontrol.txt' }
+            ];
+
+            return $q.all(serviceProductUrls.map(function(item) {
+                    return $http({
+                        method: 'GET',
+                        url: item.url
+                    });
+                }))
+                .then(function(results) {
```
WilliamBZA: I'm not a huge fan of forEach.

```Javascript
return results.reduce(function (acc, val, i) {
    acc[serviceProductUrls[i].product] = val.data;
    return acc;
}, {});
```

src/ServicePulse.Host/app/js/services/services.platform-update.js
```diff
@@ -2,30 +2,41 @@
...
+                    var resultObj = {};
+                    results.forEach(function(val, i) {
+                        resultObj[serviceProductUrls[i].product] = val.data;
+                    });
+                    return resultObj;
+                }, function () {
```
WilliamBZA: Shouldn't swallow the error here.

We're returning a promise, so let the caller of `getReleases()` decide whether to swallow the exception or not.

I would just leave the fail overload out so that callers can decide what to do.

petersgiles: normally I would agree but the purpose of this service is to return something nice for the directive to work with. I don't want to directive making decisions about how to react to network errors etc I only want it to act on the data provided. For now I'm going to leave it this way. When we get to improve SP's overall design we can standardize our approach to this kind of thing.

src/ServicePulse.Host/app/js/services/services.platform-update.js
petersgiles: yes, I keep forgetting about reduce. A bit Mathy for my brain but much more efficient. I'll swipe your code and use it instead, thanks

PR #248, opened 2016-01-27
Upgrade servicecontrol link

petersgiles:
@sergioc merge when happy


petersgiles:
@gbiellem you might like this one merge if happy

gbiellem:
@petersgiles  I'm nitpicking here but when SC is offline the text displayed in the right hand pane is `ServiceControl V     `.  I would have expected it not to be displayed at all.  

Also on this branch the script shows an "error.fatal" notification when SC is not available prior to the notification that says SC is not available.  I assume this is the version check throwing but not being swallowed.   I think you had another branch that addressed this.

petersgiles:
Thanks @gbiellem those problems are fixed now

Issue #251, opened 2016-02-01
Alert notification messages stack unnecessarily and push the rest of the UI down


sergioc:
When events (messages fail, failed messages successfully archived or retried) occur in rapid succession in the failed messages screen, the UI gets littered with several alert messages, and the rest of the UI is pushed down:

![image](https://cloud.githubusercontent.com/assets/122651/9186569/982c8db4-4009-11e5-9f40-5c258d5ffc8e.png)

There's a few issues with this:
- the alert style we use is UI pattern that doesn't really work well with a rapid succession of several events
- the messages push the UI down (thus possibly making the user miss a click in the UI ([related Desk case](
https://nservicebus.desk.com/agent/case/16617))
- they require manual dismissal (possibly a lot of labor involved, depending on no. of messages)

This can be addressed in the following way:

- Remove alerts for failed messages. Instead show a count badge next to the "Failed messages" nav link and clean up the title for the failed messages container. This way it's more clear when it gets updated and the count badge is visible at all times
- Show only a maximum number (e.g. 3) of toast messages for archive and retry successes and dismiss automatically
  - Further improvements could be done by replacing those toast messages with in-context status messages: https://github.com/Particular/ServicePulse/issues/252

Demonstration of problem http://screencast.com/t/MbSRv32fw

petersgiles:
quick example of badge next to failed messages
![screenshot 2016-02-01 13 51 30](https://cloud.githubusercontent.com/assets/1644481/12708345/aa8a94de-c8f2-11e5-83d1-fc8b59052e36.png)

mauroservienti:
this should be prioritized, @ramonsmits and I experienced that in an environment where the failed message throughput is high the SP UI is unusable.

petersgiles:
@mauroservienti I'm working on it now

johnsimons:
Because the menu does not scroll, and that info quite possible should be acessible all the time, I would instead move the display of the  info to the top bar and I would make it static so it does not scroll, here is an example of what I mean:
![image](https://cloud.githubusercontent.com/assets/122651/12998279/6aa5a4aa-d18d-11e5-8835-924a7c6c7b91.png)

@sergioc I am pretty sure you have a design like this ?

sergioc:
@johnsimons yes, memory didn't fail you:

![screenshot 2016-02-12 10 42 57](https://cloud.githubusercontent.com/assets/493451/12998348/67fcf3e2-d175-11e5-9937-181eda19874b.png)

[...]

PR #254, opened 2016-02-01
[WIP]: toastr notifications

petersgiles:
Started to replace notifications with toasts. This helps with bugs around the interface moving when a notification is presented.

related to: https://github.com/Particular/ServicePulse/issues/251


petersgiles:
@sergioc this has the toast pop up in it now, it's still a little 'noisy' we shall have to aggregate some of the notifications I think.

sergioc:
Toast message types:

**Error type:**
- ServiceControl connection lost (warning)
- *other types of SC errors*

**Success:**
- **Group retry successful** N messages from group XXX successfully retried 
- **Group archive successful** N messages from group XXX successfully archived

**Warning**
- N messages failed (this toast wouldn’t be repeated, but the number inside the toast would simply update)
- **Group retry failed** N messages from group XXX failed to retry
- **Group retry failed** N messages from group XXX failed to archive

**Additional requirements:**
- Assuming we can easily aggregate the "N messages failed” toast, then all messages should be sticky and require the user to manually dismiss.
- Clicking on toast related to failed messages dismisses it and takes user to failed messages screen, “Messages group” showing
- Nice to have: a version in the singular for messages with counts.

johnsimons:
Instead of fly in toasts (which can become quite distracting), should we consider the approach that Azure portal uses (the static panel at the bottom) ?

sergioc:
@johnsimons Azure screenshot?

PR #285, opened 2016-02-15
Hotfix 1 3 9

petersgiles:


WilliamBZA:
I don't think this is necessary. I'd rather not release a hotfix for this, and then another release once the toastr notifications are ready.  I'd much rather just release the toastr one.

There's no need to knee-jerk to fix this, so rather than trying to move fast, let's rather do the right thing and release a proper fix when it's ready.

sergioc:
@WilliamBZA :+1:

petersgiles:
@WilliamBZA Oh I totally agree...we can park this

johnsimons:
:+1: to close this one and not release

petersgiles:
Sounds like no one want this to go ahead, I'm closing this and removing the issues from this milestone.
/cc @ramonsmits @hmemcpy

Issue #297, opened 2016-03-14
Feedback on ServicePulse 1.4.1 Endpoints screen


petersgiles:
Task Force: @sergioc @petersgiles

### Plan of Attack

- [x] take the example of tabbing on the Messages Screen and apply it to the endpoints screen to separate active from in active

---

_Raised on behalf of Sudheendra Bhat_

It is great to see that there are new versions keep coming for all of your products. I have a feedback on ServicePulse version 1.4.1. It is for endpoints overview page.

The purpose of the dashboard (at least for us), is to know what's going wrong in the system. So maybe it is best to show inactivate endpoints in the top than active endpoints. From the notification I see the count of endpoints that are inactive but I need to scroll down to see the names. And it is difficult to stop scrolling when the inactive endpoints section starts due to the current display style. Please let me know if it is possible to customize so that we can make that change in our instance. Thank you :)

petersgiles:
@sergioc  solutions that come to mind

It currently looks like this.

![endpoint overview](https://cloud.githubusercontent.com/assets/1644481/13767813/e4fd4f2a-eac2-11e5-89ee-4a4a25b14644.PNG)

Maybe we could do something like the failed messages screen with the tabs

![failed messages](https://cloud.githubusercontent.com/assets/1644481/13767856/60052bc0-eac3-11e5-98bc-e79854ec9aeb.PNG)

sergioc:
I think tabbing is a great solution.

petersgiles:
![capture](https://cloud.githubusercontent.com/assets/1644481/13801244/367d154e-eb84-11e5-8c38-d338603f52ba.PNG)

PR #299, opened 2016-03-16
Added Tabs to Endpoint Screen

petersgiles:
fixes https://github.com/Particular/ServicePulse/issues/297


sergioc:
Mouse pointer when hovering tabs should be pointer and the text should turn blue and underlined. See failed messages screen.

Issue #241, opened 2016-01-04
Add the ability to Restore messages from an Archived State


petersgiles:
Sometimes customers archive messages they didn't mean to. There needs to be a way to reverse the archive action.

@sergioc any thoughts on how this might work?

related to: https://github.com/Particular/ServiceInsight/issues/506

sergioc:
@pete as a simple and possibly effective solution, I think we could show an on/off toggle for an archived messages view for each of the failed messages tabs (Error groups and Messages).

When toggled on, SP only shows messages or message groups that were archived, and the action to unarchive next to each item.

sergioc:
Another option is to add 2 new tabs "Archived messages" and "Archived error groups".

petersgiles:
I like the second option or at least the notion that this would appear in another place. I would hate to clutter up the current page...it already feels like its doing too much.

sergioc:
I think the first option has potential to add little clutter as well:

![screenshot 2016-01-05 11 37 16](https://cloud.githubusercontent.com/assets/493451/12107716/5085bfd8-b3a1-11e5-837c-9460dc040c4a.png)

petersgiles:
When you described option 1 I thought you meant that the existing buttons would still be visible.

Using a toggle like that is clean but it is effectively a navigation to a completely different view and therefore more like how I imagined option 2. And because it is pretty much a fresh view I'd be inclined to just navigate to it from the sidebar as there is no related context to the tasks you would perform on the failed messages screen.

sergioc:
Placing on the sidebar is another valid option, though you'd be cluttering it with a new link that is possibly rarely used, but would then permanently show on any view.

How likely are you to jump directly to the Archived messages view from another screen which doesn't relate to failed messages?

petersgiles:
I think its the kind of thing, as an ops person, I would want to go to directly. The scenario I'm imagining would be something like I've just archived a bunch of messages related to database connectivity errors and then the DBA phones up and says OK you're good to replay all those now. 

But I just archived them, so I'd go to the archive screen look at my archive history and restore the relevant messages from the archive screen.

Then I'd navigate to the failed messages screen so I could review them as failed messages and make a choice about what to do as I normally would

sergioc:
> I think its the kind of thing, as an ops person, I would want to go to directly. The scenario I'm imagining would be something like I've just archived a bunch of messages related to database connectivity errors and then the DBA phones up and says OK you're good to replay all those now. 

[...]

PR #298, opened 2016-03-18
Create a Retention policy screens for retention policy changes in ServiceControl

johnsimons:
task-force: @petersgiles @johnsimons @sergioc @boblangley 

//cc @Particular/servicepulse-maintainers @mat-mcloughlin @WilliamBZA 

Component of https://github.com/Particular/PlatformDevelopment/issues/667
Related to https://github.com/Particular/ServiceControl/issues/691

related to: https://github.com/Particular/ServicePulse/issues/241

## Plan of Attack

- [x] A new screen for archive
- [x] display a count of archived messages

failed-groups-tab

- [x] review individual archived messages

archived-messages

- [x] unarchive (when messages are unarchived the retry retention countdown resets

archived-messages-select-to-unarchive

- [x] sorting & filtering

archived-messages-time-groups

- [x] ServiceControl version incompatibility screen that will stop ServicePulse from working with lower/incompatible versions 

to test go to factory.shareddata and change the minimum version
version_incompatibility

You'll get redirected to an about screen with some information on it the top nav will hide and toasts will pop too

incompatible

- [x] Changes to documentation

Connects to https://github.com/Particular/PlatformDevelopment/issues/667


sergioc:
@petersgiles Since the About screen isn't accessible anywhere normally, I changed it to focus on incompatibility warning notice.

sergioc:
@petersgiles to-do from from today's sync:

- check width of archived messages list
- adjust dropdown filter button text to better indicate active state
- Do not show “Scheduled for deletion immediately” messages and subtract those from total counts

Notes from testing with @mauroservienti:

- Failed groups count doesn’t update if you unarchive a message from a group that had no failed messages before
- Should we add a help tooltip icon next to “Scheduled for deleting” with more info about what it means and how to configure the deletion policy?

petersgiles:
> Since the About screen isn't accessible anywhere normally, I changed it to focus on incompatibility warning notice.

I was thinking that one day it will grow to include current configuration of everything so that devops know what they are dealing with without having to hunt around for stuff. but that's cool for now.

petersgiles:
> Do not show “Scheduled for deletion immediately” messages and subtract those from total counts

The list management that this requires is a little tricky. I will attempt this as a nice to have after all the doco etc is done.

> Failed groups count doesn’t update if you unarchive a message from a group that had no failed messages before

fixed (the counts will update on a 5 sec timer as I have to poll for accurate data here)

boblangley:
Ran ServicePulse from this branch with ServiceControl delete_errors branch. Everything related to failed messages works as expected, I just have not yet been unable to confirm the archive date filtering since I do not have any messages that old.

Found issues with unarchive button tooltip and the unarchive confirm dialog, they appear to share the message with the archive button / confirm. I am working on a commit to adjust these.

petersgiles:
just storing these here for history

## original mockups

new mockup 1_2

new mockup 1

sergioc:
> Found issues with unarchive button tooltip

Buttons with text labels should not require tooltips, that adds unnecessary redundancy. An issue was recently closed to address this, but it seems there's some leftovers, so I cleaned that up.

sergioc:
@petersgiles during testing with @ramonsmits:

- we missed the option to unarchive all messages. Should that be made possible in this release or is there a reason you didn't include it?
- the notation for failure timestamp for archived messages is not consistent with that in the "all failed messages" tab
- there's no option to immediately delete a (selected) archived message. Why?

petersgiles:
@sergioc @ramonsmits 

> we missed the option to unarchive all messages. Should that be made possible in this release or is there a reason you didn't include it?

I don't think we can do that in any meaningful way until grouped messages are implemented in servicecontrol

> there's no option to immediately delete a (selected) archived message. Why?

That would require an extension to the api that is outside the scope of this fix. Users can use the SC management tool to reduce the time to keep down to a minimum value. We were afraid if gave the option to select all and then a delete immediately function that users would risk putting too much load on the system. we can review and refine later.

sergioc:
>> we missed the option to unarchive all messages. Should that be made possible in this release or is there a reason you didn't include it?

> I don't think we can do that in any meaningful way until grouped messages are implemented in servicecontrol Particular/PlatformDevelopment#635

What does this have to do with grouping?

petersgiles:
Ok what is the 'ALL' I am meant to select? 

- Is it something that is time bracketed?
- Is it what is currently displayed?
- Is it everything in the current filter?
- Is it actually everything?
- Is it the exception group of the currently selected message?

The current API doesn't support these scenarios easily. From the client I can take a solid guess at some of them but I don't always have full information. I want to avoid querying everything in a data store that potentially has 1,000,000+ records just to determine if I can delete some of them or not.

sergioc:
"All" means whatever can be showed on this screen (may not yet be displayed, but will be if you scroll to load more), based on the filtered time period you selected.

> I want to avoid querying everything in a data store that potentially has 1,000,000+ records just to determine if I can delete some of them or not.

We're referring to the ability to *unarchive* all, not delete.

PR #304, opened 2016-03-18
Relative instead of absolute paths

ramonsmits:
Menu links referred to the root of the website. If hosting in IIS this prevents the deployment to for example:

    http://localhost/management/ServicePulse


ramonsmits:
https://nservicebus.desk.com/agent/case/17684

petersgiles:
will be included in 1.5.0

PR #328, opened 2016-04-20
Attempt to "switch on" an endpoint in the "Endpoints Configuration" fails

johnsimons:
## Symptoms
When trying to switch on the monitoring of an endpoint in the "Endpoints Configuration", the UI updates as if it all succeeds, but if the page is refreshed the monitoring setting resets back to off.

## Who's affected
All users of ServiceControl that are currently running v1.11.

## Description of issue
It seems when angular sends a PATCH request the `Content-Type` is set to `application/json-patch+json`, this content-type is not recognised by Nancy, hence Nancy is not able to correctly bind to the model and default the flag to `false`.

### DIFF
--- a/src/ServicePulse.Host/app/js/app.http.js
+++ b/src/ServicePulse.Host/app/js/app.http.js
@@ -9,10 +9,8 @@
 
         $httpProvider.defaults.headers.patch = {
             'Accept': 'application/json, text/javascript',
-            'Content-Type': 'application/json-patch+json'
+            'Content-Type': 'application/json'
         }
-
-   
     };
 
     httpProvider.$inject = [


johnsimons:
@Particular/servicepulse-maintainers I need help with this one, not sure what makes Angular send `application/json-patch+json`!

petersgiles:
There's an example of sending a header with rest call in one of the recent
service Libs I think from memory if you look in the archive service you
might find what you are after

colin-higgins:
To get the fix out, can we just use a POST for now until we re-evaluate?

johnsimons:
> can we just use a POST for now until we re-evaluate?

@colin-higgins nope, ServiceControl won't accept that

johnsimons:
@Particular/servicepulse-maintainers please review

johnsimons:
thanks @colin-higgins, I get this released now

Issue #326, opened 2016-04-08
Service Control URL without Trailing Slash 404s


colin-higgins:
## Problem
If the ServiceControl URL in app.constants does not have a trailing slash:
The concatenation for requests throughout ServicePulse results in 404s.

## Fix
When loading the URL, add the trailing slash if it is missing

petersgiles:
That's what this code is for https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host/app/js/services/services.uri.js

And these tests https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host.Tests/tests/js/services/services.spec.js

Do we need to add a specific scenario to the tests?

petersgiles:
Hmmm weird

![trailing slash](https://cloud.githubusercontent.com/assets/1644481/14371188/8c6fc9b4-fd78-11e5-87f6-46edc67567df.PNG)

colin-higgins:
Is uri.join definitely being used for every concatenation?

petersgiles:
I certainly hope so, can you dig around and find the call that's failing for you?

petersgiles:
looks like there's been a bug with this since 1.4.0. the immediate fix is to have a trailing slash on the url.
i

petersgiles:
It looks like the urls in the routes don't work when extracted to IIS if the trailing slash is missing

petersgiles:
This needs to become
``` C#
templateUrl: 'js/views/dashboard/dashboard.html',
```
that
``` C#
 templateUrl: '/js/views/dashboard/dashboard.html',
```
but it reverts to absolute urls which hurts people who extract to sub directories of IIS


we could add another config param to mark the deployment as relative and then construct a different url.

@colin-higgins what do you think?

colin-higgins:
why something like this pseudo code?
```
if last character of uri is not '/' 
    then add '/'
concatenate uri and relative url
replace all instances of '//' with '/'
```

petersgiles:
yeah that might do it.

I think the whole thing needs to be setup once in the shared data factory. The uri service was a bit of a hack I put together before I looked at making a place to store global variable. I never really liked it but it got the job done.

adamralph:
Moving back to backlog after discussing with @colin-higgins

johnsimons:
@colin-higgins @adamralph so did anyone updated the desk case and gave the answer to the customer ?

noctovis:
[...]

PR #329, opened 2016-04-21
Refactor uri service

colin-higgins:
Connects to https://github.com/Particular/ServicePulse/issues/326

Refactor uri service to handle leading slashes.
Remove sample tests that fail.


johnsimons:
@colin-higgins does this still need to be merged ?

colin-higgins:
I think it's still a good PR for cleaning some stuff up. Thoughts @petersgiles ?

petersgiles:
yes I agree that the unused sample tests need to be dropped

colin-higgins:
Added a test for the leading slash trimming and rebased

PR #335, opened 2016-04-27
Fix Route Change Error when hosting in IIS

### LINKED ISSUE #334: Route Change Error when hosting in IIS
When hosting ServicePulse in IIS, the page fails to load with a "Route Change Error". 

This issue was described by @noctovis here https://github.com/Particular/ServicePulse/issues/326#issuecomment-214860543:

> I have the exact same issue right now as we're trying to setup SP under IIS. We have different ServiceControl instances per environment, so I have no choice but setting up a SP instance per environment too. Manually modifying the route files and adding that slash hasn't helped either. Also, modifying the code would make upgrading to new versions terribly hazardous. I keep getting 406 errors and the generic "Route change error"-popup. Any idea if and when this issue will be fixed? Thanks in advance!

by inspecting the requests shows that requesting views via xhr returns a HTTP 406 Error from IIS. A fist ~~investigation~~ google search indicates that this is an issue with headers set by angular (e.g. ` $httpProvider.defaults.headers[...]`, see http://stackoverflow.com/questions/33718861/angularjs-on-iis-406-error. Removing these headers indeed resolves the issue and loads the page correctly with IIS.

timbussmann:
fixes #334 by removing headers which seem to cause issues when hosting in IIS. I didn't see any impact by this change when hosting with Nancy?

also changed the links on the dashboard to use relative urls (like the navbar) so it works correctly with virtual directories.

@Particular/servicepulse-maintainers @petersgiles @colin-higgins @WojcikMike please review

### DIFF
--- a/src/ServicePulse.Host/app/js/app.http.js
+++ b/src/ServicePulse.Host/app/js/app.http.js
@@ -2,15 +2,7 @@
     'use strict';
 
     function httpProvider($httpProvider) {
-
-        $httpProvider.defaults.headers.common['Accept'] = 'application/json, text/javascript';
-        $httpProvider.defaults.headers.common['Content-Type'] = 'application/json; charset=utf-8';
         $httpProvider.defaults.useXDomain = true;
-
-        $httpProvider.defaults.headers.patch = {
-            'Accept': 'application/json, text/javascript',
-            'Content-Type': 'application/json'
-        }
     };
 
     httpProvider.$inject = [
--- a/src/ServicePulse.Host/app/css/particular.css
+++ b/src/ServicePulse.Host/app/css/particular.css
@@ -242,7 +242,7 @@ footer a:hover {
 }
 
 .loading {
-    background-image: url('/img/loader_spinner.gif');
+    background-image: url('../img/loader_spinner.gif');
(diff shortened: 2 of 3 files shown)


WojcikMike:
Looks good to me :+1:

colin-higgins:
Tested on self host, IIS and IIS nested. Works great.

We also resolved this issue: https://github.com/Particular/ServicePulse/issues/336

src/ServicePulse.Host/app/js/app.http.js
```diff
@@ -2,15 +2,7 @@
     'use strict';
 
     function httpProvider($httpProvider) {
-
```
johnsimons: Why remove these ?
Are these headers also used when loading views and other resources from ServicePulse itself ? I thought they were only used for ServiceControl REST api calls

timbussmann: @johnsimons yes, these headers are added to all XHR requests. Quote from Angular docs:
> The $http service will automatically add certain HTTP headers to all requests. These defaults can be fully configured by accessing the $httpProvider.defaults.headers configuration object, which currently contains this default configuration

Issue #343, opened 2016-05-10
IIS hosting only supports site root for ARR proxy


rlaveycal:
## Symptoms

ServicePulse does not receive updates from the ServiceControl

## Who is affected

Any user attempting to install ServicePulse using ARR proxy in a non-root directory.

## POA

- [x] Update ServicePulse SignalR version to latest stable. #345
- [x] Smoke Test ServiceControl / ServicePulse
- ~~[ ] Create new release of ServicePulse~~
- ~~[ ] Announcement~~

## User Report:
Following the instructions to install under IIS (http://docs.particular.net/servicepulse/install-servicepulse-in-iis) works so long as the `api` proxy folder to service control is in the site root. If it's not (e.g. to host multiple service pulse instances talking to multiple service control instances) then the SignalR connect and related calls don't respect the correct URL and continue to use `/api/messagestream'

E.g. with service pulse installed in a /ServicePulse folder IIS logs show
```
GET /servicepulse/api/messagestream/negotiate _=1462878929505&X-ARR-CACHE-HIT=0&X-ARR-LOG-ID=6425b0eb-624e-4787-a75d-cbe5b3696c9d&SERVER-STATUS=200 80 - ::1 Mozilla/5.0+(Windows+NT+6.3;+WOW64;+Trident/7.0;+rv:11.0)+like+Gecko http://localhost/servicepulse/ 200 0 0 0
[...]

PR #345, opened 2016-05-12
Signalr 2.2.0

boblangley:
## Symptoms

ServicePulse does not receive updates from the ServiceControl

## Who is affected

Any user attempting to install ServicePulse using ARR proxy in a non-root directory.

connects to #343


src/ServicePulse.Host/app/js/services/factory.shareddata.js
```diff
@@ -32,7 +32,7 @@
 
         var environment = {
             sc_version: undefined,
-            minimum_supported_sc_version: "1.14.0",
+            minimum_supported_sc_version: "1.15.0",
```
johnsimons: The version of SC will be 1.16.0

johnsimons:
looks good, but It seems gitversion is getting confused with the branch name called `signalr-2.2.0` so we need to merge this to a `release-1.6` before merging it to `master` to get the correct version.

PR #346, opened 2016-05-12
Perform a case-insensitive lookup for the manifest resource

WilliamBZA:
Connects to https://github.com/Particular/ServicePulse/issues/347

Instead of only doing a case-sensitive load of the resource, try and find a case-insensitive match if the requested case results in no resource found.


mat-mcloughlin:
LGTM

Issue #354, opened 2016-05-27
NoIE redirect doesn't work with IIS hosting in a sub directory


rlaveycal:
## Who's affected

Any users using Internet Explorer less than version 9 and is running ServicePulse within a sub directory

## Symptoms

If the browser is an unsupported version of Internet Explorer and ServicePulse is hosted within a sub directory, the browser will be redirected to the root directory looking for a file `NoIE.html` in the root.

## Original bug report

The browser check in Index.html
```html
    <!--[if lt IE 9 ]>
        <script type="text/javascript">
            window.location = '/NoIE.html';
        </script>
    <![endif]-->
```

assumes that pulse is in the root folder. It needs changing to be a relative url
```javascript
            window.location.href = 'NoIE.html';
```

WilliamBZA:
Thanks @rlaveycal. I've made the change and submitted a PR. Just waiting for another set of eyes to confirm.

WilliamBZA:
@rlaveycal ServicePulse version `1.6.2` has just been released and should resolve this.

PR #355, opened 2016-05-27
Changed NoIE html url to not force root

WilliamBZA:
Fixes https://github.com/Particular/ServicePulse/issues/354


WilliamBZA:
Ping @Particular/servicepulse-maintainers for review.

I can't think of a scenario where the path would be different from the current path, so this should be a safe change.

sergioc:
:+1:

Issue #312, opened 2016-03-31
Expanded mobile nav menu won't auto-close after clicking to switch screens


sergioc:
### Symptom
When using navigation menu on mobile device it doesn't collapse after clicking outside of menu.

### Who's affected
Users who use Service Pulse on mobile devices.

### Steps to reproduce:

- switch to an SM breakpoint
- open nav menu
- click to switch another screen

PR #356, opened 2016-05-31
Hiding mobile navigation when the link is clicked

WojcikMike:
Fixes: #312

### DIFF
--- a/src/ServicePulse.Host/app/js/app.controller.js
+++ b/src/ServicePulse.Host/app/js/app.controller.js
@@ -27,6 +27,11 @@
             return active;
         };
        
+        $($window.document).on('click', function (e) {
+            if ($('.navbar-collapse.in').is(':visible')) {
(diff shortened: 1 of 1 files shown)


WojcikMike:
@Particular/servicepulse-maintainers please review and merge. 
Current behaviour is that when the mobile menu is expanded clicking on any of the links on the menu will collapse it. 
There is an question should we actually expand the functionality to collapse the menu even when you click outside of the menu itself. I am gently leaning towards this as currently being on main screen one can click on "Failed Message" big button and the menu will stay expanded witch is a bit counter intuitive.

WilliamBZA:
@WojcikMike I'm happy with the code, but should we not address the question you outlined before merging?

WojcikMike:
After discussing with @WilliamBZA @kbaley the decision was made to hide the menu when user click outside of it. The code was changed accordingly and issue description as well

sergioc:
:+1:

Issue #358, opened 2016-06-10
Error when entering /failedMessages page directly


johansuperstar:
The following error occurs and the page won't render properly when visiting url **/failedMessages** without having clicked the ALL FAILED MESSAGES link in the navigation. 

`angular.js:12416 TypeError: Cannot read property 'hasOwnProperty' of undefined
    at new controller (http://secc3459:9090/js/views/failed_messages/controller.js:20:39)
    at invoke (http://secc3459:9090/lib/angular/angular.js:4473:17)
    at Object.instantiate (http://secc3459:9090/lib/angular/angular.js:4481:27)
    at http://secc3459:9090/lib/angular/angular.js:9108:28
    at link (http://secc3459:9090/lib/angular/angular-route.js:977:26)
    at invokeLinkFn (http://secc3459:9090/lib/angular/angular.js:8746:9)
    at nodeLinkFn (http://secc3459:9090/lib/angular/angular.js:8246:11)
    at compositeLinkFn (http://secc3459:9090/lib/angular/angular.js:7637:13)
    at publicLinkFn (http://secc3459:9090/lib/angular/angular.js:7512:30)
    at boundTranscludeFn (http://secc3459:9090/lib/angular/angular.js:7656:16) <div id="view" ng-view="" class="ng-scope" data-ng-animate="1">`

It's probably due to this line: https://github.com/Particular/ServicePulse/blob/9b2159b4e4de4f40326fc3b87c6d901ba67c066a/src/ServicePulse.Host/app/js/views/failed_messages/controller.js#L20

That line of code also seems to want to redirect to the **/failedGroups** page if the selectedExceptionGroup is incomplete. However, I would like to have the possibility to link directly to the **/failedMessages** page.

To reproduce: 
- Go to **/failedMessages** 
- Clear local storage
- Refresh page

WilliamBZA:
Good catch, thanks @johansuperstar!

I've gone through the code path that gets us to that error. The data for this page is loaded by the `failed_groups` controller and is persisted in LocalStorage. The `failed_messages` view then expects that the data has been loaded and persisted already. Obviously this is not going to be the case when navigating directly to the Failed Messages page.

To fix this we need to do the following:
* [ ] Extract the error messages loading logic to a service that can be called by both the `failed_groups` controller as well as the `failed_messages` controller
* [ ] Refactor the controllers to use the new service instead of loading the data directly
* [ ] Wire the loading phase up to the loading indicators

[...]

PR #359, opened 2016-06-15
Load failed messages if not in local storage

WilliamBZA:
## Who's affected

* Any user of ServicePulse that attempts to browse directly to the `FailedMessages` page without first going to the `FailedGroups` page

## Symptoms

An error occurs and the page won't render properly when visiting url /failedMessages without having clicked the ALL FAILED MESSAGES link in the navigation.

## Original bug report

[Error when entering /failedMessages page directly](https://github.com/Particular/ServicePulse/issues/358)

Also included are a few code style change.


WilliamBZA:

WojcikMike:
I looked throught the code and verified that the fix works

PR #365, opened 2016-07-01
Add X-UA-Compatible for Servicepulse

shamglam:
Servicepulse fails to work on Internet Explorer 11 in intranet environments where IE is configured by group policy to display intranet sites in compatiblity mode. This fix adds an http header to servicepulse to fix this issue for affected browsers.

### DIFF
--- a/src/ServicePulse.Host/Nancy/SpecialEmbeddedFileResponse.cs
+++ b/src/ServicePulse.Host/Nancy/SpecialEmbeddedFileResponse.cs
@@ -24,6 +24,7 @@
             }
 
             this.WithHeader("ETag", GenerateETag(content));
+            this.WithHeader("X-UA-Compatible", "IE=edge,chrome=1");
             Contents = GetFileContent(content);
         }
 
--- a/src/ServicePulse.Host/Nancy/SpecialEmbeddedFileResponse.cs
+++ b/src/ServicePulse.Host/Nancy/SpecialEmbeddedFileResponse.cs
@@ -24,7 +24,7 @@
             }
 
             this.WithHeader("ETag", GenerateETag(content));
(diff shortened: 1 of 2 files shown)


WilliamBZA:
Thanks for this @shamglam

A couple notes though:

Firstly, can you please sign our [Contributors License Agreement](http://particular.net/contributors-license-agreement-consent)?

Secondly, we have to support hosting Service Pulse on non-Nancy endpoints. Can you please rebase your change to include a HTML meta tag that does the same instead of injecting it as a HTTP header?

`<meta http-equiv="X-UA-Compatible" content="IE=edge">`

And finally, I'm not sure if we'd like to include `chrome=1`. Can you explain why you want to force Chrome Frame instead of simply using Edge?

shamglam:
Well actually now that I see various places that chrome=1 is indeed deprecated so it probably isn't nescessary. I'll add the modifications to the index.html instead. I've also signed the agreement.

WilliamBZA:
The way I read that Stack Overflow answer is that the meta tag is exactly what we want. The typical deployment scenario for Service Pulse is within the Local Intranet zone, meaning the compatibility view still comes into play.

By using the meta tag it means one less step for users deploying in IIS as they don't have to set the configuration variable. The performance hit of a re-render for IE browsers is likely to be negligible as Service Pulse is not a UI heavy application, and being a SPA means the initial re-rendering only happens once (while there's plenty other _loading_ going on too).

It also means that if we want to support any future deployment scenarios that we won't have to re-visit how to implement the HTTP headers in each one.

shamglam:
Ok, reverted the two previous commits and added the header to index.html instead :)

WilliamBZA:
Great, thanks so much. Can I ask you to drop me a [mail](mailto:william.brander@particular.net) so that I can arrange to get a thank you gift sent to you for helping us out?

Issue #342, opened 2016-05-08
Update available notification, mismatch


johnsimons:
The following 2 notification messages should display the same text:
![image](https://cloud.githubusercontent.com/assets/122651/15100889/a7b6a13a-15c3-11e6-98d9-d190c6ead057.png)

I propose we update ServicePulse text to say "(Upgrade to v1.14.0)" same as ServiceControl Management Utility.

This was raised by a user.

WojcikMike:
Fix was deployed with version 1.6.6.

PR #366, opened 2016-07-06
Display the version of service control available for upgrade

WilliamBZA:
Connects to https://github.com/Particular/ServicePulse/issues/342

Shows the text "Upgrade to V1.19.0" (or the corresponding version number) when an update to Service Control is available. This matches with the format used in the Service Control Management facility.

Also fixes a bug where the link would not display if Service Pulse was up to date but Service Control was not.


WojcikMike:
Looks good

Issue #371, opened 2016-08-10
Incorrect version shown when using ServicePulse in IIS


harumph:
I just followed the instructions to run ServicePulse in IIS http://docs.particular.net/servicepulse/install-servicepulse-in-iis

Once I navigated to the website, I immediately noticed the following message:

> ServicePulse v1.2.0 ( update available) 

This is incorrect as I installed Service Pulse using the 1.6.5 installer.

I looked into the installation folder as well as into the extracted ServicePulse under inetpub.
The app.constants.js files have different version values.

C:\Program Files (x86)\Particular Software\ServicePulse\app\js\app.constants.js
```javascript
;(function (window, angular, undefined) {  'use strict';

    angular.module('sc')
        .constant('version', '1.6.5')
        .constant('scConfig', {
            service_control_url: 'http://localhost:33333/api/'
        });

}(window, window.angular));
```

C:\inetpub\websites\ServicePulse\js\app.constants.js:
```javascript
;(function (window, angular, undefined) {  'use strict';

    angular.module('sc')
        .constant('version', '1.2.0')
        .constant('scConfig', {
            service_control_url: '/api'
        });

}(window, window.angular));
```

If the only duplicated and outdated file is the app.constants.js then it's not much of an issue but I don't know if the rest of the application is up to date.

Here's the Powershell script I wrote to extract ServicePulse to inetpub:
```powershell
Function Install-ServicePulse-IIS
{
    param(
      [Parameter(Mandatory=$false)] [Int] $ServicePulsePort
    , [Parameter(Mandatory=$false)] [Int] $ServiceControlPort
    , [Parameter(Mandatory=$false)] [System.IO.FileInfo] $ServicePulseExe
    )

    # Get Defaults
    If ($ServicePulsePort -eq $null) {
        $ServicePulsePort = 9090
    }
    If ($ServiceControlPort -eq $null) {
        $ServiceControlPort = 33333
    }
    If ($ServicePulseExe -eq $null) {
        $ServicePulseExe = Get-Item 'C:\Program Files (x86)\Particular Software\ServicePulse\ServicePulse.Host.exe'
    }

    # Disable default installation
    Set-Service -Name Particular.ServicePulse -StartupType Disabled -Status Stopped
    netsh http delete urlacl http://+:$ServicePulsePort/

    $path = "C:\inetpub\websites\ServicePulse"

    # Extract to inetpub
    & $ServicePulseExe --extract --outPath="$path"

    # Create a new subdirectory called api
[...]

PR #372, opened 2016-08-11
Fix the issue with version when extracting to host in IIS

WojcikMike:
## Who's affected

* Every ServicePulse user that hosts ServicePulse in IIS or any other web server.

## Symptoms

When extracting files according to documentation the version number always show 1.2.0 which is incorrect.

## Original bug report

[Incorrect version shown when using ServicePulse in IIS](https://github.com/Particular/ServicePulse/issues/371)

cc: @Particular/servicepulse-maintainers please review


WilliamBZA:
Looks great!

Issue #370, opened 2016-07-29
Add package manager to web libraries


WojcikMike:
At this moment we are using Nuget to manage dependencies. However for web frameworks we don't have anything. So for upgrading angular, jquery or any other web library we have ot do it manually. 

I think we should introduce something like bower, so that we can update used libraries easily.

@Particular/servicepulse-maintainers what do you think?

WilliamBZA:
:100: :+1:

kbaley:
What @WilliamBZA said

WojcikMike:
Done

PR #373, opened 2016-08-18
Add package manager to web libraries

WojcikMike:
Connects to: #370 

For now I used exactly the same versions of libraries, and pushed them to git, so that we don't need to do any change on builder yet.

The end goal would be to install on team city nodejs and bower and download dependencies before build.

ATM I create: lib-manual to hold a library which I couldn't find on bower:
page-width-functions - I am not sure where it comes from. 

cc: @Particular/servicepulse-maintainers please have a look.


kbaley:
Do we need all the packages in app/lib to be in source control? My memory of bower is fuzzy but I thought the idea was we could use it to get them so we wouldn't need them in source.

WojcikMike:
> Do we need all the packages in app/lib to be in source control? My memory of bower is fuzzy but I thought the idea was we could use it to get them so we wouldn't need them in source.

Your memory serves you well my young padawan. I did it for now, so that Team City doesn't need bower installed. After we coordinate the effort of installing bower on build agents we will be able to exclude the folder from git.

WojcikMike:
For now I will pause work on this as it was a bit hasty to pull bower in. First we want to make decision where we want to take ServicePulse technically and adjust used frameworks to that vision.

WilliamBZA:
I would rather configure Bower on TC than include the packages in the Repo. Once they're in git, they're in forever.

WojcikMike:
Closing for now as it needs to be discussed in more details

Issue #239, opened 2015-12-23
Add option to group messages by name or full namespace in the group view


christoffer-lundqvist:
If your handlers don't take care of exception and just bubble them up, you have no clear exception message it's hard to actually know what is the message type. Specially if the code crashed in a library you don't own. It would be better to always (or and option) to always show the type of the message that failed.

![image](https://cloud.githubusercontent.com/assets/16411065/11972744/d573ffd0-a950-11e5-876c-2aaef2a33ef4.png)

petersgiles:
related to: https://github.com/Particular/PlatformDevelopment/issues/282

boblangley:
related to https://github.com/Particular/ServiceControl/issues/721

WojcikMike:
In the latest version of ServicePulse it is already implemented (1.12)

PR #386, opened 2016-11-22
Allow selection of group classifier

WilliamBZA:
Adds support for displaying groups displayed by different classifications.

Addresses: https://github.com/Particular/ServicePulse/issues/239
Addresses: https://github.com/Particular/PlatformDevelopment/issues/1038

Ping @Particular/servicepulse-maintainers, @janovesk, @mikeminutillo, @SzymonPobiega, @tmasternak, @boblangley

### DIFF
--- a/src/ServicePulse.Host/app/js/views/failed_groups/view.html
+++ b/src/ServicePulse.Host/app/js/views/failed_groups/view.html
@@ -15,7 +15,7 @@
 
         <no-data ng-show="vm.exceptionGroups.length === 0 && !vm.loadingData" title="message groups" message="There are currently no grouped message failures"></no-data>
 
-        <div class="row" ng-show="vm.availableClassifiers.length && vm.exceptionGroups.length">
+        <div class="row" ng-show="vm.availableClassifiers.length && vm.initialLoadComplete">
             <div class="col-sm-12">
                 <div class="btn-group pull-right">
                     <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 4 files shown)


WojcikMike:
LGTM

WilliamBZA:
NB: This should only be merged and released once SC 1.27.0 has been released.

src/ServicePulse.Host/app/js/views/failed_groups/view.html
```diff
@@ -15,9 +15,24 @@
 
         <no-data ng-show="vm.exceptionGroups.length === 0 && !vm.loadingData" title="message groups" message="There are currently no grouped message failures"></no-data>
 
-        <div class="row">
-            <div class="col-sm-12" ng-show="vm.exceptionGroups.length > 0">
+        <div class="row" ng-show="vm.availableClassifiers.length && vm.exceptionGroups.length">
```
tmasternak: This condition means that if we switch to classification that has no groups we won't be able to go back, right?

I think we should remove `vm.exceptionGroups.length` part.

WilliamBZA: Good catch. I've updated.

WilliamBZA:
Changed from [WIP] to ready as SC has been released.

@tmasternak We just need the release notes now and we can release this.

tmasternak:
@WilliamBZA I think that considering final version we have chosen for SC (reusing the already existing classifications) I don't think it's critical to release SP ASAP. In other words this could be bundled with next SP release when maintainers decide to do a release.

/cc: @mikeminutillo

PR #405, opened 2017-02-17
Fix tags

timbussmann:
VS shows me the following two changes as errors. Was this maybe intended?

since my whitespace plugin causes some noise, I'll highlight the relevant changes with comments.

### DIFF
--- a/src/ServicePulse.Host/app/css/particular.css
+++ b/src/ServicePulse.Host/app/css/particular.css
@@ -501,7 +501,7 @@ h1 {
     margin: 0 0 32px;
 }
 
-p.metadata { 
+p.metadata {
     margin-bottom: 6px;
 }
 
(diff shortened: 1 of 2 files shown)


src/ServicePulse.Host/app/css/particular.css
```diff
@@ -973,7 +973,7 @@ p.lead hard-wrap.ng-binding {
     padding-left: 0;
 }
 
-..box-no-interaction {
+.box-no-interaction {
```
timbussmann: note the `..` in the previous version

src/ServicePulse.Host/app/js/views/failed_groups/view.html
```diff
@@ -157,6 +157,7 @@
                                                                         </div>
                                                                     </div>
                                                                 </div>
+                                                            </div>
```
timbussmann: added "missing" div (and removed the one on Line 202.

WojcikMike:
LGTM

Issue #408, opened 2017-02-22
JavaScript error when using IE 11


timbussmann:
When looking at the FailedMessages -> All Failed Messages view with Internet Explorer 11, the rendering doesn't seem to work:

![image](https://cloud.githubusercontent.com/assets/3524870/23200123/e64ed3ca-f8d2-11e6-9596-fa7e31f2631f.png)

It seems this is caused by a JavaScript error:

```
TypeError: Object doesn't support property or method 'startsWith'
   at $scope.isActive (http://localhost:8081/js/app.controller.js:26:13)
   at fn0 (Function code:2:494)
   at expressionInputWatch (http://localhost:8081/lib/angular/angular.js:14191:11)
   at Scope.prototype.$digest (http://localhost:8081/lib/angular/angular.js:15707:21)
   at Scope.prototype.$apply (http://localhost:8081/lib/angular/angular.js:15986:13)
   at done (http://localhost:8081/lib/angular/angular.js:10511:36)
   at completeRequest (http://localhost:8081/lib/angular/angular.js:10683:7)
   at requestLoaded (http://localhost:8081/lib/angular/angular.js:10624:9) undefined
```

Other browsers seem to work fine.

timbussmann:
actually, the mentioned error isn't responsible for failing to render the angular app properly on that view, but another error which is also caused by missing ES2015 support from IE11.

Related to this issue, I located 3 ES2015 usages which cause problems when using IE11:
* 2x function lambdas
  * Those can be fixes easily by replacing the lambdas with a function declaration
* 1x startsWith
  * This can be replaced by a polyfill as described here: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith or a more robust version e.g. this one: https://github.com/mathiasbynens/String.prototype.startsWith

not sure whether there are further ES2015 usages in the codebase though.

WojcikMike:
@timbussmann good catch, from what I checked this are the only arrow functions (=>)in the source code that we have and we should eliminate them to support IE 11. Do you want to raise a PR for that?

timbussmann:
@WojcikMike raised PR #409

PR #409, opened 2017-02-22
Fix IE11 compatibility

timbussmann:
fixes https://github.com/Particular/ServicePulse/issues/408
The minimal changeset required to make SP work from IE 11

It might be even easier to just use `substr` instead of `startsWith` to remove the polyfill again. Thoughts @Particular/servicepulse-maintainers ?


WojcikMike:
Thanks @timbussmann !

Issue #413, opened 2017-02-24
Dismiss button is not obvious, change its design and location to make it a button


ramonsmits:
I can't really say that the size and position of the Dismiss button make it a button. Took me a while to realize that is a button to dismiss that box.

![image](https://cloud.githubusercontent.com/assets/152998/23299534/29831934-fa82-11e6-81ca-a1f286ffcc38.png)

Wouldn't it be better if it is similar to this? Combined with an icon?

![image](https://cloud.githubusercontent.com/assets/152998/23299538/2ceddf78-fa82-11e6-97fc-030417eff849.png)

sergioc:
This issue would need additional user testing for a proper validation - best done in combination with other tasks.

Not all buttons need an icon and it certainly shouldn't be a requirement to make a button.

For the time being, I propose enlarging the button a bit to make it more prominent - #415

sergioc:
https://github.com/Particular/ServicePulse/pull/415 released, closing.

PR #415, opened 2017-02-24
Enlarge group retry dismiss button

sergioc:
Connects to #413


WojcikMike:
This change made the button bigger, like this: 
image

sergioc:
Also increased white space left to the button.

PR #417, opened 2017-02-24
Fix debug in si button

### LINKED ISSUE #416: View in ServiceInsight button is not working
When opening message page (/failed-messages/message/id) clicking on button: View in ServiceInsight makes no effect. It should open ServiceInsight with the details of that message.

WojcikMike:
fixes #416

cc: @Particular/servicepulse-maintainers please review and merge


gbiellem:
@WojcikMike  I [commented]( https://github.com/Particular/ServicePulse/issues/416#issuecomment-282696513)  on the issue but should probably have done so here.  If this isn't working it's the fault of the ServiceInsight installer.  It relies on the URL protocol handler for ServiceInsight being registered.  That only happens during the ServiceInsight MSI install.

PR #422, opened 2017-03-03
Adding column description to configuration screen

### LINKED ISSUE #392: Configuration screen is unclear
The configuration screen contains a list of endpoints with an On/Off toggle. It does not state what those On/Off toggles actually turn on or off. 

This is changing whether an endpoint is monitored but that isn't very clear. Maybe we should add some text explaining what the buttons do. 

---

See [this google group discussion](https://groups.google.com/forum/#!topic/particularsoftware/x9j3U4qjgb4) for evidence of this lack of clarity.

WojcikMike:
fixes #392 

Adding labels to make it clear what toggle means. @sergioc please have a look from a UI perspective

### DIFF
--- a/src/ServicePulse.Host/app/js/directives/ui.particular.configurationTabs.tpl.html
+++ b/src/ServicePulse.Host/app/js/directives/ui.particular.configurationTabs.tpl.html
@@ -1,7 +1,7 @@
 ﻿<div class="row">
     <div class="col-sm-12">
         <div class="tabs">
-            <h5 ng-class="{ active: isActive('/configuration/endpoints') }"><a href="#/configuration/endpoints">Endpoints ({{counters.endpoints | number}})</a></h5>
+            <h5 ng-class="{ active: isActive('/configuration/endpoints') }"><a href="#/configuration/endpoints">Endpoints monitoring ({{counters.endpoints | number}})</a></h5>
             <h5 ng-class="{ active: isActive('/configuration/redirects') }"><a href="#/configuration/redirects">Retry redirects ({{counters.redirects | number}})</a></h5>
         </div>
     </div>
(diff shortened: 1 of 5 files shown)


sergioc:
@WojcikMike I'm not sure it adds much benefit. I think a more effective solution is to just rename the view's tab name from "Endpoints" to "Endpoints monitoring". That'll help infer that the purpose of the toggles is to turn monitoring ON or OFF.

WojcikMike:
I think that renaming the page is a step in the right direction, but I don't think it solves the problem. What if we would also added tooltip to the toggle?

sergioc:
I didn't mean renaming the page, only the tab label. Anyway, adding those column headers will at most add some redundancy, so I can live with that without testing. 

I've updated the PR.

WojcikMike:
LGTM to be merged. I tested it in chrome and IE and looks good.

WojcikMike:
@WilliamBZA can you have a look and merge if you are happy with it

src/ServicePulse.Host/app/js/directives/ui.particular.configurationTabs.tpl.html
```diff
@@ -1,7 +1,7 @@
 ﻿<div class="row">
     <div class="col-sm-12">
         <div class="tabs">
-            <h5 ng-class="{ active: isActive('/configuration/endpoints') }"><a href="#/configuration/endpoints">Endpoints ({{counters.endpoints | number}})</a></h5>
+            <h5 ng-class="{ active: isActive('/configuration/endpoints') }"><a href="#/configuration/endpoints">Endpoints monitoring ({{counters.endpoints | number}})</a></h5>
```
WilliamBZA: `Endpoints monitoring` sounds strange, i feel like the tense is perhaps wrong.

What do you think about `Monitored endpoints`? @sergioc

sergioc: :+1:

WojcikMike: @WilliamBZA @sergioc change introduced

PR #428, opened 2017-03-20
Create a view for tracking archive process status

WojcikMike:
Connects to Particular/PlatformDevelopment#1164

Creating a view for displaying archiving status

Ping @WilliamBZA


WojcikMike:
@Particular/servicepulse-maintainers please review as this is ready. To run it you need to use: https://github.com/Particular/ServiceControl/pull/934

sergioc:
Nice @WojcikMike. Some feedback:

- Why isn't there a dismiss action to make it consistent with the behavior of the group retry operation?
- Once the progress view (the black container) disappears the group doesn't disappear along right away. It takes a few seconds (2-3 secs). I'd expect it to disappear immediately.

WilliamBZA:
@sergioc re the disappearing: it depends on the Raven indexes 😞 

We'd have to implement some custom delay type code to prevent it from still staying. Getting into the realm of really hacky then though.

sergioc:
@WilliamBZA you could set an event to fire when the operation is complete (or better, dismissed) and hide it with CSS `display: none;` (via a adding a class with Angular, perhaps).

WilliamBZA:
@sergioc The removal is fired from signalr when the operation has completed, but the page reload timer that kicks off reloads the data from SC. If the indexes aren't up to date yet SC will report that the group is still there, except it's not currently being archived. So SP then shows the group. When the interval kicks off again, raven might have updated the indexes and would then remove the group.

Dismissing might be a workaround for this as it would hopefully give raven enough time to catch up. Unless the user is quick on the dismiss, and then the problem might still happen.

sergioc:
Could you insert a delay before the UI indicates the operation is complete and can be dismissed?

WilliamBZA:
@sergioc Yes, but how long should that delay be? The indexes could be stale forever. I think the best option is the Dismiss. That will give a reasonable amount of time between the completion and the refresh to give raven time to catch up.

sergioc:
The user can also dismiss right away and group remains there afterwards. I'd do a combination. Maybe a 10sec. delay + dismiss. Better than nothing.

WilliamBZA:
Ping some more @Particular/servicecontrol-maintainers to review please. If no comments are made by Wednesday morning, we will merge and release.

PR #430, opened 2017-03-24
Require dismissal of archive operations

WilliamBZA:
Added dismissal code @WojcikMike.

Check if you're happy and you and merge into your branch

### DIFF
--- a/src/ServicePulse.Host/app/js/services/services.service-control.js
+++ b/src/ServicePulse.Host/app/js/services/services.service-control.js
@@ -221,6 +221,13 @@
                 });
         }
 
+        function acknowledgeArchiveGroup(groupId, successText, failureText) {
+            var url = uri.join(scConfig.service_control_url, 'recoverability', 'groups', 'unacknowledgedgroups', groupId);
+            return $http.delete(url).error(function () {
+                notifications.pushForCurrentRoute('Archive messages failed', 'danger');
+            });
+        }
+
         function acknowledgeGroup(id, successText, failureText) {
             var url = uri.join(scConfig.service_control_url, 'recoverability', 'unacknowledgedgroups', id);
             return $http.delete(url).error( function () {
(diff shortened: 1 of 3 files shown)


WilliamBZA:
Also added an extra state

WilliamBZA:
Also fixed the status check in the archive in progress function

src/ServicePulse.Host/app/js/services/services.service-control.js
```diff
@@ -221,6 +221,13 @@
                 });
         }
 
+        function acknowledgeArchiveGroup(groupId, successText, failureText) {
```
WojcikMike: from what I see successText and failureText is not being used in this code. Please remove those parameters

src/ServicePulse.Host/app/js/views/failed_groups/controller.js
```diff
@@ -123,8 +123,8 @@
             return group.workflow_state.status !== 'none' && (group.workflow_state.status !== 'completed' || group.need_user_acknowledgement === true) && !vm.isBeingArchived(group.workflow_state.status);
         };
 
-        vm.isBeingArchived = function (status) {
-            return status === "archivestarted" || status === "archiveprogressing" || status === "archivefinalizing" || status === "archivecompleted";
+        vm.isBeingArchived = function (group) {
```
WojcikMike: This fix require some more work. 
 * There are at least 2-3 places where we still call isBeingArchived passing status not a  group (easy fix)
 * `operation_status` that you check contains status that is not to lower, so either tha names need to be adjusted or tolower used (easy fix)
 * `operation_status` is not being updated by our event handlers. So only when the list is refreshed the state will be actual. So we should update operation_status field whenever an event is handled.

WilliamBZA:
Changes made @WojcikMike

src/ServicePulse.Host/app/js/views/failed_groups/controller.js
```diff
@@ -124,7 +122,7 @@
         };
 
         vm.isBeingArchived = function (group) {
```
WojcikMike: The parameter is named group, but you use status in the method ;)

WilliamBZA: You wrote the method, I'm just using it ;)

WilliamBZA: Nevermind, I'm dumb haha

Issue #387, opened 2016-12-09
Revise display of badge counts next to each tab in Failed Messages


sergioc:
## Who's affected

* All users

## Description

With the [recent introduction of extra failed message group views](https://github.com/Particular/ServicePulse/releases/tag/1.6.9) (by "message type" in addition to the existing "by exception and stacktrace"), the count showing next to the `Failed message groups` tab is not clear anymore, because the number of groups based on "exception and stacktrace" maybe different than the number of groups based on "message type"

Another issue is that the count next to "Failed message groups" will change depending on which group view is active.

I've discussed the problem with @WojcikMike @janovesk and we considered the following options:

1. Get rid of the count - one could argue this is OK because users land on Failed groups by default and can see there's groups. However, it wouldn't be consistent with the rest of the UI because it this tab view contains items.
2. Sum the number of groups - not really optimal because it may give the impression that a given message is present in only one type of group, but in reality it's always part of two groups
3. Split each group type into a separate tab - not scalable in the UI if we add more group types. Also creates a similar problem as 2. if you sum all group types' counts.
4. Show an empty badge - the best option because even though it has no number, it should alert the user that there's item inside the tab requiring attention.

## Proposed UI changes

- Adopt the red count badge for counting items next to each Failed Messages tab. This will make counts more uniform across the UI and allow more flexibility when dealing with count inconsistencies since we can drop the number
- Do not display the badge if any tab in Failed messages is empty
- Show an empty badge for Failed groups if there are any failed groups
- Add a tooltip to Failed Groups badge - "There's varying numbers of failed message groups depending on group type"

Old:

![2016-12-09 at 10 18](https://cloud.githubusercontent.com/assets/493451/21047502/89e5a198-be0a-11e6-9b9b-25ee26598807.png)

New:

![2016-12-09 at 12 24](https://dl.dropboxusercontent.com/s/p8bwmq87fxdgkhn/2016-12-09%20at%2012.27.png)

WilliamBZA:
This has been implemented and will be released as part of 1.9.2

PR #442, opened 2017-05-10
Revised style of counts for failed msg tabs

sergioc:
Connects to #387

### DIFF
--- a/src/ServicePulse.Host/app/css/particular.css
+++ b/src/ServicePulse.Host/app/css/particular.css
@@ -1314,6 +1314,11 @@ span.metadata.danger, i.fa.fa-trash-o.danger, sp-moment.danger {
     margin: 12px 0 0;
 }
 
+.tabs .badge {
+    position: relative;
+    top: -2px;
+}
+
 /* RESPONSIVE TWEAKS */
 
 @media only screen and (min-width: 993px) {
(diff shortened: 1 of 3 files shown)


sergioc:
@WojcikMike given the incongruity between counts for each failed group type, the count next to the "Failed groups" tab should be `!` whenever there's at least 1 failed group. Can you help with showing that instead of a number?

WojcikMike:
@sergioc have a look. I made the ! to be dependent on a number of error messages. As if there are any messages than certain group classification can return a different number. Let me know what you think.

sergioc:
Looks good. I didn't understand your comment fully. To confirm, it will always use `!`, right?

I'm also wondering whether we could indicate the no. of groups next to the name of each failed message group type in the dropdown. Like so:

2017-05-30 at 16 42


This would make it easier to correlate the group type counts with the `!`.

WojcikMike:
@Particular/servicecontrol-maintainers for this UI change to be made an api to get group counts would have to be created (ideally one that takes classification). Without it, it would be hard to implement this

sergioc:
Then let's split group type counts into a separate issue

sergioc:
OK, new issue created: https://github.com/Particular/ServicePulse/issues/454

@Particular/servicecontrol-maintainers please review and merge this one.

PR #446, opened 2017-06-07
Launch Wave Metrics Feature Branch

tmasternak:


src/ServicePulse.Host/ServicePulse.Host.csproj
```diff
@@ -227,6 +227,8 @@
     <Content Include="app\js\directives\ui.particular.tabset.tpl.html" />
     <Content Include="app\js\views\endpoints\endpoints.controller.js" />
     <Content Include="app\js\views\endpoints\endpoints.route.js" />
+    <Content Include="app\js\views\monitored-endpoints\monitored-endpoints.controller.js" />
```
WojcikMike: @WilliamBZA in every other spot we use _ not - for directories. Was it intentional or accidental? I am fine with changing it in all places to - if we think it is better.

WilliamBZA: @WojcikMike Not intentional at all. I'll change to `_`

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
```diff
@@ -0,0 +1,37 @@
...
+
+        $scope.$on('$destroy', function () {
+            $timeout.cancel(timeoutId);
+        });
+
+        function updateUI() {
+            monitoringService.getRaw().then(function (data) {
+                $scope.endpoints = data["NServiceBus.Endpoints"];
+
+                timeoutId = $timeout(function () {
```
WilliamBZA: 🤢 

Rather than starting off on a timer, I'd rather go with a signalr push after the initial render staright away. The longer we leave a timeout in the less likely it will be that we'll change it. And future whoever-is-going-to-work-on-this is going to curse us for it.

Also, should use $interval instead of $timeout for repeated events that should be triggered.

E.g.
```JavaScript
function updateUI() {
    monitoringService.getRaw().then(function(data) {
        $scope.endpoints = data['NServiceBus.Endpoints'];
    });
}

$interval(updateUI, 5000);
```

PR #449, opened 2017-06-09
Change attribute to element

WilliamBZA:
Changes:

1. Changed attribute directive to element directive
2. Load data points from binding instead of hardcoded in directive


timbussmann:
looks good to me, definitely something we would have to do anyway in case we want to move ahead with the directive approach 👍 

Merging this as it targets my PR.

@WilliamBZA sidennote: For performance reasons, I was considering to feed the graphs from data which isn't on the same scope-property as the data which generates the endpoints table, as this will cause a full recalculation of the whole table on every update of the metrics data. Instead there could be two separated data properties, one which is used to generate the table and one which is then referenced by the graphs do render the data. As that approach would be more complex and it doesn't look like the current implementation causes huge performance issues, I'm good to leave this for now and continue with the current approach.

PR #450, opened 2017-06-13
Call and combine multiple endpoints

WilliamBZA:

Uses RxJs to create an observable of various http requests and merge the results of those requests into a subscribable that can be used to update the UI.

Changes the binding to be dictionary based instead of array based so that we can update each endpoint uniquely instead of having to update all the results from each monitoring service.

Note that the graphs don't actually render now because the data format is different. I'll fix that next.

### POA:
1. Fix UI to show warnings if a node is only in SC, but not in the monitoring instances
2. Same thing, but vice versa
3. Investigate loading statueses for each source (this is difficult because we don't know if an endpoint will be part of a result set or not)
4. Verify that error handling code works as expected
5. Verify that there are no (or at least limited) unecessary rebinds to data which would cause re-rendeirings
6. Add filtering by time period on the screen


src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
```diff
@@ -13,12 +13,9 @@
...
 
         function updateUI() {
-            monitoringService.getRaw().then(function (data) {
-                $scope.endpoints = data["NServiceBus.Endpoints"];
-
-                timeoutId = $timeout(function () {
-                    updateUI();
-                }, 5000);
+            monitoringService.getEndpoints().subscribe(function (endpoint) {
+                $scope.endpoints = $scope.endpoints || {};
```
timbussmann: what is this for? Does this need to be in the subscription callback?

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
```diff
@@ -13,12 +13,9 @@
...
+                $scope.endpoints[endpoint.endpointName] = endpoint;
```
timbussmann: wouldn't this trigger a re-rendering for each endpoint?

WilliamBZA: Correct. Unless we change the api to return data for a specific endpoint instead of a list of endpoints + data, we'll have to redraw the entire endpoint (including all data) when retrieved.

Unless of course we do some clever hackery around "last updated timestamp" or whatever.

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
WilliamBZA: No, guess it can get moved out.

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
timbussmann: not too familiar with rxJs, but pretty sure in rx.net you could merge different streams into a single result. so instead of producing from:
[a1, a2, a3] , [b1, b2] -> [a1, a2, a3, b1, b2]
you would produce
[a1, a2, a3], [b1, b2] -> [[a1, a2, a3, b1, b2]]
this would still have the disadvantage that you trigger complete re-rendering on each http request, but clearly less than with the current approach.

WilliamBZA: That's exactly what this does.

Each request results in multiple endpoints, each endpoint is passed as a result to the subscribers. Each endpoint is rebound if it already exists, and is added to the object if it doesn't yet exist.

To be clear: and endpoint is a endpoint that is being monitored, not a monitoring service instance. So the graphs and label would be re-bound as it's updated.

WilliamBZA: The map function in the service maps the array of URLs into http GETs, which are wrapped in observables each to be repeated every 5 seconds (and auto retried), and then merged into a single stream which pipes the individual results as a merged stream to this code

src/ServicePulse.Host/app/js/app.constants.js
```diff
@@ -5,7 +5,7 @@
         .constant('showPendingRetry', false)
         .constant('scConfig', {
             service_control_url: 'http://localhost:33333/api/',
-            monitoring_url: 'http://localhost:1234/metrics/'
+            monitoring_urls: ['http://localhost:1234/metrics/', 'http://localhost:1235/metrics', 'http://localhost:1236/metrics']
```
WojcikMike: Do we want to push this list url's Or just leave an array of 1 metric url?

src/ServicePulse.Host/app/js/services/services.monitoring.js
```diff
@@ -2,27 +2,36 @@
...
+        function getEndpoints() {
+            return rx.Observable.merge(scConfig.monitoring_urls.map(function (url) {
+                var requestUri = uri.join(url, '/raw');
+
+                var httpRequest = rx.Observable.just(url)
+                    .flatMap(function (requestUrl) {
+                        return $http.get(requestUrl);
+                    });
+
+                var repeatRequests = rx.Observable.interval(5000).flatMapLatest(httpRequest.retry().delay(5000));
```
WojcikMike: Am I understanding this correctly that we will retry the call indefinitely? Do we want to log it somehow when the call has failed?

src/ServicePulse.Host/app/js/app.constants.js
WilliamBZA: This is maintained by the installer and is read only, once the array is writted it's only updated with a full restart. No need for any client-side changes.

src/ServicePulse.Host/app/js/services/services.monitoring.js
WilliamBZA: It still shows the error in the console. Where else should we log it?

WilliamBZA: The `retry()` just means if it fails, try again in another 5 seconds. It's actually a real pain to get the continual retry stuff working properly. I suspect it's not working correctly at the moment anyway.

Scenarios we need to support:

* Service is configured but is restarting or temporarily unavailable
    * This should auto-retry until connected again
* Service is configured but not available
    * Should probably show a notification to the user?
    * Same for incorrectly configured service

So thinking about it while typing this, there's only 1 viable use case for the retries, and those should probably be limited to, say, 10 retries max before stopping?

src/ServicePulse.Host/app/js/app.constants.js
WojcikMike: What I am saying is that you configured it to hold, 3 values. I am asking if we should push it with those 3 values. As every dev environment will use those values.

src/ServicePulse.Host/app/js/services/services.monitoring.js
WojcikMike: 👍 on my end

src/ServicePulse.Host/app/js/app.constants.js
WilliamBZA: Ah ok, I follow you now. I'll change back to one for the PR.

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
timbussmann: > Each request results in multiple endpoints, each endpoint is passed as a result to the subscribers. 

I'm suggesting to not pass each endpoint individually but to pass the whole set of endpoints as a single value (therefore `[[ .. ], ..]`).

timbussmann:
btgw. why did the angular.js files change?

WilliamBZA:
Angular had to be updated to support rxjs

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
WilliamBZA: I much prefer passing each endpoint individually. The consuming UI can then evaluate each response on the merits of that specific endpoint without having to know which service it came from. This way the controller doesn't need to know anything about the number of monitoring services involved, which should keep the code simpler. We can also in the future use the same observable to listen for updates to specific endpoints (maybe use the RxJs `filter` method?) and only send those updates through to the controller instead of sending all endpoints to the controller again.

The alternative is to have code in the UI controller to iterate the list of endpoints returned and then only evaluate those endpoints. What benefit are you seeing of doing it that way?

To change it to pass the arrays though it's as simple as removing the `selectMany` from the `getEndpoints` method, so it's completely trivial to achieve.

timbussmann: @WilliamBZA I feel we're talking past each other. I fully agree on hiding the source of the endpoint information, this doesn't change the fact that I propose to not push endpoint by endpoint, but a merged (concat) single value (an array of endpoints) instead. Or alternatively, to decouple the updated values from changing the scope values directly to prevent excessive re-rendering.

timbussmann:
personally, I have a hard time to see the value rxjs is bringing to the table here. If that aligns well with the pulse maintainers vision, that's alright for me though.

WilliamBZA:
RxJs is providing an easy way to both combing multiple sources of data into a single subscription, but it also allows for switching to push based changes easily at a later stage. I think it's a must for if we want to support data coming from multiple sources.

Speaking as a pulse maintainer, including it will also allow us to unify the pages where polling and pushing are mixed together. Those pages are a painful experience to work on and I've been evaluating RxJs as a way to help make those pages easier to work with.

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.controller.js
WilliamBZA: Should we wait for all requests to complete then before piping the results to the controller? That's the logical conclusion of using concat. What happens when one service responds in 5ms but the second service only responds in 50 seconds?

Further more, the filtering logic for "are there any changes to this endpoint that should result in a re-bind or re-render" would have to sit in the controller, and not the service. There's a lot of benefit to treating each endpoint as a unit in the controller vs a bulk.

timbussmann:
I still think that it might make more sense to not "flatten" the http request result but to combine them into a single array as this is easier for angular to process. E.g. something like this:

        var scData = rx.Observable
            .interval(5000)
            .flatMap(() => {
                return Rx.Observable
                    .fromPromise(uri.join(scConfig.service_control_url, 'recoverability', 'endpoints'))
                    .map(endpoints => {
                        //map
                    });
            });

        var monitoringData = rx.Observable
            .interval(5000)
            .flatMap(() => {
                return Rx.Observable.fromPromise(uri.join(url, '/data'));
            });

        var endpoints = scData.zip(monitoringData).map(results => {
            // convert to desired format
        });

        // optionally avoid duplicate requests:
        var endpointsPublished = endpoints.publish();
        endpointsPublished.connect();

src/ServicePulse.Host.Tests/SpecsRunner.html
```diff
@@ -28,6 +28,8 @@
     <script src="../servicepulse.host/app/lib/moment.min.js"></script>
     <script src="../servicepulse.host/app/lib/angular/angular-momentjs.js"></script>
     <script src="../servicepulse.host/app/lib/angular-ui-select/select.js"></script>
+    <script src="../servicepulse.host/app/lib/rx.angular/rx.all.js"></script>
```
timbussmann: is this code even using anything from the rx.angular package?

WilliamBZA: Yep, it's the injection of the `rx` object into the service

timbussmann: @WilliamBZA that seems to be a trivial thing which we could do ourselves which would bring us the benefit of referencing the rx libs directly (and have a better versioning)

src/ServicePulse.Host.Tests/tests/js/services/services.monitoring.spec.js
```diff
@@ -42,58 +42,55 @@
...
+    it('should return added collections when sc returns matching rows', function () {
         $httpBackend.whenGET('http://localhost:1234/diagrams/data').respond(monitoredEndpointWithData);
         $httpBackend.whenGET('http://localhost:33333/api/recoverability/endpoints').respond(scEndpointWithMatchingData);
 
         var monitoredEndpoints;
-        monitoringService.getData().then(function (response) {
-            monitoredEndpoints = response.data;
+        monitoringService.getEndpoints().subscribe(function (response) {
+            monitoredEndpoints = response;
+            if (monitoredEndpoints.IsFromSC) {
```
timbussmann: seeing this logic https://github.com/Particular/ServicePulse/pull/450/files#diff-db4ff24b7b6be144aa01f62db6bf3a2bR64 I'm even more convinced that we should not treat it as a single stream of data

src/ServicePulse.Host.Tests/tests/js/services/services.monitoring.spec.js
```diff
@@ -42,58 +42,55 @@
         $httpBackend.whenGET('http://localhost:33333/api/recoverability/endpoints').respond([]);
 
         var monitoredEndpoints;
-        monitoringService.getData().then(function (response) {
-            monitoredEndpoints = response.data;
+        monitoringService.getEndpoints().subscribe(function (response) {
+            monitoredEndpoints = response;
```
timbussmann: that naming looks off. Also, is that assignment required at all?

src/ServicePulse.Host/app/js/services/services.monitoring.js
```diff
@@ -2,74 +2,65 @@
...
-                    .catch(function(fallback) {
-                        notifications.pushSticky('Can not connect to Monitoring Service');
-                    });
-            });
-            return outcome;
-        }
 
-        var previousExceptionGroupEtag;
+                        return request;
+                    }).retryWhen(automaticRetry);
```
timbussmann: do we really need a retry mechanism here?

WilliamBZA: Given that the monitoring services could be running in different regions, I'd say it's a good thing to have. What are your specific concerns around it?

src/ServicePulse.Host.Tests/SpecsRunner.html
WilliamBZA: :+1:

PR #451, opened 2017-06-14
Adding npm bower

WojcikMike:

Adding npm and bower support.

To prep the solution one needs to run 
`npm run setup` in the ServicePulse.Host directory
Or simply
npm install
bower install
node wiredep.js

Wiredep make sure that index.html is updated with bower dependecies. So one doesn't need to add it manually. 

This PR requires:
 - documenting the build process
 - Investigating if we can use npm, bower in visual studio to ease the experience for some people
 - Probably invest some time in Team City to install node on build agents, and update build steps. We could also commit js/css files in which case team city doesn't have to have node installed.

@WilliamBZA please have a look


WojcikMike:
At the moment, there is one bug, that I need to solve:

Uncaught TypeError: container.height is not a function
    at handler (ng-infinite-scroll.js:28)
    at eventHandler (angular.js:3293)

WojcikMike:
So I fixed the bug, It was related to order of dependecies. JQuery has to be referenced before angularjs so that JQLite is not used. 

With regards to testing in release. I tested it and everything seems to work. I extracted files from exe and whole bower_components are there with all the files. Which is probably a bit redundant. But might be ok for now.

The next step is to update build agents to have node installed on them

WojcikMike:
I added a step to Team City to run npm install, bower install and it works correctly the packages end up in the exe file. The step for now is disabled, until this PR get merged.

The step is run using cmd step, which is not ideal as it never fails the build. I will ask to install node plugin to team city

WojcikMike:
@WilliamBZA Nodejs is isntalled on the builder. I've configured the step and disabled it so that builds without npm packages build correctly. For me this is ready to be merged. And then we need to rebase all ongoing PR's

WojcikMike:
@Particular/servicepulse-maintainers can we merge this?

PR #460, opened 2017-06-27
Only load monitoring data from monitoring overview

timbussmann:
targets #450 

based on todays call with @WojcikMike @Scooletz and @sergioc, this code changes the service to not load and aggregate data from SC for now and limit itself to data from the monitoring server.

thoughts?

### DIFF
--- a/src/ServicePulse.Host/app/js/services/services.monitoring.js
+++ b/src/ServicePulse.Host/app/js/services/services.monitoring.js
@@ -4,11 +4,11 @@
 
     function Service($http, rx, scConfig, uri, $q) {
 
-        var mappedUrls = scConfig.monitoring_urls.map(function (url) {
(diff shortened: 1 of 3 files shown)


src/ServicePulse.Host/app/js/services/services.monitoring.js
```diff
@@ -3,63 +3,43 @@
...
-                return streamedResults.selectMany(function (endpoints) {
-                    return endpoints;
+        var mappedUrls;
+        var source = Rx.Observable.create(function (observer) {
+            mappedUrls = scConfig.monitoring_urls.map(function (url) {
+                return uri.join(url, '/data');
+            });
+
+            updateData(observer);
+            var interval = setInterval(function() { updateData(observer); },
```
WilliamBZA: 😢

PR #463, opened 2017-06-29
Optimize graph y-axis range

timbussmann:
* Min value is always 0 based to not cut away areas
* Max value is designed so that the average bar will always show up in case the current data points are smaller than the average (not an issue with the current implementation which calculates average based on the data points, but that will probably change)

the following graph shows how different data points & average values would look like with the current (left) implementation and the proposed changes (right). Each row contains the exact same data on both sides.

image

### DIFF
--- a/src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.module.js
+++ b/src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.module.js
@@ -17,10 +17,9 @@
                         var width = 130;
                         var margin = 2;
                         var average = d3.mean(scope.data);
-                        var max = d3.max(scope.data);
-                        var min = d3.min(scope.data) !== d3.max(scope.data) ? d3.min(scope.data) : 0;
+                        var max = Math.max(average * 1.5, d3.max(scope.data));
                         var scaleY = d3.scaleLinear()
-                            .domain([min, max])
+                            .domain([0, max])
                             .range([heigth - margin, margin]);
 
                         var scaleX = d3.scaleLinear()


WilliamBZA:
What was the data used for each? I assume the big changes int he last few graphs is because of the min value?

timbussmann:
@WilliamBZA here's the sample data used (last value is the average):

        drawGraphs([10, 15, 30, 50, 20, 10, 10, 15, 0], 25);
        drawGraphs([10, 15, 30, 50, 20, 10, 10, 15, 0], 35);
        drawGraphs([10, 15, 30, 50, 20, 10, 10, 15, 0], 15);
        drawGraphs([10, 15, 30, 50, 20, 10, 10, 15, 0], 200);
        drawGraphs([10, 15, 30, 50, 20, 10, 10, 15, 0], 5);
        drawGraphs([1000, 1100, 1200, 1010, 990, 1000, 1010, 1015, 1000], 1000);
        drawGraphs([1000, 1100, 1200, 1010, 990, 1000, 1800, 1015, 1000], 1000);

yeah the changes in the last graphs came from the min value being quite high (990) and therefore the graph only showing a range from 990 to 1200, instead from 0 to 1200.

src/ServicePulse.Host/app/js/views/monitored-endpoints/monitored-endpoints.module.js
```diff
@@ -17,10 +17,9 @@
                         var width = 130;
                         var margin = 2;
                         var average = d3.mean(scope.data);
-                        var max = d3.max(scope.data);
-                        var min = d3.min(scope.data) !== d3.max(scope.data) ? d3.min(scope.data) : 0;
+                        var max = Math.max(average * 1.5, d3.max(scope.data));
```
WojcikMike: It is rather minor but i wonder if as a max we shouldn't just get (d3.max(scope.data) * 1.1). So we take the maximum value and add 10% to it. Which means that highest point in the diagram will still have a bit of space to the top of the chart. But we can decide on that at later stage

PR #464, opened 2017-06-29
Rxjs directive

WilliamBZA:
* Changes our packages to the bower packages
* Removes rx.angular and uses a custom factory (with no error handling)


src/ServicePulse.Host/app/js/services/factory.rxjs.js
```diff
@@ -0,0 +1,8 @@
+; (function (window, angular, undefined) {
+    'use strict';
+
+    angular.module('sc')
+        .service('rx', function () {
```
timbussmann: why not use `constant` ?

src/ServicePulse.Host/ServicePulse.Host.csproj
```diff
@@ -101,6 +101,7 @@
     <Content Include="app\js\directives\ui.particular.failedMessageTabs.tpl.html" />
     <Content Include="app\js\directives\ui.particular.nodata.js" />
     <Content Include="app\js\directives\ui.particular.nodata.tpl.html" />
+    <Content Include="app\js\services\factory.rxjs.js" />
```
timbussmann: tbh I don't see any value in having a dedicated file for this, especially since I'm not aware of any minification steps.

src/ServicePulse.Host/app/js/services/factory.rxjs.js
WojcikMike: why not use `factory` ? For now it is so simple that it could be constant as @timbussmann mentions. But semantically I read it as factory. I wonder if we envision adding here more code. I am fine with either of the solutions

timbussmann: > But semantically I read it as factory.

how so? `Rx` is a constant provided by rxjs which we define here for property injection reasons. I don't see what type of code you would add here at any time.

WilliamBZA: Good point. In all honesty, we're just paying lipservice to injecting a dependency here. Why not simply use RxJs without declaring it as an angular dependency? We'd run into problems if we were using AMD, but we have no modules at all.

WilliamBZA:
Updated @WojcikMike, @timbussmann

src/ServicePulse.Host/app/js/services/factory.rxjs.js
timbussmann: fine with that approach as well.

timbussmann:
@WilliamBZA changes look good, can you have a look at the tests?

WilliamBZA:
Tests passing

PR #466, opened 2017-07-04
Remove unused columns

timbussmann:


WilliamBZA:
What ☝️ said

timbussmann:
> According to Bootstrap Grid the sum of columns in a row should be 12

@WojcikMike do you have a source on that statement? Bootstraps supports up to 12 columns, but I've never seen that it is somewhat necessary to fully use them. In fact, for nested columns, this is what the bootstrap docs state:

>  a set of columns that add up to 12 or fewer (it is not required that you use all 12 available columns).

as long as we don't want to use the extra space available by removal of the two columns, I don't see a need to change the layout config?

making the columns wider kinda looks weird on max width setting:
image
but okay on smaller:
image
while going with a width of 2 works better on max width:
image
but not on smaller:
image

sergioc:
I'd say the 2nd or 3rd options look best for now. How about leaving it crude just for now and later tweak the presentation in a separate issue?

timbussmann:
@sergioc there are only two options right now (shown at different screen resolutions):
* wider columns (screen 1 & 2)
* smaller columns (screen 3 & 4)

sergioc:
wider is better then to prevent what happens on screen 4

WilliamBZA:
I guess as long as it's wrapped in a `row` it shouldn't matter that it doesn't add up to 12. 👍 My OCD is twitching though.

timbussmann:
* switched to the wider column width (still doesn't add up to 12, but I think it's not an issue)
* removed the `ms` paragraphs

WojcikMike:
I can live with the numbers not adding to 12, but I don't like it :| My understanding would be that in reality, our columns are not 4, 3 and 3. But 4.66, 3.66, 3.66. But as I said I am fine with it staying how it is.

WojcikMike:
@WilliamBZA do we want to merge it? Or do you still want to review it?

WilliamBZA:
Merged

PR #467, opened 2017-07-04
Show average number

timbussmann:
show the value of the average in the vertical middle on the right side of the graph

image

note that it doesn't contain any trimming logic and just cuts off large values (e.g. second last graph is 8 millions)

ping @WojcikMike @WilliamBZA @sergioc


WojcikMike:
Can we merge this or do we want to wait for @sergioc ?

sergioc:
Just merge, it's good enough for testing.

PR #471, opened 2017-07-07
Show a notification when a metrics endpoint isn't reachable

timbussmann:

shows a toastr notification when a connection to a monitoring server fails. The notification does not fade out until clicked on by the user or when the page is switched (or to be more precise: once no more subscriber is subscribed to the monitoring servers data) or data has been successfully received from that server again.

### DIFF
--- a/src/ServicePulse.Host/app/js/services/service.toast.js
+++ b/src/ServicePulse.Host/app/js/services/service.toast.js
@@ -32,6 +32,14 @@
     ];
 
     angular.module('sc')
-        .service('toastService', service);
+        .service('toastService', service)
+        .factory('toastr', function () {
+            toastr.options = {
+                preventDuplicates: true,
+                positionClass: "toast-bottom-right",
+            };
+
+            return toastr;
+        });
 
 } (window, window.angular));
--- a/src/ServicePulse.Host/app/js/services/services.monitoring.js
+++ b/src/ServicePulse.Host/app/js/services/services.monitoring.js
@@ -2,8 +2,7 @@
 (function (window, angular, $, undefined) {
     'use strict';
 
-    function Service($http, rx, scConfig, uri) {
-        toastr.options.preventDuplicates = true;
+    function Service($http, rx, scConfig, uri, toastr) {
 
         var connectionToasts = [];
         var mappedUrls;
@@ -40,7 +39,7 @@
                         observer.onNext(result.data);
                     }, function (error) {
                         if (connectionToasts[url] === undefined) {
-                            var message = "unable to connect to " + url;
+                            var message = "unable to connect to ServiceControl.Monitoring at: " + url;
                             var x = toastr.warning(message,
                                 '',
                                 {
@@ -60,7 +59,7 @@
         return service;
     }
 
-    Service.$inject = ['$http', 'rx', 'scConfig', 'uri'];
+    Service.$inject = ['$http', 'rx', 'scConfig', 'uri', 'toastr'];
 
     angular.module('services.monitoringService', ['sc'])
         .service('monitoringService', Service);
--- a/src/ServicePulse.Host/app/js/services/services.monitoring.js
+++ b/src/ServicePulse.Host/app/js/services/services.monitoring.js
@@ -4,7 +4,7 @@
 
     function Service($http, rx, scConfig, uri, toastr) {
 
-        var connectionToasts = [];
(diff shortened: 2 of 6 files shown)


timbussmann:
updated the PR to use toastr directly instead of the toaster angular plugin which I have no idea what it does on top of toastr itself, but it doesn't allow to remove specific toasts which makes toaster unusable for this feature.

src/ServicePulse.Host/app/js/services/services.monitoring.js
```diff
@@ -29,7 +42,24 @@
...
                 $http.get(url)
                     .then(function (result) {
+                        if (connectionToasts[url] !== undefined) {
+                            var message = connectionToasts[url];
+                            // doesn't exist
+                            toastr.clear(message);
+                            connectionToasts[url] = undefined;
+                        }
                         observer.onNext(result.data);
+                    }, function (error) {
```
WilliamBZA: I'm not entirely sure this is in the right place. Shouldn't the service merely tell the observers that the error occurred and let the observers decide whether to show a notification or not?

src/ServicePulse.Host/app/js/services/services.monitoring.js
```diff
@@ -16,6 +17,18 @@
 
             return function () {
                 clearInterval(interval);
+
+                // clear open toasts
+                for(var toastEntry in connectionToasts) {
```
WilliamBZA: This seems like a UI specific concern being handled in the service, doesn't seem like it's in the right place

src/ServicePulse.Host/app/js/services/services.monitoring.js
timbussmann: so there are other options which I ruled out:
* call OnError on the endpoints stream: Worst approach. Will cause all subscribers to unsubscribe and will require "reconnect" logic on every subscriber. Additionally it will require a hold time on every subscriber to ensure we're not pulling constantly in case of an error. This massively complicates the rx code and I can tell you from my previous rx experience, you don't ever want to have OnError on a stream which should just reliably push data to several subscribers repeatedly.
* Make the pushed data on the stream an `Either` value which provides either the retrieved data or the error so subscribers can be aware of an error happening. This would work well in case you want to show a view specific error but it doesn't align with the concept of toast notifications which are not view-bound (which subscriber is responsible of raising the notification?). You would require a dedicated subscriber raising the notification which is tricky, because you don't want the notification logic to subscribe the same way as a view because it would start polling data even when no view is interested in data. This approach would work best with view-specific error messages.
* to tackle the last problem, you would have to provide a dedicated error-stream as a hot observable which is completely detached from the actual endpoints data stream for the component which subscribers to service failures and will display them in a view-agnostic way. I think this approach would align best with the current design and the fact we're using toast-notifications.

Given that I was not sure whether we will stick with the toastr notifications or not (view specific logic provides much more flexibility and visualization options), I didn't want to make the choice between option 2/3 and they heavily depend on this design decision.

src/ServicePulse.Host/app/js/services/services.monitoring.js
timbussmann: I wanted to use the notification service which seems to be the design intended for this kind of scenarios, but since it's not working I decided to go with using toastr directly. It is UI specific concern, so is the whole angular application ;) By abstracting toastr away I think you would gain the flexibility to decide where and how to render the information (e.g. a central notification error instead of toastr notifications). That seems to be a fairly easy refactoring once needed (and once the existing toast and notification services are cleaned up)

src/ServicePulse.Host/app/js/services/service.toast.js
```diff
@@ -32,6 +32,14 @@
     ];
 
     angular.module('sc')
-        .service('toastService', service);
+        .service('toastService', service)
```
timbussmann: tbh I don't see value in toastService right now. That would be an easy dependency to get rid of.

timbussmann:
@WilliamBZA @tmasternak please let me know whether you're interested in this PR, otherwise I'll close it.

WilliamBZA:
As I mentioned int he comments, I'm not entirely happy with where the de-dupe logic is sitting and we don't have time at the moment to come up with a proper solution due to how the squad is prioritizing and scoping our work. So I'm going to close, but will leave the branch so that when we get a squad issue for notifications we can reference at this one.

PR #480, opened 2017-07-25
Only show 2 decimal places in graphs

WilliamBZA:


sergioc:
Why do we need decimal places at all?

WilliamBZA:
@ramonsmits raised the issue and the @Particular/launchwave-devopsadvancedmonitoring-squad prioritised it for this iteration. Best to ask them.

ramonsmits:
@sergioc good comment, during the demo we saw values that had 5 or 6 digits.  I think it depends on the value reported. If you report something in seconds, it probably is wise to report 1 or 2 digits if you report in milliseconds you probably do not need any decimal places.

So it depends per metric and the unit of measurement that is reported.

It also depends on the reporting interval. If the interval is 30 seconds and you report the average number of messages per second seeing 0 or 1 isn't really useful but if you see ~0.20/s you know it isn't idle because it got rounded to 0.

Which could also be rendered differently by not reporting in message per second but in messages per hour or per day.

PR #486, opened 2017-08-02
Adding throughput

WojcikMike:
Adding diagram to display throughput


WojcikMike:
We need to work a bit on making the graphs smaller. At the moment 5 graphs don't fit in 1 row, especially when the name of the endpoint is long and averages are big numbers:
image

Details look better as there is 1 less graph:
image

WilliamBZA:
The throughput is intended to be only on the details page anyway, and it's actually a separate graph to the current list there according to the wireframes.

This is fine for now, I'll add a card to move it to a dedicated throughput graph

PR #487, opened 2017-08-02
Show stale data warning icon

WilliamBZA:
Shows a staleness warning if required

### DIFF
--- a/src/ServicePulse.Host/app/css/particular.css
+++ b/src/ServicePulse.Host/app/css/particular.css
@@ -1467,12 +1467,12 @@ span.metadata.danger, i.fa.fa-trash-o.danger, sp-moment.danger {
     width:100%;
 }
 
-.endpointrow {
+.endpoint-row {
     height:60px;
     position:relative;
 }
 
-.endpointrow .graphicon {
+.endpoint-row .graphicon {
     top:50%;
     position:absolute;
     transform:translateY(-50%);
--- a/src/ServicePulse.Host/app/css/particular.css
+++ b/src/ServicePulse.Host/app/css/particular.css
@@ -1458,3 +1458,11 @@ span.metadata.danger, i.fa.fa-trash-o.danger, sp-moment.danger {
         overflow: hidden !important;
     }
 }
+
+.graph {
+    width:80%;
+}
+
+.graph svg {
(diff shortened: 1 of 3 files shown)


sergioc:
I think the easiest thing to do for now is to replace the numeric value next to each graph by `<i class="fa fa-exclamation-triangle danger" aria-hidden="true"></i>`

Not entirely sure if adding the `danger` class will paint the icon in red, but it should be easy in some other ways if that's not the case.

sergioc:
I would also add a danger icon after the endpoint (instance) name to better signal the issue and we need a tooltip for the icons.

WilliamBZA:
The graph is rendered within a `svg`, we can't put HTML elements inside a `svg` except the [defined list](https://developer.mozilla.org/en-US/docs/Web/SVG/Element). So I don't think we can put the icon in place of the numeric value. :sad:

sergioc:
Why render the numeric value inside the SVG instead of directly in HTML?

WilliamBZA:
> Why render the numeric value inside the SVG instead of directly in HTML?

So that it sits next to the line that represents the average

src/ServicePulse.Host/app/css/particular.css
```diff
@@ -1465,4 +1465,15 @@ span.metadata.danger, i.fa.fa-trash-o.danger, sp-moment.danger {
 
 .graph svg {
     width:100%;
+}
+
+.endpointrow {
```
sergioc: Can we agree on a convention for naming HTML classes? I've been separating multiple names in class with a hyphen to make it easier to read.

sergioc: Bootstrap also separates multiple words with an hyphen.

WilliamBZA: I've changed to `endpoint-row`

sergioc:
My last comment about naming classes with multiple words gave one example, but you probably want to apply the convention to all newly introduced classes with multiple names. We can also later do a clean-up PR for this.

WilliamBZA:
Good point @sergioc, but this PR only introduced 1 of those so I agree a later clean-up PR would be better rather than holding this specific one up.

PR #493, opened 2017-08-16
Endpoints overview UI update

sergioc:


sergioc:
Only need @WojcikMike to make the failed messages status icon link to the respective endpoint's failed messages group.

sergioc:
I think it's best to deal with the failed msgs link in its own issue (not really related to the scope of presentation here)

src/ServicePulse.Host/app/js/app.constants.js
```diff
@@ -5,7 +5,7 @@
         .constant('showPendingRetry', false)
         .constant('scConfig', {
             service_control_url: 'http://localhost:33333/api/',
-            monitoring_urls: ['http://localhost:33633/']
+            monitoring_urls: ['http://localhost:1234/']
```
WilliamBZA: Can we undo this change?

sergioc: We can, but if I undo the monitoring server won't connect to SP, at least for me. Are you sure?

WojcikMike: now the question is do we want to, as SC.Monitoring while being run from VS use 1234 as a port

WilliamBZA: SC.Monitoring should change to have it's default be reasonable. 1234 might work for now, but 33633 will be easier for development and debugging later. Once SC.Monitoring has been installed, it is extremely unlikely that 1234 will remain the port, whereas 33633 has a much higher chance.

WojcikMike: Ok let us change default port of SC.Monitoring to be 33633, and let's revert this change.

sergioc: Done.

sergioc:
@WilliamBZA @WojcikMike please merge if happy.

PR #496, opened 2017-08-28
Change tests to async flush

WilliamBZA:
Fixes the tests by changing them to async and flushing the mocked http requests asynchronously

### DIFF
--- a/src/ServicePulse.Host.Tests/tests/js/services/services.monitoring.spec.js
+++ b/src/ServicePulse.Host.Tests/tests/js/services/services.monitoring.spec.js
@@ -29,9 +29,9 @@
     }));
 
     it('should push endpoints retrieved from monitoring server', function () {
-        scConfig.monitoring_urls = ['http://localhost:1234/diagrams'];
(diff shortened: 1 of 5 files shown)


src/ServicePulse.Host.Tests/tests/js/services/services.monitoring.spec.js
```diff
@@ -28,40 +28,54 @@
...
-        scConfig.monitoring_urls = ['http://localhost:33633/', 'http://localhost:5678/'];
+        scConfig.monitoring_urls = ['http://localhost:1234/', 'http://localhost:5678/'];
 
         var monitoredEndpoints = [];
-        monitoringService.createEndpointsSource(5).subscribe(function (response) {
+        var subscription = monitoringService.createEndpointsSource(5).subscribe(function (response) {
             monitoredEndpoints.push(response);
+
+            if (monitoredEndpoints.length == 4) {
+                expect(monitoredEndpoints.length).toEqual(4);
```
WilliamBZA: This is just to get around the jasmine of no expectations. This test fails if it takes longer then 5 seconds, so the test will still show the failure

PR #497, opened 2017-08-28
Add big graph

WojcikMike:
Adding big graph on the details screen

### DIFF
--- a/src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
+++ b/src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
@@ -1,0 +1,89 @@
+﻿(function(window, angular, undefined) {
+    'use strict';
+
+    angular.module('ui.particular.largeGraph', [])
+        .directive('largeGraph',
+            function() {
+                return {
+                    restrict: 'E',
+                    scope: {
+                        data: '=plotPoints',
+                        avg: '=plotAverage'
+                    },
+                    template: '<svg></svg>',
+                    link: function link(scope, element, attrs) {
+                        var svg = element.find('svg')[0];
+                        var heigth = 600;
+                        var graphWidth = 750;
+                        var totalWidth = 800;
+                        var margin = 20;
+                        var points = scope.data;
+                        var average = scope.avg;
+                        var max = Math.max(average * 1.5, d3.max(points));
+
+                        var scaleY = d3.scaleLinear()
+                            .domain([0, max])
+                            .range([heigth - margin, margin]);
+
+                        var scaleX = d3.scaleLinear()
+                            .domain([0, points.length - 1])
+                            .range([margin, graphWidth - margin]);
+
+                        var area = d3.area()
+                            .x(function (d, i) {
+                                return scaleX(i);
+                            })
+                            .y(function (d, i) { return scaleY(d); })
+                            .y1(function (d) { return scaleY(0); }) 
+                            .curve(d3.curveLinear);
+
+                        var line = d3.line()
+                            .x(function (d, i) {
+                                return scaleX(i);
+                            })
+                            .y(function(d, i) {
+                                return scaleY(d);
+                            })
+                            .curve(d3.curveLinear);
+
+                        var chart = d3.select(svg)
+                            .attr('width', totalWidth)
+                            .attr('height', heigth);
+
+                        chart.append('path')
+                            .datum(points)
+                            .attr('d', area)
+                            .attr('fill', attrs.color);
+
+                        chart.append('path')
+                            .datum(Array(points.length).fill(average))
+                            .attr('d', line)
+                            .attr('stroke', 'black');
+
+                        chart.append('path')
+                            .datum(Array(points.length).fill(0))
+                            .attr('d', line)
+                            .attr('stroke', 'gray');
+
+                        chart.append("text")
+                            .attr("x", graphWidth - margin + 3)
+                            .attr("y", heigth / 2 + 5)
+                            .attr("text-anchor", "start")
+                            .attr("font-size", 12)
+                            .attr("font-family", "sans-serif")
+                            .text(average.toFixed(2));
+
+                        var scaleYValue = (scaleY(average == 0 ? 1 : average) + 3) | 0;
+
+                        chart.append("text")
+                            .attr("x", 0)
+                            .attr("y", scaleYValue)
+                            .attr("text-anchor", "start")
+                            .attr("font-size", 12)
+                            .attr("font-family", "sans-serif")
+                            .text("ø");
+                    }
+                };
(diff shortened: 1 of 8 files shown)


src/ServicePulse.Host/app/js/views/endpoint_details/endpoint_details.controller.js
```diff
@@ -45,6 +45,15 @@
                     $scope.loading = false;
                 }
 
+                $scope.throughput = $scope.endpoint.graphData.throughput;
```
WilliamBZA: Why the second reference? What's wrong with a long binding reference?

`<large-graph plot-points="endpoint.graphData.throughput.points"></large-graph>`

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
```diff
@@ -0,0 +1,113 @@
...
+                    scope: {
+                        data: '=plotPoints',
+                        avg: '=plotAverage',
+                        dates: '=xaxisPoints',
+                        width: '=plotWidth',
+                        heigth: '=plotHeight'
+                    },
+                    template: '<svg></svg>',
+                    link: function link(scope, element, attrs) {
+                        scope.$watch('data', function () {
```
WilliamBZA: Is the watch really necessary? Does it not rebind when the entire `throughput` object is rebound?

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
```diff
@@ -0,0 +1,113 @@
...
+                            d3.selectAll("svg > *").remove();
```
WilliamBZA: If the watch goes then this remove can also go

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
```diff
@@ -0,0 +1,113 @@
...
+                                .datum(points)
+                                .attr('d', area)
+                                .attr('fill', attrs.color);
+
+                            chart.append('path')
+                                .datum(Array(points.length).fill(average))
+                                .attr('d', line)
+                                .attr('stroke', 'black');
+
+                            chart.selectAll("dot")
```
WilliamBZA: Can you please switch to \` instead of `"`'?

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
```diff
@@ -0,0 +1,113 @@
...
+                                .data(points)
+                                .enter().append("circle")
+                                .attr("r", 3)
+                                .attr("cx", function (d, i) { return scaleX(i); })
+                                .attr("cy", function (d) { return scaleY(d); })
+                                .append("svg:title")
+                                .text(function (d, i) { return dates[i] + " | " + d; }); 
+
+                            chart.append("text")
```
WilliamBZA: Same here ` instead of "

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
```diff
@@ -0,0 +1,113 @@
...
+                                .text(average.toFixed(2));
+
+                            chart.append("g")
+                                .attr("class", "x axis")
+                                .attr("transform", "translate(0," + (heigth - margin) + ")")
+                                .call(d3.axisBottom(scaleX).tickFormat(function(d) {
+                                    return dates[d];
+                                }));
+                            
+                            chart.append("g")
```
WilliamBZA: And here...

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
```diff
@@ -0,0 +1,113 @@
...
+
+                            chart.append("text")
+                                .attr("x", graphWidth - margin + 3)
+                                .attr("y", heigth / 2 + 5)
+                                .attr("text-anchor", "start")
+                                .attr("font-size", 12)
+                                .attr("font-family", "sans-serif")
+                                .text(average.toFixed(2));
+
+                            chart.append("g")
```
WilliamBZA: And here...

WilliamBZA: And here...

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
```diff
@@ -0,0 +1,113 @@
...
+                                }));
+                            
+                            chart.append("g")
+                                .attr("class", "y axis")
+                                .attr("transform", "translate(" + margin + ", 0)")
+                                .call(d3.axisLeft(scaleY));
+
+                            var scaleYValue = (scaleY(average == 0 ? 1 : average) + 3) | 0;
+
+                            chart.append("text")
```
WilliamBZA: Zmien to, tam I tam

src/ServicePulse.Host/app/js/views/endpoint_details/endpoint_details.controller.js
```diff
@@ -45,6 +45,15 @@
...
+                $scope.throughput.xAxisPoints = $scope.endpoint.graphData.timeAxisValues;
+
+                $scope.throughput.xAxisPoints.forEach(function(item, index) {
```
WilliamBZA: I think this would be clearer to read as a `map`

```JavaScript
$scope.throughput.xAxisPoints = $scope.endpoint.graphData.timeAxisValues.map(function(item) {
    var date = new Date(item);
    return date.toLocaleDateString() + ' ' + date.toLocaleTimeString();
});
```

That means the assignment 2 lines above can also be removed.

WilliamBZA: Can potentially also remove the entire property as it's not used anywhere besides to create the xAxisPoints.

```JavaScript
delete $scope.endpoint.graphData.timeAxisValues;
```

But this is a purely superfluous suggestion.

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
WojcikMike: It is necessary. The reason why it is not needed in other places is due to the fact that it is displayed in grid where we change whole collection => all html is rerendered. In here the control stays, and it was not refreshing to data changes. I verified and that was the suggested solution.

src/ServicePulse.Host/app/js/directives/ui.particular.largeGraph.js
WojcikMike: And as I refresh the control old rendered values stayed.

PR #505, opened 2017-09-13
Only show monitoring if url is configured

WilliamBZA:


src/ServicePulse.Host/app/js/app.controller.js
```diff
@@ -17,6 +17,11 @@
         scConfig,
         uriService
     ) {
+        $scope.isMonitoringEnabled = scConfig.monitoring_urls && scConfig.monitoring_urls.reduce(function (currentlyEnabled, url) {
+            return currentlyEnabled || url;
+        }, false);
+
+        $scope.isRecoverabilityEnabled = scConfig.service_control_url;
```
WojcikMike: I assume that the field `isRecoverabilityEnabled` is for future use only. As currently there is no usage of that file. What is more it doesn't contain bool value, but actual SC URL. You can use famous bang bang operator ;)

WilliamBZA: Correct, future use.

They both container url values 😉 Truthy enough for me

PR #506, opened 2017-09-13
ServicePulse installer with monitoring feature toogle

tmasternak:
## Overview
This PR enables users to run `ServicePulse` installer in feature selection mode. In such case they are able to specify if they want to enable `Recoverability and Auditing` and/or `Monitoring` capabilities.
In order to run the installer in feature selection mode one has to run installer `.exe` with predefined flag:
```
Particular.ServicePulse-1.9.2.exe ENABLE_FEATURE_SELECTION="TRUE"
```
otherwise users will see old configuration dialog. When flag is specified the new configuration dialog is show:
image

## SevicePulse.Host Integration
Selection performed by the user on the new configuration page results in changes to `app.constants.js` file. Installer will provide values for `service_control_url` and `monitoring_urls`. In case feature is disabled it will be set to empty string value for `service_control_url` and array of single empty string value for `monitoring_urls`.

### DIFF
--- a/src/ServicePulse.Install.CustomActions/CustomAction.cs
+++ b/src/ServicePulse.Install.CustomActions/CustomAction.cs
@@ -210,7 +210,7 @@
                     }
                     else
                     {
-                        Log(session, string.Format(@"Extracted monitoring URI {0} from {1}", extracted, filePath));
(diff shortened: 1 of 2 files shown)


WilliamBZA:
👍 Tested the build from TC, appears to work as expected.

PR #508, opened 2017-09-14
Force a digest to trigger update

WilliamBZA:


WilliamBZA:
Seems like the data was binding at initial load, but angular wasn't displaying it. Probably due to how we're calling the updateUI function.

PR #513, opened 2017-09-20
Number formatting

WilliamBZA:
Don't forget to run `bower install` to download the latest bower requirements (side note: we should move away from bower to vanilla npm I think).

### DIFF
--- a/src/ServicePulse.Host/app/js/views/endpoint_details/endpoint_details.controller.js
+++ b/src/ServicePulse.Host/app/js/views/endpoint_details/endpoint_details.controller.js
@@ -8,7 +8,8 @@
         toastService,
         serviceControlService,
         monitoringService,
-        historyPeriods) {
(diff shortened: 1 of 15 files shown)


src/ServicePulse.Host/app/js/views/endpoint_details/endpoint_details.controller.js
```diff
@@ -32,6 +33,10 @@
             updateUI();
         };
 
+        $scope.formatTime = function (value) {
```
WilliamBZA: I'm not sure if there's a better way to call js functions from a view. Ideally these would be global functions on the `formatter` object, but I'm not sure how to bind the view to something outside of $scope

sergioc:
Given the chosen solution (for now), I think it looks good.

WilliamBZA:
Ok, I much prefer this version.

PR #519, opened 2017-10-02
Display actual value instead of an average

WojcikMike:


WilliamBZA:
Do the `displayValue` value creators not need to be updated here as well? I think they still format the average

WilliamBZA:
I've rebased

WilliamBZA:
Do the `displayValue` value creators not need to be updated here as well? I think they still format the average

PR #522, opened 2017-10-06
Polyfills for Array.findIndex and Array.fill

tmasternak:
This is partial solution to https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/270.

Now the page renders but metrics data is not properly refreshed.


tmasternak:
@WilliamBZA I've added a check to prevent exception throwing when data coming from SC does not match those from SC.Monitoring

PR #533, opened 2017-10-20
Introduce webpack and modules

WilliamBZA:
This is the first step towards componentizing ServicePulse.

This PR introduces WebPack as the tool to bundle capabilities into single bundles that can be included. I foresee this eventually allowing us to move each capability to it's own repo and then to import them into ServicePulse via npm packages.

There is a build step change required which is to do the actual webpack compilation. For now it's as simple as running `./node_modules/webpack/bin/webpack.js --config app/modules/monitoring/monitoring.webpackconfig.js` but that should be automated on the build server. This step generates a `monitoring.dist.js` file which is the ENTIRE monitoring capability (currently including js and html, but should also include the css changes specific to monitoring).

To develop locally I recommend setting up a file watcher to run the webpack step. [Watchpack](https://github.com/webpack/watchpack) looks like it could be a good fit.

### Pros

* This allows us to componentize capabilities in the future without blurring the lines
* We can use webpack to do the bundling and minification for us
* Already works with ES6 functions, so we can code js as if we don't hate ourselves

### Problems and future requirements

* [x] Configure build server
* [ ] Find out a way to update the `index.html` includes to load the hashes generated to prevent caching problems
* [ ] Separate the recoverability portion to a dedicated module
* [ ] Treat dependencies (which are currently loaded via bower) as package requirements, so that we can start versioning module dependencies separately


SzymonPobiega:
Looks awesome. Can't wait to be able to add my own SP modules ❤️

WojcikMike:
Added support for Team City and rebased

WilliamBZA:
I propose we drop the WIP and deal with the other items as future work.

WilliamBZA:
Verified that the installer produced by this has the expected results

WojcikMike:
I agree, let's merge this

PR #534, opened 2017-10-20
Adding period of 1 minute. Changing frequency of refresh

WojcikMike:
@WilliamBZA @tmasternak this pr shouldn't be merged before SC.Monitorig change. As sending a request with 1 minute period cause an exception in SC.Monitoring

### DIFF
--- a/src/ServicePulse.Host/app/js/services/services.monitoring.js
+++ b/src/ServicePulse.Host/app/js/services/services.monitoring.js
@@ -5,7 +5,7 @@
     function Service($http, rx, scConfig, uri, $q) {
 
         function createEndpointsSource(historyPeriod) {
-            return Rx.Observable.interval(5000).startWith(0)
+            return Rx.Observable.interval(calculateIntervalInMiliseconds(historyPeriod)).startWith(0)
                 .flatMap(function (i) {
                     return Rx.Observable.fromArray(loadEndpointDataFromMonitoringService(historyPeriod))
                         .flatMap(function (p) {
@@ -43,10 +43,14 @@
         }
 
         function createEndpointDetailsSource(endpointName, sourceIndex, historyPeriod) {
-            return Rx.Observable.interval(5000).startWith(0)
+            return Rx.Observable.interval(calculateIntervalInMiliseconds(historyPeriod)).startWith(0)
                 .flatMap(function (i) {
                     return Rx.Observable.fromPromise(loadEndpointDetailsFromMonitoringService(endpointName, sourceIndex, historyPeriod));
                 });
+        }
+
+        function calculateIntervalInMiliseconds(period) {
+            return (period * 1000 * 60) / 20;
         }
 
         function getHistoryPeriod() {
@@ -55,7 +59,8 @@
 
         var service = {
             createEndpointsSource: createEndpointsSource,
-            createEndpointDetailsSource: createEndpointDetailsSource
+            createEndpointDetailsSource: createEndpointDetailsSource,
+            calculateIntervalInMiliseconds: calculateIntervalInMiliseconds
         };
 
         return service;
--- a/src/ServicePulse.Host/app/js/services/services.monitoring.js
+++ b/src/ServicePulse.Host/app/js/services/services.monitoring.js
@@ -4,8 +4,8 @@
 
     function Service($http, rx, scConfig, uri, $q) {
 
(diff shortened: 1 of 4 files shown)


src/ServicePulse.Host/app/js/views/monitored_endpoints/monitored_endpoints.controller.js
```diff
@@ -64,7 +64,7 @@
             });
 
             endpointsFromScSubscription =
-                Rx.Observable.interval(5000)
+                Rx.Observable.interval(monitoringService.calculateIntervalInMiliseconds($scope.selectedPeriod.value))
```
WilliamBZA: Do we really want to change the rate at which we query ServiceControl?

src/ServicePulse.Host/app/js/services/services.monitoring.js
```diff
@@ -43,19 +43,24 @@
...
         function createEndpointDetailsSource(endpointName, sourceIndex, historyPeriod) {
-            return Rx.Observable.interval(5000).startWith(0)
+            return Rx.Observable.interval(calculateIntervalInMiliseconds(historyPeriod)).startWith(0)
                 .flatMap(function (i) {
                     return Rx.Observable.fromPromise(loadEndpointDetailsFromMonitoringService(endpointName, sourceIndex, historyPeriod));
                 });
         }
 
+        function calculateIntervalInMiliseconds(period) {
+            return (period * 1000 * 60) / 20;
```
WilliamBZA: Every 1 minute? Shouldn't we refresh more frequently than the window?

WilliamBZA: ☝️  Assuming the user has selected "The last 1 minute" as their window

WojcikMike: if period is 1 than
1 * 1000 * 60 = 60000
60000 / 20 = 3000 ms = 3 sec
So for period 1 minute we refresh every 3 seconds, which is a length of the interval. 

For 5 minutes period it will be 15 seconds (also the length of the interval). At the moment we query every 5 seconds. But it doesn't make sense to query more frequently than interval length.

src/ServicePulse.Host/app/js/views/monitored_endpoints/monitored_endpoints.controller.js
WojcikMike: This address this requirement: https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/339

src/ServicePulse.Host/app/js/services/services.monitoring.js
WilliamBZA: Oops, missed the `/ 20` part!

src/ServicePulse.Host/app/js/views/monitored_endpoints/monitored_endpoints.controller.js
WilliamBZA: I would rather us not query SC more often. We can query the monitoring endpoint more often, but SC struggles enough as it is. And having the number refresh a bit slower (I'd say even slower than 5 seconds) is just fine for an error count @Particular/devops-launch-wave-squad

src/ServicePulse.Host/app/js/services/services.monitoring.js
tmasternak: What if we added the refresh interval to `historyPeriods` as third property? We could hardcode it there and not require `calculateIntervalInMiliseconds`.

src/ServicePulse.Host/app/js/views/monitored_endpoints/monitored_endpoints.controller.js
tmasternak: I agree that we don't need to increase the refresh rate for data coming from SC instance.

tmasternak:
PR on for the SC.Monitoring has been raised https://github.com/Particular/ServiceControl.Monitoring/pull/69

src/ServicePulse.Host/app/js/services/services.monitoring.js
WilliamBZA: I'm ok with using a heuristic calculation.

PR #543, opened 2017-11-01
Updating signing thumbprint

kbaley:
The signing certificate is about to expire on December 22. A replacement has been installed on all the build agents. This is the thumbprint for the new one

### DIFF
--- a/src/Setup/ServicePulse.aip
+++ b/src/Setup/ServicePulse.aip
@@ -98,7 +98,7 @@
     <ROW Path="&lt;AI_DICTS&gt;ui_en.ail"/>
   </COMPONENT>
   <COMPONENT cid="caphyon.advinst.msicomp.DigCertStoreComponent">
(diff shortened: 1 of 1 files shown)


kbaley:
@Particular/servicepulse-maintainers Not familiar with the deployment process here. Do I need to do anything to release a new version of the installer? Note that the existing certificate doesn't expire for another month.

WilliamBZA:
@kbaley We sign the build that is released on github, but I'm not sure how that gets packaged into the platform installer. @Particular/platforminstaller-maintainers do you have any idea?

HEskandari:
Platform Installer as a _thing_ on its own, has its own repo/build process. At runtime, it reaches out grabs the latest version of SI/SC/SP from AWS S3 and just runs the .exe/msi.

kbaley:
So it sounds like this should be good as long as it's released on AWS.

PR #545, opened 2017-11-09
[SPIKE] Spike display unit

WojcikMike:
The purpose of this spike is to experiment with possible formats of times in the metrics section 

@sergioc @ramonsmits This is somehow early but has a look. To see any results.
 - Pull the branch
 - Run it with an endpoint that has configured a monitored endpoint.
 - navigate to the details screen of that endpoint. 
 - look only on the processing time column:
image

 - you can change the formats of data using this sophisticated UI:
image - 6,8,9 represents different options.

 - at the moment the text and unit has separate classes assigned and the styles are set using spike.css file

To answer questions that will be raised:
 - Why have you done it on details screen not on the main monitoring screen - by mistake :D If needed I can move it, but I think it can be useful in the current form
 - Why are you generating endpoints in JS randomly generating data for them instead of preparing demo solution - it was more time efficient to do it this way. What is more, I can add new endpoint in a matter of seconds. 
 - Can we add/change the list of endpoints and data - yes we can.


sergioc:
- can the no. of decimal places (if decimal place is > 0) in option 6 be reduced to one?
- can the fake instances values range between 700ms and 1000ms to get an idea of annoying can the transition get in option 6?
- can the options also apply to critical time so that we have more things happening at the same time?
- can you remove the colors and instead leave add another prototype option to toggle colors for units only?

WojcikMike:
@sergioc please have a look

WojcikMike:
A new PR was created to implement chose format: https://github.com/Particular/ServicePulse/pull/545

PR #551, opened 2017-11-20
Moving graph directives into the monitoring module

WojcikMike:


src/ServicePulse.Host/app/modules/monitoring/monitoring.js
```diff
@@ -1,2 +1,7 @@
 ﻿require('./js/monitored_endpoints.module');
-require('./js/endpoint_details.module');
\ No newline at end of file
+require('./js/endpoint_details.module');
+
+require('./js/directives/ui.particular.graph.js');
+require('./js/directives/ui.particular.graphdecimal.js');
```
WilliamBZA: These are used in the `enpoind_details` and `monitored_endpoints` modules (separately). I think the requires should be in those sub-modules rather than in the root level here

WojcikMike:
@WilliamBZA have a look

PR #553, opened 2017-11-22
Adding dist into ignore

WojcikMike:
Updating ignore file to include the results of webpack build


WilliamBZA:
Just waiting for build so I can smoketest the installer

WojcikMike:
Build is greent, I tested the installer on my box and it works

PR #554, opened 2017-11-22
Implementing the desired time format

WojcikMike:
As the formatting for values looks bad, I will move it to HTML from SVG


src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.graph.js
```diff
@@ -75,17 +75,28 @@
...
 
                         var displayValue = lastValue.toFixed(2);
+                        var unit = "";
                         if (typeof scope.plotData.displayValue !== "undefined") {
-                            displayValue = scope.plotData.displayValue;
+                            displayValue = scope.plotData.displayValue.value;
+                            unit = scope.plotData.displayValue.unit;
                         }
 
                         chart.append("text")
```
WilliamBZA: I know it's not part of your change, but can we switch to single quotes all over here to be consistent?

WojcikMike:
@sergioc I moved the values outside of svg into the html, so please style it as at the moment it could have looked better ;)

WojcikMike:
The rest is done, I am waiting for a green build

sergioc:
Fixed presentation of values, but only in overview screen. The details screen should be fixed when we match the presentation there with the overview screen.

WojcikMike:
LGMT @WilliamBZA can you review and merge pls?

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
```diff
@@ -99,7 +111,8 @@
         'serviceControlService',
         'monitoringService',
         'historyPeriods',
-        'formatter'
+        'formatter',
```
WilliamBZA: Do we still need the formatter to be injected?

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
```diff
@@ -9,7 +9,9 @@
         serviceControlService,
         monitoringService,
         historyPeriods,
-        formatter) {
+        formatter,
+        $filter) {
```
WilliamBZA: I've never seen `$filter` injected like this, and my google-fu is failing. Do you have a link you used that I could take a look through?

WilliamBZA: Does this apply the `| filter` operation that would normally be applied in HTML binding?

src/ServicePulse.Host/app/js/services/service.formatter.js
```diff
@@ -4,22 +4,32 @@
     function formatter() {
         var secondDuration = moment.duration(1000);
         var minuteDuration = moment.duration(60 * 1000);
-        var hourDuration = moment.duration(60 * 60 * 1000);
+        var hourDuration = moment.duration(60 * 1000); //this ensure that we never use minute formatting
```
WilliamBZA: How does making the hour duration the same length as a minute ensure we never use minute formatting? Shouldn't it be the other way around (i.e. make minute duration the same as hour duration)?

src/ServicePulse.Host/app/js/services/service.formatter.js
```diff
@@ -4,22 +4,32 @@
...
             var duration = moment.duration(value);
 
+            var time = { value: 0, unit: '' };
             if (duration >= dayDuration) {
-                return duration.format('D [d] h [h]');
+                time.value = duration.format('D [d] h [h]');
+                return time;
             } else if (duration >= hourDuration) {
-                return duration.format('hh:mm [h]');
+                time.value = moment(duration._data).format('HH:mm');
```
WilliamBZA: Why is this the only one accessing the `_data` property?

WojcikMike: The issue was that without it, when you use format HH:mm the hours will be skipped if equal to 00, leaving the value of for example: `34`.

This hack was suggested in few places for example here: https://stackoverflow.com/a/41443446

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
WojcikMike: Correct, it is to use filter in js code. as then I do:
`$filter('graphduration')(instance.metrics.processingTime)`

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
WojcikMike: No, I think it can be removed, let me do this

src/ServicePulse.Host/app/js/services/service.formatter.js
WojcikMike: Our chose way of formatting was to use hour formatting starting from a value of 1 minute. So for 1 minute we display: 00:01. 
The fact that those 2 values are equal (minute and hour duration) causes that minute duration will never be used.

PR #555, opened 2017-11-23
Making changes to number formatting to match what was agreed

WojcikMike:


sergioc:
- Need to ensure that numbers > 9.99 and < 1M do not show decimal places
- Time-based metrics should never show decimal places (because of dynamic units)
- Changes should also apply to Overview screen (I don't see thousands separators there) 

I couldn't generate a number bigger than one million to see the display above that type of figure

WojcikMike:
@sergioc few questions:
> Need to ensure that numbers > 9.99 and < 1M do not show decimal places

Wasn't the agreement that only below 1 we show decimal places (I don't know if this was implemented). I just want to confirm

> Time-based metrics should never show decimal places (because of dynamic units)

You mean s, ms etc.? If so I think it is already like that

> Changes should also apply to Overview screen (I don't see thousands separators there)

I will check it

sergioc:
>Wasn't the agreement that only below 1 we show decimal places (I don't know if this was implemented). I just want to confirm

If you look at the spreadsheet showing examples for each option, you'll see on all options that below 10 we use `1.2` as an example for for values lower than 10, and we use one decimal place. Also above one million the same rule applies (show one decimal place).

> You mean s, ms etc.? If so I think it is already like that

For that type of units. Perhaps decimal places won't show for time values in #554 ?

WojcikMike:
The PR #554 should be merged first, as there is one change that I will have to do so that new directive is used in metrics code.

WojcikMike:
@WilliamBZA can you review the latest commits?

PR #557, opened 2017-11-29
Returning constant max yAxis value for empty data series

tmasternak:
## Overview
This PR makes sure that yAxis has a fixed height for large graphs with empty data sets.


WojcikMike:
let's wait with this one, I will add configuration piece that will allow the minimum graph height on a metric basis.

PR #558, opened 2017-11-29
Fixes for time value formatting for values below 10s

tmasternak:


src/ServicePulse.Host/app/js/services/service.formatter.js
```diff
@@ -23,11 +23,11 @@
                 time.unit = 'min';
                 return time;
             } else if (duration >= secondDuration) {
-                time.value = duration.format('ss.S');
+                time.value = duration.format('ss');
                 time.unit = 's';
                 return time;
             } else {
-                time.value = duration.format('S');
+                time.value = duration.format('s,SSS');
```
WilliamBZA: Why `,` over other separators?

WojcikMike: it's to use as a thousand separator, so 1.5 s is presented as 1,500 ms

PR #559, opened 2017-11-29
Align monitoring screen styles

sergioc:
## Overview
This PR introduces number of styling fixes for small graphs.


src/ServicePulse.Host/app/css/particular.css
```diff
@@ -1694,16 +1779,16 @@ p.col-sort-active {
     text-transform: uppercase;
 //    text-align: left;
 //    width: 90px;
```
WilliamBZA: I know this has nothing to do with this change, but can we delete these lines? The `//` comment doesn't even work in css

tmasternak: done

Issue #563, opened 2017-12-01
Remove decimal point for Heartbeats and Failed Messages counters


sudheendra-bhat:
Remove decimal point for Heartbeats and Failed Messages counters

![image](https://user-images.githubusercontent.com/5879266/33476961-2b293136-d684-11e7-8beb-7d7d221ec695.png)

***************

-ServicePulse version 1.10.2
-ServiceControl version 1.43.4

WilliamBZA:
Thanks @sudheendra-bhat, we'll get right on that!

WilliamBZA:
PR submitted, waiting for review

PR #564, opened 2017-12-01
Stop using decimals for badge counts

WilliamBZA:
Connects to https://github.com/Particular/ServicePulse/issues/563

Stops using a decimal point for things that will be whole numbers only.

Smaller number will show the actual value, larger will be e.g. `25k` instead of `25.3k`


sergioc:
Whole numbers should indeed not show any decimal places, but otherwise this will introduce an inconsistency with the new number formatting scheme. Examples:

#### < 1 (one decimal place)
- 0.1

#### < 10 (one decimal place if number is not whole)
- 1.2

#### < 1,000,000 (never use decimal places)
- 12
- 123
- 1,234
- 12,345
- 123,456
 
#### < 1,000,000,000 (abbreviate with **M** and use one decimal place if number is not whole)
- 123,4M
 
#### < 1,000,000,000,000 (abbreviate with **T** and use one decimal place if number is not whole)
- 123,4T

WilliamBZA:
This uses the same formatter as the rest of the screens. There are simply no decimal places as these are always whole numbers.

WilliamBZA:

PR #565, opened 2017-12-01
Add experimental indicator

sergioc:


src/ServicePulse.Host/app/css/particular.css
```diff
@@ -1793,3 +1799,57 @@ p.col-sort-active {
...
+    text-align: center;
+    color: #fff;
+    background-color: #1A1A1A;
+    font-size: 12px;
+    padding: 6px;
+    border-radius: 4px;
+    white-space: nowrap;
+}
+
+    .tooltip-contents a {
```
WilliamBZA: The alignment of these 2 selectors and rules is different to the rest of the rules

src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
```diff
@@ -35,10 +35,16 @@ <h1>{{endpointName}}</h1>
...
-                            Queue Length
-                        </span>
+                    <div class="row queue-length-values tooltip-trigger">
+                        <div class="metric-digest-header">
+                            Queue Length <i class="fa fa-flask fake-link"></i>
+                        </div>
+                        <div class="interactive-tooltip">
+                            <div class="tooltip-contents">
+                                <p>Queue length: The estimated number of messages in an endpoint's queue.</p>
+                                <p>NOTE: This is an experimental feature. <a href="https://docs.particular.net/nservicebus/operations/metrics/#metrics-captured-queue-length" target="_blank">Learn more</a> <i class="fa fa-external-link fake-link"></i></p>
```
WojcikMike: I think we can't do this. I remember talking to @Particular/docs-maintainers about creating links to our documentation and the decision was to not do it. The reason for that is increased complexity in maintaining doc urls. Unless we will get an explicit allowance for creating links to doco I am suggesting to remove this link or to link to google search results of a certain phrase.

src/ServicePulse.Host/app/modules/monitoring/views/monitored_endpoints.html
```diff
@@ -34,42 +34,50 @@ <h1>No endpoints available for monitoring</h1>
...
                 <div class="row box-header">
-                    <div class="col-sm-12 no-side-padding" uib-tooltip="Queue length (# msgs): The estimated number of messages in an endpoint's queue.">
-                        <p>Queue Length <span>(msgs)</span></p>
+                    <div class="col-sm-12 no-side-padding">
+                        <div class="tooltip-trigger">
+                            Queue Length <span class="table-header-unit">(msgs)</span> <i class="fa fa-flask fake-link"></i>
+                            <div class="interactive-tooltip">
+                                <div class="tooltip-contents">
+                                    <p>Queue length: The estimated number of messages in an endpoint's queue.</p>
+                                    <p>NOTE: This is an experimental feature. <a href="https://docs.particular.net/nservicebus/operations/metrics/#metrics-captured-queue-length" target="_blank">Learn more</a> <i class="fa fa-external-link fake-link"></i></p>
```
WojcikMike: As above

src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
tmasternak: I would argue that the link is necessary here. My arguments are:
  * QL is experimental and want to point to exact section in a doco page - otherwise the link is useless IMHO,
  * This is to be removed when problems with Ql are solved.

DavidBoike: Thinking more critically about the "no links in source code" thing, I believe that's absolutely critical for NServiceBus because those versions stick around forever. I'm not sure if it's as critical in ServicePulse where customers are expected to upgrade more regularly, especially if it's a temporary thing.

WilliamBZA: I'm not sure about that. Customers only seem to update when they log a support case and we ask them to update. If we can get away without direct links that would be better.

tmasternak: @WilliamBZA I think that in this case a search query url is not good enough :/. What we need is specific [section](https://docs.particular.net/nservicebus/operations/metrics/#metrics-captured-queue-length). 
Otherwise it will be confusing to the users. I click to learn more why QL is experimental and I get to a search results page. That's confusing IMHO.

WilliamBZA: A section that has to be maintained at the same url forever and never moved or deleted. I'd much rather we put a concise description in a tooltip if possible? I know it's difficult to fit all the required information in, but even just the note that's currently in the documentation page should be enough.

tmasternak: I agree to disagree. I've changed the direct link to search query link.

sergioc:
I don't really agree with linking to search results. After clicking the link, how am I supposed to know which page should I open? Did you have click a "Learn more" link and got dumped in a search results page where only one result mattered?

Not to mention the weird impression users will get. 

Also, we're already tagging the feature as experimental. Just like everything else in the product, if you need to "learn more" you need to check documentation. Why should this require a direct link and a mediocre experience?

WojcikMike:
@sergioc regarding links, we already have hyperlinks that take you to google search results. In this launch wave, we added 2 warning messages (when there are no endpoints and some other) that take you to google search results. Just saying for clarity sake

tmasternak:
I don't agree to linking to query search results but I understand that the decision is to be made by maintainers. The search query is good enough.

BTW (if the link is not valid anymore it redirects to search ;)).

Can we merge it with search query?

sergioc:
Until we sort out a better plan for dealing with doco links, I'll ask again: why not just get rid of the link here?

tmasternak:
> Until we sort out a better plan for dealing with doco links, I'll ask again: why not just get rid of the link here?

Becasue as a @Particular/devops-launch-wave-squad we decided that search query is better.

WilliamBZA:
> Because as a @Particular/devops-launch-wave-squad we decided that search query is better.

_Nitpick: It's not a squad decision, it's a TF decision. It's also not a maintainer decision. All the maintainers can do is provide feedback on the quality of a PR. The maintainers could come up with rules in a contribution guide, but that's too much noise._

I'm happy as long as there are no perma-links. I would have preferred @sergioc's suggestion of just dropping the link altogether, but a search is ok.

src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
```diff
@@ -39,10 +39,16 @@ <h1>{{endpointName}}</h1>
...
-                            Queue Length
-                        </span>
+                    <div class="row queue-length-values tooltip-trigger">
+                        <div class="metric-digest-header">
+                            Queue Length <i class="fa fa-flask fake-link"></i>
+                        </div>
+                        <div class="interactive-tooltip">
+                            <div class="tooltip-contents">
+                                <p>Queue length: The estimated number of messages in an endpoint's queue.</p>
+                                <p>NOTE: This is an experimental feature. <a href="https://www.google.pl/search?q=%2Fsite%3A+docs.particular.net+metrics-captured-queue-length" target="_blank">Learn more</a> <i class="fa fa-external-link fake-link"></i></p>
```
sergioc: Better search results page:

https://docs.particular.net/search?q=nservicebus+queue+length+metric+%2bexperimental

src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
weralabaj: > I think we can't do this. I remember talking to @Particular/docs-maintainers about creating links to our documentation and the decision was to not do it. The reason for that is increased complexity in maintaining doc urls.

I think that the main concern was linking from inside the NSB code, as that indeed will be permanent (alternative would be to release a patch every time the link changes, we just won't do it). And if I recall that was mainly about linking to exceptions or more info about particular feature.

The general guideline is to avoid it, if possible. In many cases that can be achieved by crafting a meaningful comment or error message and giving user a "hint" what to look for. But I think in the context of a tool such as SP this is a common industry practice to link to help articles, etc. and I don't see a problem here.

Also note that maintainers in reviews were checking things like adding redirects to page headers when page name or location changed, so even if URL is different over time, that should not break the link. Assuming, we do that work properly (which I believe we do).

Personally, I think it's better to not give link than link to search results, it's not helpful for the user and is not what you'd expect after clicking a link. As for including link to a particular page from SP, I don't have objections. Just remember about redirects when updating the content ;)

PR #567, opened 2017-12-04
Critical and Processing times values swapped under big graph

tmasternak:
## Outcome
image


sergioc:
@tmasternak if I understand the logic of the change correctly, shouldn't we also swap Throughput with Retries rate?

sergioc:
Scheduled retries are not a subset of total throughput, so it's not necessary to follow the same pattern as processing -> critical time

PR #570, opened 2017-12-05
Endpoint connectivity error display adjustment

sergioc:


src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
```diff
@@ -72,7 +74,9 @@
                         }
                     }, function (err) {
                         // Warn user?
-                    });
+                        });
+
+                    $scope.endpoint.isStale = $scope.endpoint.isStale && instance.isStale;
```
WilliamBZA: Should this not be an `||`? Is an endpoint stale only if all the instances are stale or is it stale if at least one instance is stale?

tmasternak: The endpoint is stale if all instances are stale => `isStale` is true for all instances => `&&`.

WilliamBZA: :+1:

WilliamBZA:
Has conflicts. I also think a milestone and a label (feature or bug) is needed for some of these PRs so that the release notes builder knows what to do

tmasternak:
@WilliamBZA rebased.

PR #574, opened 2017-12-05
Adjust failed msgs indicator

sergioc:


src/ServicePulse.Host/app/modules/monitoring/js/monitored_endpoints.controller.js
```diff
@@ -66,7 +66,12 @@
                 endpoint.isConnected = true;
                 fillDisplayValuesForEndpoint(endpoint);
                 if (index >= 0) {
+                    var serviceControlId = $scope.endpoints[index].serviceControlId;
```
WilliamBZA: It took me a while to work out what was going on here. Can we maybe rename the variables to `previousServiceControlId` and `previousErrorCount` or something else that represents the intention a bit clearer?

src/ServicePulse.Host/app/modules/monitoring/js/monitored_endpoints.controller.js
```diff
@@ -87,18 +92,21 @@
...
                 }).subscribe(function (endpoint) {
                     var index = $scope.endpoints.findIndex(function(item) { return item.name === endpoint.title });
                     if (index >= 0) {
+                        $scope.endpoints[index].serviceControlId = endpoint.id;
                         $scope.endpoints[index].errorCount = endpoint.count;
                     } else {
                         $scope.endpoints.push({ name: endpoint.title, errorCount: endpoint.count, isConnected: false, isServiceControlOnly: true, serviceControlId: endpoint.id });
                     }
+
+                    $scope.$apply();
```
WilliamBZA: Is this needed for _every_ subscription update, or only the first?

src/ServicePulse.Host/app/modules/monitoring/js/monitored_endpoints.controller.js
tmasternak: 👍 for renaming

src/ServicePulse.Host/app/modules/monitoring/js/monitored_endpoints.controller.js
tmasternak: This is actually not needed at all. Removed.

PR #575, opened 2017-12-06
Show endpoint name instead of id on instace breakdown grid

tmasternak:
## Overview
Currently we are showing `instance.id` on the instance breakdown grid. This does not work as expected when users overrides `HostDisplayName` and does not use `instanceId` argument.

/cc: @danielmarbach


Scooletz:
@tmasternak Could you clarify the following. Does it mean that name is assigned either to the instance Id (if set) or the `HostDisplayName` ?

PR #583, opened 2017-12-12
Less UI flickering with better model updates

tmasternak:
## Overview
This PR decreases a flickering effect on the UI when failed message groups are shown on the grids.


WilliamBZA:
Still marked as WIP, but looks good

PR #590, opened 2018-01-12
Failed msg group links removed from message type detail table

tmasternak:
## Overview
This PR removes the failed message group links from breakdown tables by message type on monitored endpoint details. This is caused by the fact that current grouping in SC is performed on the basis of `messageType` value. 

This is problematic as different logical endpoints can process the same message type. This can be solved in the future by introducing new groupings in SC that will be based on (endpoint_name, message_type) pairs.

## Sample problematic setup (based on `MonitoringDemo`)
Two endpoints `Shipping` and `Billing` each subscribed to `OrderCreated` event.


tmasternak:
@WilliamBZA this removes only `messageType` breakdown and keeps by instance breakdown.

PR #595, opened 2018-01-23
Deep link to docs articles

sergioc:
Currently the "Learn more" button in the [monitoring empty state](https://user-images.githubusercontent.com/493451/35278600-dcf7ea02-004a-11e8-9535-b9096b967664.png) links to a list of search results:

image

This experience is just not worth it for the sake of fearing that the link may break (if that ever happens). @Particular/docs-maintainers already indicated there's no issue in deep linking to docs from SP.

The link from the Queue Length experimental indicator should similarly also deep link straight to the relevant portion of the article that covers the subject.


sergioc:
> Are we 100% certain we will never move that documentation ever?

We're not 100% sure, but again, the risk of breaking is not worth the poor experience provided by search results. If we detect a broken link, we can set up redirects. And we can also check links on a regular basis to ensure they don't break.

tmasternak:
Can we automate doco links checking? Maybe we can have a test that would validate them during build?

WilliamBZA:
If the docs maintainers are ok with direct links, and the Launchwave squad is ok with direct links, then I am ok. I am in no way going to pull a veto.

This needs a rebase before merging though.

PR #599, opened 2018-02-01
Add monitoring warning to config page and 2018 to footer

sergioc:


src/ServicePulse.Host/app/js/views/configuration/configuration.html
```diff
@@ -17,14 +17,21 @@ <h1>Configuration</h1>
...
             <div class="col-sm-12">
+
+                <div class="alert alert-warning">
+                    <i class="fa fa-warning"></i> <strong>Warning:</strong> The list of endpoints below only contains endpoints with the hearbeats plug-in installed. Toggling heartbeat monitoring on/off won't toggle <a href="https://docs.particular.net/monitoring/metrics/in-servicepulse" target="_blank">performance monitoring</a> <i class="fa fa-external-link fake-link"></i>
+                </div>
+
                 <div class="row box box-no-click table-header">
                     <div class="col-xs-3 col-sm-2 col-lg-1 no-side-padding">
-                        <p>Monitoring</p>
+                        <p>Heartbeats monitoring</p>
```
WilliamBZA: This title feels clunky.

What about just `Heartbeats`?
Or perhaps even dropping the title completely given that the description explains the list below?

sergioc: I'm good with making it just `Heartbeats`. We should keep those headers to leave no room for doubts about what those toggles are for.

sergioc:
I've renamed all titles and links for old monitoring screens to hearbeats to eliminate confusion with the new monitoring screens.

src/ServicePulse.Host/app/layout/footer.html
```diff
@@ -6,7 +6,7 @@
                     <div>
                         
                         <span>
-                            2017 © Particular Software
+                            2018 © Particular Software
```
WilliamBZA: :tada:

src/ServicePulse.Host/app/js/app.constants.js
```diff
@@ -6,7 +6,7 @@
         .constant('scConfig', {
             default_route: '/dashboard',
             service_control_url: 'http://localhost:33333/api/',
-            monitoring_urls: ['http://localhost:33633/']
+            monitoring_urls: ['http://localhost:33533/']
```
WojcikMike: What is the reason to change this URL? is it only local change?

sergioc: Not meant to be changed, I probably forgot to revert the change before commiting.

PR #614, opened 2018-04-17
Storing group selection

### LINKED ISSUE #597: Store view options in cookies
Store sorting/filtering dropdown selections and monitoring period in cookies so that a user doesn't have to re-select those options every time the page is reloaded. 

Example story: "as a user I want the list of failed message groups to default to group messages by the criteria of my choice so that I don't have to change grouping everytime I open that screen"

WojcikMike:
Resolves #597, #614


HEskandari:
This is ready to be reviewed.

WojcikMike:
This looks good to me

src/ServicePulse.Host/app/js/views/archive/controller.js
```diff
@@ -72,15 +73,29 @@
             vm.loadingData = false;
         };
 
+        var storeGroupOption = function (amount, unit) {
+            $cookies.put('archive_amount', amount);
+            $cookies.put('archive_unit', unit);
+        }
+
+        var loadDefaultGroup = function() {
```
WilliamBZA: This isn't loading a default group, this is loading the group from the cookies. Maybe `loadInitialGroupFromCookies`?

Also, what is the result if there aren't cookies here? What is the default behaviour?

WilliamBZA: I also kind of feel like this should be separated into 2 methods:

1 to load the values from cookies (or provide a default)
2 to save the values in the `timeGroup` variable

src/ServicePulse.Host/app/js/views/archive/controller.js
```diff
@@ -72,15 +73,29 @@
...
 
             vm.configuration = sharedDataService.getConfiguration();
             vm.error_retention_period = moment.duration(vm.configuration.data_retention.error_retention_period).asHours();
             vm.total = 1;
             vm.archives = [];
             vm.sort.page = 1;
-            vm.sort.start = moment.utc().subtract(vm.timeGroup.amount, vm.timeGroup.unit).format('YYYY-MM-DDTHH:mm:ss');
-            vm.sort.end = moment.utc().format('YYYY-MM-DDTHH:mm:ss');
+
+            loadDefaultGroup();
```
WilliamBZA: See above, it feels hidden as to what the `loadDefaultGroup` method does. Once it's loaded the default group, there's no indication in code that it sets the values to the `timeGroup` variable making it seem like there's no link between the load and the use of the timeGroup variable, someone might think it's ok to swap the lines around.

src/ServicePulse.Host/app/js/views/archive/controller.js
```diff
@@ -245,6 +260,8 @@
                 vm.timeGroup.buttonText = 'All Archived';
                 vm.sort.start = vm.sort.end = undefined;
             }
+
+            storeGroupOption(amount, unit);
```
WilliamBZA: `storeGroupSelectionInCookies`

WilliamBZA: Or `storeInitialGroupInCookies`

src/ServicePulse.Host/app/js/views/archive/controller.js
```diff
@@ -72,15 +73,29 @@
...
             vm.configuration = sharedDataService.getConfiguration();
             vm.error_retention_period = moment.duration(vm.configuration.data_retention.error_retention_period).asHours();
             vm.total = 1;
             vm.archives = [];
             vm.sort.page = 1;
-            vm.sort.start = moment.utc().subtract(vm.timeGroup.amount, vm.timeGroup.unit).format('YYYY-MM-DDTHH:mm:ss');
-            vm.sort.end = moment.utc().format('YYYY-MM-DDTHH:mm:ss');
+
+            loadDefaultGroup();
+            vm.selectTimeGroup(vm.timeGroup.amount, vm.timeGroup.unit);
```
WilliamBZA: This method also sets the `vm.timeGroup.amount` and `vm.timeGroup.unit` variables. Is it worth refactoring this so that the load method doesn't set the variables but rather returns it's own object that gets passed in as a variable here?

src/ServicePulse.Host/app/js/views/archive/controller.js
WilliamBZA: Maybe since it's a new function, switch to arrow function?

```
var loadDefaultGroup = () => {
};
```

src/ServicePulse.Host/app/js/views/failed_groups/controller.js
```diff
@@ -159,10 +160,26 @@
                 });
         };
 
+        var storeSelectedClassification = function (classification) {
+            $cookies.put("failed_groups_classification", classification);
+        };
+
+        var getDefaultClassification = function(classifiers) {
```
WilliamBZA: maybe `load` instead of `get`? It's minor, but load to me implies getting data from somewhere, while get implies the data is available already.

Maybe get is actually better then. If we choose get, maybe the previous methods should also be changed to get

src/ServicePulse.Host/app/js/views/archive/controller.js
WilliamBZA: Or drop the cookies part. That's optional that you guys can decide on :+1:

src/ServicePulse.Host/app/js/views/failed_messages/controller.js
```diff
@@ -45,6 +44,11 @@
             vm.sortButtonText = (sort === 'message_type' ? "Message Type" : "Time of Failure") + " " + (direction === 'asc' ? "ASC" : "DESC");
         }
 
+        var storeSortOption = function(sort, direction) {
```
WilliamBZA: Maybe instead of using `store` a better word for these types of things would be `save`?

src/ServicePulse.Host/app/js/views/failed_messages/controller.js
```diff
@@ -68,11 +72,23 @@
             }
         }
 
+        function loadDefaultSort() {
```
WilliamBZA: load or get, but standardize

src/ServicePulse.Host/app/js/views/failed_messages/controller.js
```diff
@@ -68,11 +72,23 @@
...
+            var sort = $cookies.get("failed_messages_sort");
+            var direction = $cookies.get("failed_messages_direction");
+
+            if (typeof sort === "undefined") sort = "time_of_failure";
+            if (typeof direction === "undefined") direction = "desc";
+
+            vm.sort = sort;
```
WilliamBZA: These methods should return an object instead of relying on a hidden API that the vm object must comply to

src/ServicePulse.Host/app/js/views/archive/controller.js
HEskandari: It does provide the default. The default, if the cookies are not available is `undefined` which would automatically default the filter to 'All Archived' messages.

src/ServicePulse.Host/app/js/views/failed_groups/controller.js
HEskandari: I tend to keep the changes to a minumum so any refactoring I stay clear of. I'm happy to clean it up in a separate commit.

HEskandari:
@WilliamBZA I have pushed a commit that should address most - if not all - of your concerns.

src/ServicePulse.Host/app/js/views/archive/controller.js
```diff
@@ -72,35 +73,53 @@
             vm.loadingData = false;
         };
 
+        var saveGroupOption = (amount, unit) => {
+            $cookies.put('archive_amount', amount);
+            $cookies.put('archive_unit', unit);
+        };
+
+        var getDefaultGroup = () => {
```
WilliamBZA: Maybe `getSelectedArchiveGroup` or something to at least indicate what the group is.

src/ServicePulse.Host/app/js/views/archive/controller.js
```diff
@@ -72,35 +73,53 @@
             vm.loadingData = false;
         };
 
+        var saveGroupOption = (amount, unit) => {
```
WilliamBZA: Still not a fan of the name `default` here. Maybe `saveSelectedArchiveGroup` or something a bit more descriptive?

src/ServicePulse.Host/app/js/views/archive/controller.js
```diff
@@ -72,35 +73,53 @@
...
             vm.archives = [];
             vm.sort.page = 1;
-            vm.sort.start = moment.utc().subtract(vm.timeGroup.amount, vm.timeGroup.unit).format('YYYY-MM-DDTHH:mm:ss');
-            vm.sort.end = moment.utc().format('YYYY-MM-DDTHH:mm:ss');
 
+            var defaultGroup = getDefaultGroup();
+
+            vm.timeGroup.amount = defaultGroup.amount;
+            vm.timeGroup.unit = defaultGroup.unit;
+            vm.selectTimeGroup(vm.timeGroup.amount, vm.timeGroup.unit);
```
WilliamBZA: The `selectTimeGroup` sets the `vm.timeGroup.amount` and `vm.timeGroup.unit` properties, so we can remove the 2 previous lines and make this a bit cleaner.

src/ServicePulse.Host/app/js/views/archive/controller.js
WilliamBZA: The difference between this name and the one for classification below, is that in the classification one, if there isn't a cookie value a default value is used. In this case, if there is no cookie value then `undefined` (or `null`?) is used.

sergioc:
Can we use this PR to also enable storing the last selected period value in the monitoring screens?

WojcikMike:
@sergioc I created a separate issue to cover that:
https://github.com/Particular/ServicePulse/issues/616

src/ServicePulse.Host/app/js/views/archive/controller.js
HEskandari: yes...the `undefined` value will translate to 'All Archive' option in the filter.

HEskandari:
@WilliamBZA @WojcikMike I have rebased this on master. Please review and merge if okay.

PR #617, opened 2018-04-18
Adding a library that would pretty-print JSON and XML 

### LINKED ISSUE #398: Format messagebody for easier readability
When viewing the messagebody in ServicePulse it would be nice to have the option to turn on "pretty- print" for easier reading of the message. Should support both xml and json.

WojcikMike:
fix #398
Any other type will be presented as it is now


src/ServicePulse.Host/app/modules/shell/shell.js
```diff
@@ -19,4 +19,5 @@ require('bootstrap/dist/css/bootstrap.css');
 require('angularjs-toaster/toaster.css');
 require('animate.css/animate.css');
 require('ui-select/dist/select.css');
-require('./constant.moment');
\ No newline at end of file
+require('./constant.moment');
+require('expose-loader?vkbeautify!vkbeautify');
```
WilliamBZA: Isn't there maybe another prettifier library that works with modules correctly?

src/ServicePulse.Host/app/js/services/service.prettify.js
```diff
@@ -0,0 +1,40 @@
+; (function (window, angular, undefined) {
+    'use strict';
+
+    function service($jquery) {
+        
+        this.prettifyText = function (text) {
+            if ((typeof text === 'object' && text.constructor === Object) || isJSON(text) ) {
+                return vkbeautify.json(text);
```
WilliamBZA: Angular has a built in json filter

`<div>{{jsonvalue | json}}</div>`

So maybe we just need a library for xml?

WilliamBZA: Perhaps https://github.com/krtnio/angular-pretty-xml?

src/ServicePulse.Host/app/modules/shell/shell.js
WilliamBZA: Maybe https://github.com/krtnio/angular-pretty-xml works properly with modules?

src/ServicePulse.Host/app/js/views/message/controller.js
```diff
@@ -51,7 +52,7 @@
 
             if (!angular.isDefined(message.messageBody)) {
                 serviceControlService.getMessageBody(message.message_id).then(function (msg) {
-                    message.messageBody = msg.data;
+                    message.messageBody = prettifyService.prettifyText(msg.data);
```
WilliamBZA: does the `msg` object maybe have the `content-type` property on it that we can use to check if the object is json or xml? Thereby bypassing the `try...catch` logic.

src/ServicePulse.Host/app/js/services/service.prettify.js
sergioc: That one seems to only do XML.

WilliamBZA: Angular has a built in filter for json, why do we have to include a separate one?>

WilliamBZA: https://docs.angularjs.org/api/ng/filter/json

sergioc: Didn't know, even better.

src/ServicePulse.Host/app/js/views/message/controller.js
WojcikMike: I need to verify it, as that would be super useful.

src/ServicePulse.Host/app/js/services/service.prettify.js
WojcikMike: I am fine with that

src/ServicePulse.Host/app/js/views/message/controller.js
WojcikMike: Ok so the situaion looks as follows: when we make a call to get body of a message it sets content type well to application/json or text/xml depending on the type. However when you ask again for the body then the 304 is returned and becasue of this line:
https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Infrastructure/Nancy/NotModifiedStatusExtension.cs#L42

The contenty type is changed to `text/html` (default content type for Response object in Nancy. Even that before this line `context.Response.ContentType' has the correct value, but it is overwriten by the default html one. So to make this mechanism to work we would have to change that line of code to something as follows:

```
context.Response = new Response {StatusCode = HttpStatusCode.NotModified, ContentType = context.Response.ContentType};
```

WilliamBZA: That's looking at the content-type of the http response, we need to look at the content type of the actual underlying message which would never change with nancy's pipeline. We want the `NServiceBus.ContentType` header from the actual message object

src/ServicePulse.Host/app/js/views/message/controller.js
```diff
@@ -49,25 +50,41 @@
...
-
             if (!angular.isDefined(message.messageHeaders)) {
                 serviceControlService.getMessageHeaders(message.message_id).then(function (msg) {
                     message.messageHeaders = msg.data[0].headers;
                 }, function () {
                     message.headersUnavailable = "message headers unavailable";
                 });
             }
+
+            if (angular.isDefined(message.messageHeaders) && !angular.isDefined(message.messageBody)) {
```
WojcikMike: The reason why I added: `angular.isDefined(message.messageHeaders)`
Is that I need headers to display body properly. I could either put whole this section to handler of loading headers, Or push the rendering concept of the message body a bit further. At the moment I decided to do the if's in controller not in the view to make it a bit more clear. I am open to suggestions as I don't like this piece.

The whole method is called on every change of panel.

WojcikMike:
@HEskandari I will merge this, if you will find something that troubles you let me know, we can fix it in next PR

PR #619, opened 2018-04-19
multi selection of rows with shift-click

### LINKED ISSUE #440: Retry custom selections of multiple messages
At the moment, the retry options are: Retry 1 or more selected messages or retry all. This is fine when dealing with a few messages but it is not great when there are many failed message.

Providing shift-click de/selection would give users fine-grained selection when they need it.

HEskandari:
Resolves #440


sergioc:
Still WIP @HEskandari ?

HEskandari:
@sergioc yes. There's an issue with Angular double firing the click event.

WojcikMike:
@WilliamBZA @HEskandari can you take a look?

src/ServicePulse.Host/app/js/directives/ui.particular.multi-checkboxlist.js
```diff
@@ -0,0 +1,104 @@
...
+
+
+    function controller($scope) {
+
+        var selectActions = Object.freeze({
+            Selection: 1,
+            Deselection: 2
+        });
+
+        $scope.selectRow = function(row, value){
```
WilliamBZA: Doesn't look like this is used anywhere in the view, so we could drop this from scope

src/ServicePulse.Host/app/js/directives/ui.particular.multi-checkboxlist.js
```diff
@@ -0,0 +1,104 @@
...
+            if (row.selected === value) {
+                return;
+            }
+            row.selected = value;
+            $scope.updateSelectedIdsWithMessage(row);
+        };
+
+        $scope.selectWithShift = function(row, index){
```
WilliamBZA: Same here, doesn't look like it has to be raised to scope

src/ServicePulse.Host/app/js/directives/ui.particular.multi-checkboxlist.js
```diff
@@ -0,0 +1,104 @@
...
+            if ($scope.selectedIds.length === 0) {
+                $scope.multiselection.lastSelectedIndex = -1;
+            }
+
+            //Removes text selection that happens 
+            //due to shift key being down.
+            document.getSelection().removeAllRanges();
+        };
+
+        $scope.selectSingleRow = function(row, index){
```
WilliamBZA: Also here

src/ServicePulse.Host/app/js/directives/ui.particular.multi-checkboxlist.js
```diff
@@ -0,0 +1,104 @@
...
+
+            //Stop event propagation since 
+            //there are nested elements
+            if (event.stopPropagation) event.stopPropagation();
+            if (event.preventDefault) event.preventDefault();
+            event.cancelBubble = true;
+            event.returnValue = false;
+        };
+
+        $scope.updateSelectedIdsWithMessage = function (row) {
```
WilliamBZA: Doesn't have to be on scope

src/ServicePulse.Host/app/js/directives/ui.particular.multi-checkboxlist.js
```diff
@@ -10,23 +10,23 @@
             Deselection: 2
         });
 
-        $scope.selectRow = function(row, value){
+        function selectRow(row, value){
```
WilliamBZA: I would actually prefer these as local vars instead of functions so that they don't get hoisted 😜

PR #623, opened 2018-04-25
Making the error message for message screen more descriptive

### LINKED ISSUE #609: SP obscures an error with SC reading a failed message with a misleading error
When SC fails to retrieve a failed message (eg https://github.com/Particular/ServiceControl/issues/1119)
this page `/#/failed-messages/message/b6ee1f42-722d-5072-9e90-19c2739fe434` displays

> Could not find message. This could be because the message URL is invalid or the corresponding message was processed and no longer audited by ServiceControl.

image

but if u look in the network tab it is clear the proxied call to SC has had a critical error with a `500 Internal Server Error`

image

This error is hidden from the user and gives useless guidance as to how someone should proceed. ie that is a generic nancyfx error

Specifically

 * "not found" should not be shown when a response is an error code
 * an error message should be surfaced to the UI that includes the log location of where to view the exception message and stack trace.

WojcikMike:
fix #609

### DIFF
--- a/src/ServicePulse.Host/app/js/views/message/controller.js
+++ b/src/ServicePulse.Host/app/js/views/message/controller.js
@@ -149,8 +149,12 @@
                 updateMessageDeleteDate(message, vm.error_retention_period);
                 vm.message = message;
             },
-                function() {
-                    vm.message = null;
+                function (response) {
+                    if (response.status === 404) {
+                        vm.message = { notFound: true };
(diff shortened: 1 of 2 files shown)


SimonCropp:
@WojcikMike thanks

WojcikMike:
Your welcome @SimonCropp may the obscured error message trouble you no more

Issue #581, opened 2017-12-08
Monitoring - Some messages might break Endpoint Details UI


jenspettersson:
When looking at endpoint details that contains messages that implements something else than ONLY `IEvent` / `IMessage` etc, the UI is broken.

Look at the following screenshot:
![image](https://user-images.githubusercontent.com/125493/33765302-6e7ecda8-dc18-11e7-84a8-f0272d41e477.png)
_(had to obfuscate a bit...)_

Some of the messages handled by this endpoint implements another interface and not the `IEvent` directly and those seem to break the UI. It's the messages that ends with "Saved" and "Published", the "Added" only implements the `IEvent` and looks good (but in the wrong place due to the UI being broken).

An easy way to reproduce:

Create a "marker" interface:

`interface IMyInterface : IEvent ...`

Create a class that implement that marker interface:

`class MyEvent : IMyInterface`

In our case, the marker interface is in another project than the actual message and this is the "message type" that the UI shows:

`My.Namespace.Messages.MyEvent, My.Namespace, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null; Another.Namespace.IMyInterface, Another.Namespace, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null`

***************

*Technical information if reporting bug:*

- **ServicePulse version:** v1.10.2
- **ServiceControl version:** v1.43.4

WilliamBZA:
Thanks @jenspettersson, we're on it!

sergioc:
> In our case, the marker interface is in another project than the actual message and this is the "message type" that the UI shows:
>
>`My.Namespace.Messages.MyEvent, My.Namespace, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null; Another.Namespace.IMyInterface, Another.Namespace, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null`

@jenspettersson given that we're currently supporting ONE message type and its associated metrics per row, how would you expect to see this presented?

jenspettersson:
```
-------------------------------------------------------
My.Namespace.Messages.MyEvent   | Metrics
My.Namespace-1.0.0.0            | 
-------------------------------------------------------
```

Just like the message types that only implements `IEvent/IMessage` is shown. I don't care if the message type implements another interface than `IEvent/IMessage`.

mikeminutillo:
[...]

PR #625, opened 2018-04-27
Truncating endpoint names

### LINKED ISSUE #612: Long MessageTypes break UI
On the monitoring endpoint details screen, when the message type is long enough it will expand over it's row. It is a subset of issue reported in here: #581 

Example can be seen here:
image

On the wide screen the issue `almost` don't exist
image

***************

*Technical information if reporting bug:*

- **ServicePulse version:** 1.12

HEskandari:
Resolves #612

### DIFF
--- a/src/ServicePulse.Host/ServicePulse.Host.csproj
+++ b/src/ServicePulse.Host/ServicePulse.Host.csproj
@@ -99,6 +99,7 @@
     <Content Include="app\js\directives\ui.particular.configurationTabs.js" />
     <Content Include="app\modules\modules.webpackconfig.builder.js" />
     <Content Include="app\modules\monitoring\js\constant.diagrams.js" />
+    <Content Include="app\modules\monitoring\js\directives\ui.particular.truncate.js" />
     <Content Include="app\modules\monitoring\js\directives\ui.particular.graphdecimal.js" />
     <Content Include="app\modules\monitoring\js\directives\ui.particular.graphduration.js" />
     <Content Include="app\modules\monitoring\js\directives\ui.particular.largeGraph.js" />
--- a/src/ServicePulse.Host/app/js/directives/ui.particular.js
+++ b/src/ServicePulse.Host/app/js/directives/ui.particular.js
@@ -19,6 +19,7 @@
         'ui.particular.largeGraph',
         'ui.particular.duration',
         'ui.particular.graphduration',
+        'ui.particular.truncate',
         'ui.particular.graphdecimal',
     ]);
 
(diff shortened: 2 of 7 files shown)


sergioc:
Seems like status icons are not following the endpoint text anymore, there's now a gap:

image

And truncating doesn't seem to work on Chrome v66.0.3359.139:

image

HEskandari:
@sergioc which screens are these? I have only changed the monitoring list and details pages.

HEskandari:
@sergioc well, this is what I see:

image

HEskandari:
@sergioc BTW your screenshot above is from MessageTypes. What this PR address is the truncation of endpoint name.

sergioc:
#612 is mostly about message type names and was expanded to include endpoint names. This PR will only partially resolve that issue.

I still cannot verify the fix here. Endpoints screen:

image

WojcikMike:
@sergioc I think we need to break this problem into 2. first is to long endpoint names that we can solve with truncation of text as @HEskandari implemented. Second problem is what the issue is about that message types are to long. Example of such message type is:
Shared2.EventMessage, Shared2, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null;IMyEvent, Shared, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null

If we would apply trimming to this string we would get:
`... Culture=neutral, PublicKeyToken=null` for all of the rows

I think we should change how we display the data, and in the lable put only type of the message so:
Shared2.EventMessage (which in itself might be very long string). And show the remainging part in some sort of tooltip, or expandable space below the row. Either way something more than simply truncating the text.

sergioc:
we already split properties into a separate row of text. I'd expect this issue to address only the truncating of message type names, so that it solves #612. Using your example:

> **Shared2.EventMessage**
> Shared2 Version=1.0.0.0 Culture=neutral PublicKeyToken=null IMyEvent Shared Version=0.0.0.0 Culture=neutral PublicKeyToken=null

Correctly parsing those properties into the second row could indeed be a separate issue and solve #581

WojcikMike:
Were are we doing that? On 1.13 it looks like this:
image

sergioc:
Seems like a bug. Here's how it renders message types from the monitoring demo:

image

WojcikMike:
Please check how it looks in SP 1.13.1. Maybe this never got to the SP release, and clear the cache ;)

WilliamBZA:
Closing this as we are not in agreement on how to resolve. It can be reopened too.

PR #633, opened 2018-05-22
Removed experimental icon from queue length headers

tmasternak:
This is part of Native Queue length implementation. Native ql is not experimental anymore.

### DIFF
--- a/src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
+++ b/src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
@@ -54,13 +54,7 @@
                 <div class="col-sm-12 no-side-padding graph-values">
                     <div class="row queue-length-values tooltip-trigger">
                         <div class="metric-digest-header">
-                            Queue Length <i class="fa fa-flask fake-link"></i>
-                        </div>
-                        <div class="interactive-tooltip">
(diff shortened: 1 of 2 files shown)


src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
```diff
@@ -54,13 +54,7 @@ <h1>
                 <div class="col-sm-12 no-side-padding graph-values">
                     <div class="row queue-length-values tooltip-trigger">
```
sergioc: If you also remove the link inside the tootltip, then remove the class `tooltip-trigger` and use a standard tooltip attribute for the metric explanation.

src/ServicePulse.Host/app/modules/monitoring/views/monitored_endpoints.html
```diff
@@ -28,14 +28,7 @@ <h1>Endpoints overview</h1>
                 <div class="row box-header">
                     <div class="col-sm-12 no-side-padding">
                         <div class="tooltip-trigger">
```
sergioc: If you also remove the link inside the tootltip, then remove the class `tooltip-trigger` and use a standard tooltip attribute for the metric explanation.

Issue #654, opened 2018-08-16
Endpoint details screen should show "No Messages processed in this period of time" instead of a blank section


sergioc:
If the user is looking at the details of an endpoint for a period of time where no messages have been processed yet, it should be made explicit that no messages have been processed with a simple message in the list of message types instead of showing it empty:

<img width="1453" alt="image" src="https://user-images.githubusercontent.com/493451/35563024-a7d821aa-05b5-11e8-98d1-d5dbb01d7580.png">

WojcikMike:
Will be released as a part of 1.15

PR #667, opened 2018-10-29
Adding no data message when there is no messages processed

WojcikMike:
Addresses: https://github.com/Particular/ServicePulse/issues/654

### DIFF
--- a/src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
+++ b/src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
@@ -253,6 +253,8 @@
                         </div>
                     </div>
                 </div>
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
```diff
@@ -253,6 +253,8 @@ <h5 ng-class="{active: showInstancesBreakdown}">
                         </div>
                     </div>
                 </div>
+                
+                <no-data ng-show="endpoint.instances.length == 0" message="No messages processed in this period of time."></no-data>
```
WilliamBZA: We have a `no-data` component??

Issue #445, opened 2017-06-02
Add retry button on message on *All failed messages* view


ramonsmits:
Describe your problem / suggestion / idea / feedback here.

This is the current *All failed messages* view:
![image](https://cloud.githubusercontent.com/assets/152998/26719268/24a410b8-4784-11e7-9468-58d53e6f0ac4.png)

If you want to retry just a single message you first need to tick the checkbox and then select *Retry {0} selected* which is a bit annoying.

It would be nice to just have a button to retry a message with one click.

***************

*Technical information if reporting bug:*

- **ServicePulse version:** 1.9.1

### Solution

- Show message retry button when hovering each message row

sergioc:
@ramonsmits what's the exact use case for retrying one message at a time?

WojcikMike:
Will be released with 1.15

PR #672, opened 2018-11-02
Adding retry button on one message

WojcikMike:
fixes: https://github.com/Particular/ServicePulse/issues/445

@sergioc can you have a look on this one. I am not sure if you don't want to move the button a bit

### DIFF
--- a/src/ServicePulse.Host/app/js/views/failed_messages/controller.js
+++ b/src/ServicePulse.Host/app/js/views/failed_messages/controller.js
@@ -107,6 +107,27 @@
             vm.loadMoreResults(vm.selectedExceptionGroup);
         };
 
+        vm.viewMessage = function (message) {
+            $location.path(`/failed-messages/message/${message.id}`);
+        };
+
+        vm.retryMessage = function(message, $event) {
+            toastService.showInfo("Retrying the message " + message.id + " ...");
+            serviceControlService.retryFailedMessages([message.id])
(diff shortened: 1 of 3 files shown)


src/ServicePulse.Host/app/js/views/failed_messages/controller.js
```diff
@@ -107,6 +107,27 @@
             vm.loadMoreResults(vm.selectedExceptionGroup);
         };
 
+        vm.viewMessage = function (message) {
+            $location.path(`/failed-messages/message/${message.id}`);
+        };
+
+        vm.retryMessage = function(message, $event) {
+            toastService.showInfo("Retrying the message " + message.id + " ...");
```
WilliamBZA: Does the message id mean anything to the user? Is there a better value we can show here @sergioc ?

WojcikMike:
I've hidden the button and show it only on hover. Till now It was disabled unless the row is hovered, now it is hidden.

src/ServicePulse.Host/app/js/views/failed_messages/controller.js
sergioc: There's no value in specifying message ID in the toast notification since that's the only message you're retrying. A generic message would suffice. Suggestion:

> Message retry requested

WojcikMike:
Message is changed, @sergioc please approve so that I can merge this.

PR #674, opened 2018-11-05
Filtering out system messages

WojcikMike:
fixes: https://github.com/Particular/PlatformDevelopment/issues/1881


WojcikMike:
Yeah, but as we are doing squash I decided to not invest time into fixing it

PR #675, opened 2018-11-06
Add installer connectivity test buttons

tmasternak:
## Overview
This PR introduces `Test` buttons in the SP installer that enable manual triggering of connectivity checks for ServiceControl and Monitoring instance

## Outcome
image


sergioc:
@WilliamBZA wanna take a look at the new messages included in the PR?

WilliamBZA:
Does this also stop the check from running when the user clicks next? One of the commits says so, but I can't see how that works

sergioc:
It stops the automatic check. Once the user clicks Next, it'll go straight to the next screen.

Issue #608, opened 2018-04-04
Continually showing "connected to SC" is redundant and distracting


SimonCropp:
every time i navigate to SP or refresh the page i get the following

![image](https://user-images.githubusercontent.com/122666/38284532-1ce1ac98-37ff-11e8-939b-ea8e3d87d870.png)

SP version v1.12.0  

## Solution

- [ ] eliminate duplicate notifications
- [ ] only show notifications when unable to connect
- [x] Show a permanent SC connectivity status indicator (e.g. place a red/green light somewhere TBD)

WojcikMike:
Again thanks @SimonCropp , I can see this message being of value when the connection was lost and later on reestablished. I will keep you posted how this progress

SimonCropp:
Any update on fixing this?

WilliamBZA:
It's not done yet, but I'd like to re-open this so that it does eventually get added. I feel a simple status icon is much more appropriate that an annoying pop-up

HEskandari:
Although showing connected / disconnected from SC does add _some_ value, the repetition is what bothers me the most.

![image](https://user-images.githubusercontent.com/132117/44259082-df787980-a253-11e8-8293-984c51afaa4b.png)

sergioc:
Fixed in upcoming v1.15 release

PR #677, opened 2018-11-08
Adding a toolbar that contains SC connectivity status

WojcikMike:
Fixes: https://github.com/Particular/ServicePulse/issues/608

@sergioc please have a look as you might want to style it a bit. It looks good for me, but you might think differently.

@WilliamBZA can you have a look at SC Monitoring connectivity piece from the perspective of using RXjs and from the perspective of integrating toolbar that is part of shell and sc monitoring connectivity that is part of monitoring

### DIFF
--- a/src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.monitoringConnectivityStatus.js
+++ b/src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.monitoringConnectivityStatus.js
@@ -4,8 +4,10 @@
 
 
     function controller($scope, connectivityNotifier) {
+        $scope.isSCMonitoringConnecting = true;
         connectivityNotifier.getConnectionStatusSource().subscribe(value => {
             $scope.isSCMonitoringConnected = value;
+            $scope.isSCMonitoringConnecting = false;
         });
     }
     
--- a/src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.monitoringConnectivityStatus.js
+++ b/src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.monitoringConnectivityStatus.js
@@ -3,15 +3,39 @@
     'use strict';
 
 
-    function controller($scope, connectivityNotifier) {
+    function controller($scope, connectivityNotifier, monitoringService, $interval) {
         $scope.isSCMonitoringConnecting = true;
         connectivityNotifier.getConnectionStatusSource().subscribe(value => {
             $scope.isSCMonitoringConnected = value;
             $scope.isSCMonitoringConnecting = false;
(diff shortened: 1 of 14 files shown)


sergioc:
@WojcikMike footer tweaked. However, the SC.Monitoring version indicator doesn't seem to always show in the Monitoring view. Not sure why.

WojcikMike:
Looks good,  I reverted your change to app.constants.js. With regards to not seeing this on some monitoring pages I have no idea. Could be some wierd cacheing issue.

Still waiting for William to review before merging.

WilliamBZA:
So it's not WIP anymore?

sergioc:
@WojcikMike during testing I didn't get any notifications when SP was unable to connect to SC(.monitoring). Is that to be expected?

WojcikMike:
@WilliamBZA yes, it is not a WIP anymore.

@sergioc could you be more specific? To see the status changed, go to monitoring pages and turn of SC Monitoring. The connectivity status is changed after failing to make a call to monitoring api. This kind of calls happen only on monitoring pages.

WojcikMike:
And again ready to be reviewed

src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.monitoringConnectivityStatus.js
```diff
@@ -0,0 +1,58 @@
...
+
+    function controller($scope, connectivityNotifier, monitoringService, $interval) {
+        $scope.isSCMonitoringConnecting = true;
+        connectivityNotifier.getConnectionStatusSource().subscribe(value => {
+            $scope.isSCMonitoringConnected = value;
+            $scope.isSCMonitoringConnecting = false;
+        });
+
+        var scMonitoringConnectionPing = $interval(function () {
+            var promises = monitoringService.checkConnections();
```
WilliamBZA:
```suggestion
            var promises = monitoringService.checkConnections().map(p => {
                p.then(r => {
                    connectivityNotifier.reportSuccessfulConnection(i);
                }, e => {
                    connectivityNotifier.reportFailedConnection(i);
                });
            });
            
```

Instead of manually iterating over the array

src/ServicePulse.Host/app/modules/monitoring/js/services/services.monitoring.js
```diff
@@ -62,9 +62,16 @@
                 });
         }
 
+        function checkConnections() {
```
WilliamBZA: This call doesn't "check the connections", the connection status is inferred from the result of getting the list of monitored endpoints. Looking at this file in isolation the name of this method doesn't make sense.

```suggestion
        function getMonitoredEndpoints() {
```

PR #692, opened 2018-12-05
Migrate config to webpack bundle

WilliamBZA:
I've moved the configuration from the main section into a webpack bundle. This means we can use ES6 here and have proper bundling and minification of the code. This should make our future work a bit easier.


WilliamBZA:
The second commit is also to fix the html alignment. I'm happy to revert that one if needed. I did make them separate so that the first commit is easier to review without the noise of the second.

PR #696, opened 2018-12-06
Upgrade prot about to expire

WilliamBZA:


src/ServicePulse.Host/app/modules/configuration/js/redirect/redirect.controller.js
```diff
@@ -18,9 +17,9 @@
         }, 'RedirectsUpdated');
 
         function refreshData() {
-            redirectService.getRedirects().then((redirects) => {
-                vm.redirects = redirects.data;
-            });
+            //redirectService.getRedirects().then((redirects) => {
```
WojcikMike: Is it temp change?

WilliamBZA: Yes, already removed

PR #697, opened 2018-12-07
Tweaks

WilliamBZA:


src/ServicePulse.Host/app/modules/configuration/js/configuration.route.js
```diff
@@ -4,14 +4,16 @@
...
 
-        $routeProvider.when('/configuration/endpoints', {
-            data: {
-                pageTitle: 'Monitored endpoints - Configuration'
-            },
-            template: template,
-            controller: 'ConfigurationCtrl',
-            controllerAs: 'vm'
-        });
+        $routeProvider.when('/configuration', {
```
WilliamBZA: This was the only change that had to happen to the route

src/ServicePulse.Host/app/modules/configuration/js/license/license.module.js
```diff
@@ -3,11 +3,8 @@
 
     angular.module('configuration.license', []);
 
-    require('./license.controller');
     require('./license.route');
```
WilliamBZA: Swapped around because reasons. Not really sure why this needs to be in this order, but...

PR #698, opened 2018-12-07
License page

kbaley:


src/ServicePulse.Host/app/js/services/factory.shareddata.js
```diff
@@ -54,6 +63,10 @@
             }
         });
 
+        serviceControlService.getLicenseDetails().then(function (response) {
+            license = response || license;
```
WilliamBZA: Is the `||` here in case the response fails?

src/ServicePulse.Host/app/js/services/service.formatter.js
```diff
@@ -33,6 +33,24 @@
             }
         }
 
+        function formatDate(value) {
+            return value ? moment(value).format("MMMM DD, YYYY") : undefined;
```
WilliamBZA: The screenshot from @sergioc shows MM/dd/YYYY, should we change to match?

src/ServicePulse.Host/app/js/services/services.service-control.js
```diff
@@ -21,6 +21,13 @@
             });
         }
 
+        function getLicenseDetails() {
```
WilliamBZA: Agree with moving this to it's own service. There may be times where the license comes from somewhere different to the SC endpoint in the future.

src/ServicePulse.Host/app/js/services/service.formatter.js
kbaley: That format is a bit US-centric, isn't it? I thought Europe usually used dd/MM/yyyy.

What I _really_ wanted was to use the browser's locale but that seems to be tougher than it looks with moment. Either way, I'm a bit paranoid about the MM/dd and dd/MM formats so I tend to default to something more explicit.

src/ServicePulse.Host/app/js/services/factory.shareddata.js
kbaley: Yes, I suppose. It's not something I would have considered myself but the call to `getConfiguration` right under it uses the same thing so I figured it must have come up in the past.

PR #703, opened 2018-12-13
Complete display of license page

kbaley:
This completes display of the license page based on the mockups


src/ServicePulse.Host/app/modules/configuration/js/license/license.controller.js
```diff
@@ -14,12 +14,52 @@
 
         function mapLicenseToVm(license) {
             vm.licenseType = license.license_type;
+            vm.licenseEdition = license.edition ? ", " + license.edition : "";
```
WilliamBZA: `'` instead of `"` pretty much everywhere here

src/ServicePulse.Host/app/modules/configuration/js/license/license.controller.js
```diff
@@ -14,12 +14,52 @@
...
+                vm.expiredWarningType = "warning";
+            }
+
+            vm.expirationDaysLeft = vm.isExpired ? " - expired" : formatter.getDayDiffFromToday(license.expiration_date);
+            vm.upgradeDaysLeft =
+                vm.isExpired ? " - expired" : formatter.getDayDiffFromToday(license.upgrade_protection_expiration);
+        }
+
+        function licenseMatches(status) {
+            for (var i = 1; i < arguments.length; i++) {
```
WilliamBZA: Starting from index 1? What is this, python? 

We can use destructuring to extract the first argument from the rest and then use a regular `filter` method instead of doing a for iteration. (I guess it's personal style, but it feels better to me)

```suggestion
function licenseMatches() {
    const [status, ...matches] = arguments;
    const matched = matches.find(m => m === status);
    
    return typeof matched !== "undefined";
}
```

PR #708, opened 2019-01-09
Add can't connect screen

WilliamBZA:
Adds a screen at load if ServiceControl is not accessible.

It would probably be better with a static image instead of the same animated gif.

### DIFF
--- a/src/ServicePulse.Host/app/js/app.bootstrap.js
+++ b/src/ServicePulse.Host/app/js/app.bootstrap.js
@@ -13,6 +13,9 @@
         angular.element(document).ready(function () {
             angular.bootstrap(document, ['sc']);
         });
-    }, function (error) {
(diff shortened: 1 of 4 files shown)


src/ServicePulse.Host/app/js/app.bootstrap.js
```diff
@@ -13,6 +13,9 @@
         angular.element(document).ready(function () {
             angular.bootstrap(document, ['sc']);
         });
-    }, function (error) {
+    }, function () {
+        window.document.getElementById('cantConnectMessage').style.display = 'block';
```
WilliamBZA: Have to use DOM functions as angular isn't loaded at this point yet (which is also why the `ng-if="::false"` thing works)

WilliamBZA:
Looks like this:

image

PR #719, opened 2019-02-22
Handling large number of message type in endpoint details

tmasternak:
## Overview
This PR provides browser-side paging for the message types table in endpoint details view. Specifically, it adds:
  * Ordering on message types by message type name,
  * Paginator at the bottom of the table,
  * Deep linking for page number,
  * It does not reload the table automatically when new message types become available but prompt the user to force the view reload,
  * loader view for endpoint list and details - useful when there are larger sets of message types

NOTE: this includes updating angular-ui-bootstrap to 2.5.6 in order to get desired behavior of the pager. I've checked the other places we use the package (tooltips and modals) and they look fine after update.

### DIFF
--- a/src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
+++ b/src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
@@ -20,6 +20,7 @@
         $scope.sourceIndex = $routeParams.sourceIndex;
         $scope.showInstancesBreakdown = $routeParams.tab === 'instancesBreakdown'; 
         $scope.loading = true;
+        $scope.loadedSuccessfuly = false;
         $scope.largeGraphsMinimumYAxis = largeGraphsMinimumYAxis;
         $scope.smallGraphsMinimumYAxis = smallGraphsMinimumYAxis;
 
@@ -96,6 +97,9 @@
             var selectedPeriod = $scope.selectedPeriod;
 
             subscription = monitoringService.createEndpointDetailsSource($routeParams.endpointName, $routeParams.sourceIndex, selectedPeriod.value, selectedPeriod.refreshInterval).subscribe(function (endpoint) {
+
+                $scope.loading = false;
+
                 if (endpoint.error) {
                     connectivityNotifier.reportFailedConnection($routeParams.sourceIndex);
                     if ($scope.endpoint && $scope.endpoint.instances) {
@@ -132,8 +136,6 @@
                         return 0;
                     });
 
-                    $scope.loading = false;
-
                     processMessageTypes();
 
                     $scope.endpoint.isStale = true;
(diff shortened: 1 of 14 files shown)


sergioc:
@WilliamBZA ready to review

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
```diff
@@ -34,20 +35,62 @@
...
+            var breakdownTabName = showInstacesBreakdown ? 'instancesBreakdown' : 'messageTypeBreakdown';
+
+            return `#/monitoring/endpoint/${$scope.endpointName}/${$scope.sourceIndex}?historyPeriod=${selectedPeriodValue}&tab=${breakdownTabName}&pageNo=${breakdownPageNo}`;
+        };
+
+        $scope.updateUrl = function () {
+
+            var updatedUrl = $scope.buildUrl($scope.selectedPeriod.value, $scope.showInstancesBreakdown, $scope.endpoint.messageTypesPage);
+
+            window.location.hash = updatedUrl;
```
WilliamBZA: `$window.location.hash` and inject `$window` into the controller.

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
```diff
@@ -34,20 +35,62 @@
...
+            $scope.endpoint.messageTypesForceReload();
+        };
+
+        $scope.endpoint = {
+            messageTypesPage: !$scope.showInstancesBreakdown ? $routeParams.pageNo : 1,
+            messageTypesTotalItems: 0,
+            messageTypesItemsPerPage: 10,
+            messageTypesAvailable: false,
+            messageTypesUpdatedSet: [],
+            messageTypesForceReload: function () {
```
WilliamBZA: Not a fan of this name. If it's a Force, where is the normal non-force reload? What about simply `reloadMessageTypes`?

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
```diff
@@ -130,6 +182,18 @@
         });
 
         updateUI();
+
+        $(function () {
```
WilliamBZA: This should really be a directive. Any native DOM stuff should happen inside a directive to bridge between the angular world and the DOM world.

If we refactor some of the html markup, it's not obvious that this controller needs any of the `endpoint-data-changed` etc. items in the markup. If this was a directive like `<sticky-loading-indicator></sticky-loading-indicator>` or something, it becomes much more obvious what is happening in the DOM world.

src/ServicePulse.Host/app/modules/monitoring/js/endpoint_details.controller.js
tmasternak: how about 'refreshMessageTypes`?

tmasternak:
@WilliamBZA review fixes in places. Back to you.

WojcikMike:
I changed the way how js tests are run, if a build would kick in and fail please rebase this PR on top of master which contains the necessary changes for build to pass

tmasternak:
@WojcikMike the PR has been rebased on top of master

PR #725, opened 2019-03-04
[WIP] Allow configuring SP connectivity via the web app

mauroservienti:


mauroservienti:
rebased on `master`

mauroservienti:
Was observing a very strange behavior, like if the app was using a ghost `scConfig` in some places and the new `connectionsManager` in other places. Found a file that should not be there. Removed it in https://github.com/Particular/ServicePulse/pull/725/commits/88ec9003d66a3decd72f889151be27840222abe5 as webpack is recreating that `.dist.js` file in another dist directory.

Given that default URLs are not anymore set at installation time, the `app.constants.js` doesn't need to contain the `defaultConfig` object. Moved it to `connectionsManager` to simplify retrieval of default configuration when Pulse starts for the very first time.

mauroservienti:
To be checked:

> Given that default URLs are not anymore set at installation time, the `app.constants.js` doesn't need to contain the `defaultConfig` object. Moved it to `connectionsManager` to simplify retrieval of default configuration when Pulse starts for the very first time (https://github.com/Particular/ServicePulse/pull/725/commits/2eb1e342f04c8be895c1dec7dbc198556fa32c08).

There are installer custom actions that deal with those values at install time, I'm not sure if moving them breaks the installation experience.

WilliamBZA:
:+1: for the code. Boy scout rule followed for all the places where old code was touched and the new code doesn't introduce deeper coupling to angular.

mauroservienti:
Closing in an attempt to resurrect TC, will open a new PR.

PR #727, opened 2019-03-05
Differentiate between infinite scroll loads

### LINKED ISSUE #726: Inifnite scroll on Failed Messages screen results in double initial loads
Bug:

When loading all the failed messages for a group, the initial load happens twice if the list of messages is less than the size of the screen. This results in duplicate messages being shown on the first page load. Users have to change the sort order (even changing to the currently selected sort order) to get the messages not to duplicate.

Steps to repro:

1. Create a failed group with 1 failed message in
2. Open the failed messages link for that group
3. See the single message being duplicated

image

The reason this happens is that the infinite scroll is being triggered causing a second request to be fired off before the first one is completed.

***************

- **ServicePulse version:** 1.16.1
- **ServiceControl version:** Any
- **Browser version:** Any

WilliamBZA:
Fixes #726 

Adds a flag that can be used to differentiate between an infinite scroll request or a normal request. If the initial load is still being performed and an infinite scroll load happens at the same time, the infinite scroll request is abandoned.

### DIFF
--- a/src/ServicePulse.Host.Tests/ServicePulse.Host.Tests.csproj
+++ b/src/ServicePulse.Host.Tests/ServicePulse.Host.Tests.csproj
@@ -89,6 +89,7 @@
     <Content Include="tests\js\services\services.messageTypeParser.spec.js" />
     <Content Include="tests\js\services\services.monitoring.spec.js" />
     <Content Include="tests\js\views\failed_groups\controller.spec.js" />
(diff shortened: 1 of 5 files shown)


WilliamBZA:
I tried to come up with a way to add a unit test for this @WojcikMike, but couldn't work it out. If you have a chance I'd love to pair with you on this to see how to achieve it.

PR #734, opened 2019-03-21
Add supported browsers

sergioc:


WilliamBZA:
Should this be RFC'd?

sergioc:
Not sure about the need to RFC. I think the below rationale should be pretty safe while making things easy for development.

Based on @WojcikMike comments, I did make Chrome's latest version the supported one since the majority of users are on recent versions thanks to auto-updating.

I'll base Edge's supported version on the rendering engine version, as opposed to app versions. That will keep compatibility easier to maintain while keeping happy cautious upgraders/users in restricted IT environments. Choosing Firefox latest LTS release as opposed to latest standard release for the same reasons.

IE remains specified at IE11 for as long as Microsoft supports.

Safari's user base is likely tiny, so keeping it easy by supporting the latest only.

PR #742, opened 2019-04-05
Switch to new project system

bording:
This converts everything over to the new project system, which cleans up a lot of stuff.

BTW, the smoke test projects are still using NSB and the old plugin packages. Those should really be updated!


WojcikMike:
@WilliamBZA can you have a look as well.

WojcikMike:
@WilliamBZA @mauroservienti would it be highly inconvenient if we merge this PR in? I wonder how much pain that would cause to your PR ?

mauroservienti:
I’m pretty sure we touched none of these files. Should be an easy rebase. Last famous words 😎🤪

PR #746, opened 2019-05-13
Allow configuring SP connectivity via the web app

### LINKED ISSUE #690: Add multiple instances feature like servicecontrol
I’m order to have servicepulse be installed on 1 machine and we for multiple environments we have to jump through the iis steps outlined in the docs.  Is it possible to take the same approach as service control and have multiple instances of service pulse that run on a different port?

mauroservienti:
Resolves #690


mauroservienti:
@WilliamBZA I addressed all your comments

mauroservienti:
With https://github.com/Particular/ServicePulse/pull/747 migratgion from previous configuration mechanism is supported, and it's still supported to input default connection settings at install time for customers that don't need to change connection later.

PR #747, opened 2019-05-21
Add support to migrate existing URLs configuration and re-add support for URLs configuration in the installer

mauroservienti:
ServicePulse uses the `app.constants.js` file to set default settings. Up to 1.19 the file is an AngularJs module, and thus it requires Angular to be loaded. The new `connectionsManager` on the other hand lives outside AngularJs as it is needed during the bootstrapping process to get license details from ServiceControl.

This PR:

- changes the `app.constants.js` format to be non-Angular dependant
- makes so that the `app.constants.js` simply sets a `window` variable with default configuration values, using the same structure as `1.19` (but outside of AngularJs)
- makes so that the `app.constants.js` is loaded by `Index.html` as first script so that all the dependencies can find the values they expect where they expect them

At install time the installer collects information from the user about default URLs to use, and store them in memory. During the physical installation process the defined URLs are passed as command line arguments to ServicePulse.Host.exe that via the `ExtractAndUpdateConstantsCommand`:

- extracts from resources the `app.constants.js`
- sets the defined values
- sets the current assembly version as ServicePulse version

This PR:

- re-enables the installer screen(s) that allowed users to configure URLs
- changes the `ExtractAndUpdateConstantsCommand`, that is responsible to write configured values, to update values using the new file format

#### Before merging

- [x] Test the installer (that is created by TC once this PR is compiled)

#### Once merged

- [x] revert https://github.com/Particular/docs.particular.net/pull/4343/commits/15d43083f8bc48a3cc5b7c71126ddbd87c36201a as command line parameters will still be supported

### DIFF
--- a/src/ServicePulse.Host/Commands/ExtractAndUpdateConstantsCommand.cs
+++ b/src/ServicePulse.Host/Commands/ExtractAndUpdateConstantsCommand.cs
@@ -59,7 +59,7 @@
         {
             var appJsPath = Path.Combine(directoryPath, "js/app.constants.js");
             var appJsCode = File.ReadAllText(appJsPath);
-            var updatedContent = Regex.Replace(appJsCode, @"(constant\(\s*'version'\s*,\s*')(.*?)(')", "${1}" + GetFileVersion() + "$3");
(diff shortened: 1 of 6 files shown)


mauroservienti:
tested, it works 😄

mauroservienti:
https://github.com/Particular/docs.particular.net/commit/15d43083f8bc48a3cc5b7c71126ddbd87c36201a reverted. Merging.

PR #763, opened 2019-07-24
Allow to control (again) the showPendingRetry setting

mauroservienti:
As [stated in the documentation](https://docs.particular.net/servicepulse/intro-pending-retries) by tweaking the `app.constants.js` it should be possible to enable the "show pending retry" feature in ServicePulse. This was broken due to way Pulse was changed to introduce the ability to control che ServiceControl connection. This PR fixes the regression by re-enabling the ability to turn-on the "show pending retry" feature.

## PoA

- [x] test that Pulse and the feature work as expected when using the installer
- [ ] update documentation to reflect the change https://github.com/Particular/docs.particular.net/pull/4443
- [ ] release and announce

### DIFF
--- a/src/ServicePulse.Host.Tests/ApprovalFiles/APIApprovals.PlatformSampleApprovals.approved.txt
+++ b/src/ServicePulse.Host.Tests/ApprovalFiles/APIApprovals.PlatformSampleApprovals.approved.txt
@@ -2,5 +2,6 @@ window.defaultConfig = {
     default_route: '/dashboard',
     version: '1.2.0',
     service_control_url: 'http://localhost:33333/api/',
-    monitoring_urls: ['http://localhost:33633/']
+    monitoring_urls: ['http://localhost:33633/'],
+    showPendingRetry: false
 };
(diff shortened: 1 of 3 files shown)


WilliamBZA:
@mauroservienti What were the symptoms of if being broken? I'm in 2 minds about adding the setting to the config as a default false when a `.constant('showPendingRetry', (window.defaultConfig.showPendingRetry || false))` would also work.

On the one hand:

* It's easier for me to remember what the setting is called
* It's less likely the setting will be forgotten in the future again because it's visible

On the other hand:

* It's not a feature we really want to expose to users, by having it there they can see it.

On the middle-hand:

* I don't really know what the concrete use case for the feature is. It has always felt to me like something that could be better solved.

So I guess I'm leaning towards adding the default values, I just need a little nudge to push me over.

mauroservienti:
Doco says "change `.constant('showPendingRetry', false)` to true` in `app.constants.js` but the `app.constants.js` format changed and that `.constant('showPendingRetry', false)` doesn't exist anymore and there is no place to add it in `app.constant.js. This means that there is no way to enable the feature.

I'm fine with not adding anything to `app.constant.js` and use your code:

```
.constant('showPendingRetry', (window.defaultConfig.showPendingRetry || false))
```

This means more changes to doco. 

>It's not a feature we really want to expose to users, by having it there they can see it.
>I don't really know what the concrete use case for the feature is. It has always felt to me like something that could be better solved.

The feature is documented, this PR originates from a chat with a customer that upgraded to latest Pulse and failed to re-enable the feature.

mauroservienti:
@WilliamBZA updated.

mauroservienti:
tested: works as expected.

PR #783, opened 2019-09-26
Fixes incorrect headers issues

### LINKED ISSUE #663: Using the search API to find headers can result in the wrong headers being displayed for a message
Retrieving message headers uses the ServiceControl search API:

https://github.com/Particular/ServicePulse/blob/d373dc31a885c11311597029e544d34003532d5a/src/ServicePulse.Host/app/js/services/services.service-control.js#L117-L124

The search API will return more than one message if they each contain the searched for message id.

When headers are retrieved only the first message returned is used:

https://github.com/Particular/ServicePulse/blob/d373dc31a885c11311597029e544d34003532d5a/src/ServicePulse.Host/app/js/views/message/controller.js#L58-L60

This can cause ServicePulse to display the incorrect headers for a given message, causing confusion.

Normally this isn't a problem for a user, however in scenarios where the `MessageFailed` event is subscribed to, and itself fails, the original failed message and the `MessageFailed` event message can cause this scenario to happen. The `MessageFailed` event contains a property that indicates the id of the failed message, which is why it is returned in the search results and can be mistakenly used.

mikeminutillo:
Fixes #663

### DIFF
--- a/src/ServicePulse.Host/app/js/services/services.service-control.js
+++ b/src/ServicePulse.Host/app/js/services/services.service-control.js
@@ -147,7 +147,7 @@
                 var matchingMessage = response.data.find(function(m) { return m.message_id === messageId; });
                 return {
                     headers: matchingMessage.headers
-                }
+                };
             });
         }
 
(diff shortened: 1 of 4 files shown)


src/ServicePulse.Host/app/js/services/services.service-control.js
```diff
@@ -144,9 +144,10 @@
         function getMessageHeaders(messageId) {
             var url = uri.join(scu, 'messages', 'search', messageId);
             return $http.get(url).then(function(response) {
+                var matchingMessage = response.data.find(function(m) { return m.message_id === messageId });
```
WilliamBZA: Missing semicolon for the return statement. I've added it.

PR #784, opened 2019-09-27
Fixes missing checkbox on adding redirect

### LINKED ISSUE #706: Redirect is not visible after creating unless page is refreshed (F5)

### LINKED ISSUE #707: Checkbox not showing in latest chrome when creating a redirect
When creating a "Redirect" the checkbox is missing to indicate that failed messages that match this redirect should be automatically retried.

image

***************

- **ServicePulse version:** 
v1.15.1
- **ServiceControl version:** 
v3.5.0
- **Browser version:** 
71.0.3578.98 (64-bit)

mikeminutillo:
Fixes #707

### DIFF
--- a/src/ServicePulse.Host/app/css/particular.css
+++ b/src/ServicePulse.Host/app/css/particular.css
@@ -1251,6 +1251,10 @@ input[type=checkbox] {
     visibility: hidden;
 }
 
+input#immediatelyRetry {
+    visibility: visible;
+}
+
 .checkbox {
     margin-top: 1px;
     margin-left: 1px;


mikeminutillo:
Not sure if this is really the best approach. You can see just above my change where we hide ALL checkboxes so all I did was turn this one back on specifically. All of the other ones in the system are either On/Off toggle buttons or a part of a multi-select grid. All of those get re-rendered as a different control.

WilliamBZA:
@sergioc might be a better reviewer here than me. @sergioc could you take a look?

Issue #756, opened 2019-06-24
Should be possible to remove server from endpoint heartbeat list


nikolaimaric:
When decomissioning a server/endpoint, there is no way to permanently remove it from the list of inactive endpoints in the "Endpoint heartbeats" section. The only option is to hide the endpoint, that is then restored as inactive every time ServicePulse is upgraded to a newer version.

![image](https://user-images.githubusercontent.com/1325611/66549161-fe665900-eb42-11e9-9d38-9de813519bcd.png)

One option could be to add a `Permanently remove` button near the `on/off` toggle:

![image](https://user-images.githubusercontent.com/1325611/66549396-7c2a6480-eb43-11e9-907c-eace1dd8868b.png)

Depends on https://github.com/Particular/ServiceControl/issues/1561

---

_Original description_

When we decomission a server, there is no way to permanently remove it from the endpoint monitoring. It can be disabled, but it is enabled again every time we upgrade ServicePulse. (Spoke with @janovesk about this at NDC Oslo last week, he suggested I file an issue)

mauroservienti:
Hi @nikolaimaric,

thanks for raising this and sorry for the issue. This issue is now monitored by the relevant group, it'd be probably prioritized for the next ServicePulse maintenance release. For which unfortunately I don't have an ETA.

Please subscribe to the issue updates, we'll post here further details.

oskar:
Any news on this? We are unable to rely on ServicePulse heartbeats for production system monitoring until there is a way to remove decommissioned endpoints.

mauroservienti:
I'm looking into this now, and I'm trying to scope the issue to better understand what needs to be done.

@nikolaimaric in your comment you refer to Monitoring, are you talking about the Monitoring section of ServicePulse?

![image](https://user-images.githubusercontent.com/1325611/66547196-fd332d00-eb3e-11e9-98d8-d8491997bed7.png)

@oskar, in your case, I guess you're referring to the fact that in the Heartbeats section once an Endpoint has failed reporting Heartbeat it's marked as inactive and there isn't really a way to **permanently** remove it:

![image](https://user-images.githubusercontent.com/1325611/66547339-4e432100-eb3f-11e9-89c1-51a6bc123e26.png)


Correct?

oskar:
We have the exact same issue as @nikolaimaric.

nikolaimaric:
[...]

PR #791, opened 2019-10-11
Allow for the deletion of endpoints

WilliamBZA:
Fixes https://github.com/Particular/ServicePulse/issues/756

2 commits. One to change the config service to es6, and the second to actually implement the required change.


tmasternak:
@WilliamBZA should we delete the branch?

WilliamBZA:
Done.

PR #800, opened 2019-10-31
Update dockerfile to current

WilliamBZA:
Updates the dockerfile from the `docker-container` branch to use the current.

Easier to create the file again and delete the previous branch than deal with the rebase changes between them.


adamralph:
> The configuration tab when SP opens:

🤦‍♂ of course! 😊 

I guess it would be nicer to tell it during deployment, to save the manual step, but ROI on finding out how to do that, at least now, isn't worth it. 👍

PR #803, opened 2019-11-12
Fix typo when SC not available

evolvedlight:


janovesk:
Thanks @evolvedlight!

PR #821, opened 2020-02-06
Replace ServicePulse self-hosting stack from NancyFX to OWIN

janovesk:


src/ServicePulse.Host/Owin/EmbeddedFileFinder.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using ServicePulse.Host.Owin.Microsoft;
+
+namespace ServicePulse.Host.Owin
+{
+    public static class EmbeddedFileFinder
+    {
+        private static readonly Assembly Assembly = Assembly.GetExecutingAssembly();
```
mauroservienti: Is this what's really wanted? Is `Assembly.GetExecutingAssembly();` returning the assembly that contains this class? which makes so that embedded files will be searched only in this assembly.

janovesk: We want to search inside the ServicePulse.Host.exe assembly only, so I think this works fine. 

Or maybe I don't completely understand your concern, @mauroservienti ?

WilliamBZA: I think that's the expected behaviour. This class is in the host assembly, and all the js, html, and css files are bunclded in the host assembly (bleurgh).

src/ServicePulse.Host/Owin/EmbeddedFileFinder.cs
```diff
@@ -0,0 +1,36 @@
...
+        
+        public static bool FindEmbeddedFile(string filePath, out IFileInfo fileInfo)
```
WilliamBZA: returns `bool` and uses `out`? 🤢

src/ServicePulse.Host/Owin/FileOnDiskFinder.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.IO;
+using ServicePulse.Host.Owin.Microsoft;
+
+namespace ServicePulse.Host.Owin
+{
+    public static class FileOnDiskFinder
+    {
+        public static bool FindFile(string filePath, out IFileInfo fileInfo)
```
WilliamBZA: returns `bool` and uses `out`? 🤢

src/ServicePulse.Host/Owin/Microsoft/FileExtensionContentTypeProvider.cs
```diff
@@ -0,0 +1,1517 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
```
WilliamBZA: I can't find the License.txt that goes along with this copyright

src/ServicePulse.Host/Owin/Microsoft/FileInfo.cs
```diff
@@ -0,0 +1,116 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
```
WilliamBZA: Ditto on missing license.txt

src/ServicePulse.Host/Owin/Microsoft/StaticFileContext.cs
```diff
@@ -0,0 +1,224 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
```
WilliamBZA: Double ditto

src/ServicePulse.Host/Owin/EmbeddedFileFinder.cs
janovesk: `private static readonly Assembly Assembly = typeof(EmbeddedFileFinder).Assembly;`  is slightly more robust if this code is somehow suddenly not hosted in `ServicePulse.Host.exe`, but the self-hosting nature of ServicePulse makes that very unlikely?

src/ServicePulse.Host/ServicePulse.Host.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net40</TargetFramework>
+    <TargetFramework>net45</TargetFramework>
```
WilliamBZA: What a pity we couldn't change this to `netcoreapp3.0` as part of this taskforce, but I'm ok with it for now.

src/ServicePulse.Host/Owin/EmbeddedFileFinder.cs
mauroservienti: > We want to search inside the ServicePulse.Host.exe assembly only, so I think this works fine.

This answers my question, @janovesk. Thanks.

PR #824, opened 2020-02-12
Imitate Nancy localhost url behavior to avoid url access issues

janovesk:


src/ServicePulse.Host/Owin/UrlHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace ServicePulse.Host.Owin
+{
+    public static class UrlHelper
+    {
+        public static string RewriteLocalhostUrl(string url)
+        {
+            var newUrl = new UriBuilder(url);
+
+            if (newUrl.Host.Contains(".")) //not localhost
```
WojcikMike: I am not part of the TF and I was not asked to be a reviewer but this is a bit optimistic. A valid URL could be without a dot. You could have http://particular/servicepulse 

Given that this if is suppose to check if it includes localhost why don't make if newUrl not contain localhost

janovesk: This mimics the Nancy logic 100%, which is good enough for this fix.

WojcikMike: :+1:

PR #831, opened 2020-03-11
Resolve directory traversal vulnerability in static file middleware

mikeminutillo:


src/ServicePulse.Host.Tests/Owin/StaticMiddlewareTests.cs
```diff
@@ -8,21 +8,21 @@ namespace ServicePulse.Host.Tests.Owin
...
-                Request =
-                {
-                    Path = new PathString("/filename.unknown"),
-                    Method = "GET"
-                }
-            };
-            middleware.Invoke(context);
-            Assert.AreEqual(("application/octet-stream"), context.Response.ContentType);
-        }
+        //[Test]
```
WojcikMike: @mikeminutillo @janovesk is there any reason that this test was not removed?

mikeminutillo: The intention was to put something back in it's place after the security advisory was published. Maybe an approval test of all extensions in the assembly along with the mime-type or something.

PR #838, opened 2020-03-18
Use Particular.Packaging instead of GitVersionTask

bording:


andreasohlund:
Version on TC looks fine so I'm going ahead and merging this

Issue #812, opened 2020-01-16
Clicking failed message on dashboard should navigate to message body


TraGicCode:
When click on a failed message on the dashboard

![image](https://user-images.githubusercontent.com/6224270/72536804-98275b80-3840-11ea-9873-167c15d5bce0.png)

You are taken to the failed message groups page.  This isn't very helpful and the expectation from a user is that you are taken to the message body to see the full message along with having an understanding of what endpoint caused the error message.

ServicePulse Version: 1.23.0

WilliamBZA:
This will be fixed and released in version 1.25.0. Expect it in roughly 2 weeks.

WilliamBZA:
Release is out! You can download [1.25.0](https://github.com/Particular/ServicePulse/releases/tag/1.25.0) and this should be done.

PR #851, opened 2020-04-20
Adding an redirect to event related to failed messages.

WojcikMike:
Addresses: https://github.com/Particular/ServicePulse/issues/812

### DIFF
--- a/src/ServicePulse.Host/app/js/views/event_log_items/eventLogItems.js
+++ b/src/ServicePulse.Host/app/js/views/event_log_items/eventLogItems.js
@@ -28,7 +28,7 @@
                     break;
                 case 'MessageFailures':
                     var newlocation = '/failed-messages/groups';
-                    if(eventLogItem.related_to && eventLogItem.related_to[0].search('message') > 0) {
+                    if (eventLogItem.related_to && eventLogItem.related_to[0].search('message') > 0) {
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/app/js/views/event_log_items/eventLogItems.js
```diff
@@ -27,7 +27,11 @@
                     $location.path('/endpoints');
                     break;
                 case 'MessageFailures':
-                    $location.path('/failed-messages/groups');
+                    var newlocation = '/failed-messages/groups';
+                    if (eventLogItem.related_to && eventLogItem.related_to[0].search('message') > 0) {
+                        newlocation = '/failed-messages' + eventLogItem.related_to[0];
```
WilliamBZA, line 32: Does `related_to` already include a leading `/`?

WilliamBZA: Answering myself: Yes. Yes it does.

WojcikMike: yeah it does, what it has is /message/<guid>

I don't like the fact that we assume it will had 0 index in the array. And that it contains such string that is why I wrapped it in the if, so that we fall back to the good old: redirect to the group view

PR #852, opened 2020-04-20
Add cache-control for images

### LINKED ISSUE #236: No cache-control header on static content
Although *Etag* and *Last-Modified* headers are present the browser is still making unnecessary roundtrips as these files are static. Most static files have either a versioned folder or file so these files can be configured to be cached indefinitely. All other remaining files should be versioned to benefit the most from caching but otherwise they could atleast be cached for an hour or maybe a day.

WilliamBZA:
Fixes #236

Adds cache-control headers for image files only. These are the files least likely to change over time. And until we have a cache-busting strategy in place, I'm not confident enough to extend this to other assets yet.

### DIFF
--- a/src/ServicePulse.Host/Owin/Microsoft/StaticFileContext.cs
+++ b/src/ServicePulse.Host/Owin/Microsoft/StaticFileContext.cs
@@ -25,6 +25,7 @@ namespace ServicePulse.Host.Owin.Microsoft
         private string _lastModifiedString;
         private string _etag;
         private string _etagQuoted;
+        private bool _isStatic;
 
         private PreconditionState _ifMatchState;
         private PreconditionState _ifNoneMatchState;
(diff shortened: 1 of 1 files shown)


WilliamBZA:
@ramonsmits based on the discussion earlier today, how do you think we could change this?

WilliamBZA:
Ping @ramonsmits

ramonsmits:
Kicked the build to resolve transient error on buildserver.

src/ServicePulse.Host/Owin/Microsoft/StaticFileContext.cs
```diff
@@ -154,6 +158,15 @@ private void ComputeIfModifiedSince()
             }
         }
 
+        public void ApplyCacheControlHeaders()
+        {
+            if (_isStatic)
+            {
+                // 604800 = 1 week
+                _response.Headers.Set("Cache-Control", "public,max-age=604800,immutable");
```
ramonsmits, line 166: The intent for `immutable` is for versioned resources that really will never change. This is an improvement but the real solution would be versioned url's. Maybe that for another future release.

ramonsmits:
Closing, better implementation be done in the future via post build step.

Issue #212, opened 2015-11-07
Angular template versioning


colin-higgins:
task-force @colin-higgins @petersgiles
### Goal:

Introduce template versioning with a query string to ensure users have a smooth upgrade experience whenever we release new versions of ServicePulse.
### Plan:

Shared function that uses an angular.constant to tack the version number onto the url for the template.
We could also have a debug flag that refreshes the template every time in development.

petersgiles:
Yes to this! I'm in

petersgiles:
@colin-higgins did you have a specific way of doing this (using an existing framework)?

colin-higgins:
```
angular.module('sp').value('applyVersion', function (templateUrl) { return templateUrl + '?v=1.3.5'; });
```

That is oversimplified obviously, but it's approximately what I have done before.

johnsimons:
Shouldn't we handle this using the normal expiration headers in http?

@SimonCropp has solved this nicely in docs, majority of time it returns 304s.

colin-higgins:
@johnsimons The reason I would argue for the query string method is that it's tied to the version and automatically handles any upgrades seamlessly. Otherwise it leaves the cache alone. How would you implement the expiration headers?

johnsimons:
Sounds good @colin-higgins, your way will work best especially if users host it in IIS where we don't control the headers.

WilliamBZA:
Blast from the past, but a first step to this will finally be released in 1.25.0.

PR #854, opened 2020-04-20
Prevent cached files from disrupting upgrade process

WilliamBZA:
Fixes https://github.com/Particular/ServicePulse/issues/212

Adds a `?v=timestamp` to all js files as part of the webpack build.

Ideally we'd use the webpack chunked hash directly and generate a `dist/index.html`, but this won't cover the files that are not yet in webpack bundles.

We'd also have to change the build pipeline to host the `dist/index.html` file instead of the `/index.html` file.

This PR should cover our needs for a cache-buster, but we should create an issue to update the build to be ready to use a `/dist/index.html` file.

### DIFF
--- a/src/ServicePulse.Host/app/index.html
+++ b/src/ServicePulse.Host/app/index.html
@@ -141,111 +141,111 @@
         }
     </script>
 
-    <script src="js/app.constants.js"></script>
+    <script src="js/app.constants.js?v=1587408385690"></script>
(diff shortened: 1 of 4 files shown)


WilliamBZA:
Hold off on review for now. I want to tweak it a little.

WilliamBZA:
This is ready for review now

Issue #266, opened 2016-02-11
As a user I expect to be able to sort failed message groups


mauroservienti:
In the failed message groups view would be very handy to allow the user to sort groups given the available columns.

johnsimons:
I am pretty sure the current view is sorted by date of last failure per group, in other words, a group that has recent failures will show above olders failures on another group.

Does this make sense ?

Not sure what the value is in providing sort for all columns, can you explain the use case ?

mauroservienti:
> I am pretty sure the current view is sorted by date of last failure per group, in other words, a group that has recent failures will show above olders failures on another group.
Does this make sense ?

As a default option, yes it makes sense.

> Not sure what the value is in providing sort for all columns, can you explain the use case ?

In the scenario @ramonsmits and I were facing there were many failed messages group and we found ourselves scanning the groups several time looking for correlation across messages, would have been much easier to be able to sort by exception message (the group title) for example, or in some cases by message count.

johnsimons:
> we found ourselves scanning the groups several time looking for correlation across messages

What was the use case for this ?
I'm asking because maybe there is a different way we can achive this.

mauroservienti:
in that case we were debugging ServiceControl on a production environment, but given the way messages are goup end users may find themselves doing the same thing. The issue with grouping is that, as far as we have understood, we group by default by `exception message` causing some weird behavior.

ramonsmits:
In most environments it makes the most sense to retry old groups first as these would normally be retried after deploying a fix. After a fix its unlikely that new messages are added to the group meaning this group will slide down while it is the best candidate to retry so should be, IMHO, at the top.

PR #858, opened 2020-04-21
Add sort on failed message groups

WilliamBZA:
Closes https://github.com/Particular/ServicePulse/issues/266

Adds a sort option on the failed message groups.

Adds sort for all the main fields on the group:

* Name
* First failure
* Last failure
* Last retry
* Number of messages

image

It also saves the sort so that the next time the user views the page it restores the previously selected sort.


WilliamBZA:
Rebased

PR #859, opened 2020-04-21
Fix gitignore

WilliamBZA:


WilliamBZA:
@adamralph your commit breaks it.

image

PR #860, opened 2020-04-22
Add a version check and upgrade link for outdated Monitoring Instances

WilliamBZA:
Adds a check that displays the monitoring version and if there is an upgrade available.

image

Requires https://github.com/Particular/ServiceControl/pull/1972 to function, but if https://github.com/Particular/ServiceControl/pull/1972 is not merged this gracefully degrades to the current behaviour.


WilliamBZA:
Ping @ramonsmits and @WojcikMike  and @sergioc

WilliamBZA:
@ramonsmits how about we merge this as is, and add a small tasks board to combine the connection icons and the version check?

PR #865, opened 2020-04-22
Distinction between "no connection" and "no data" and improved troubleshooting hints

### LINKED ISSUE #755: The message when monitoring is not available is not clear
When you go to the monitoring tab, if there is no data to display you see the following:

image

The third point is not very clear. I believe the original intent was to capture endpoints that don't have the monitoring plugin installed. They would normally appear on this page if they are sending data to ServiceControl. If there is no data at all to display, then you see this message.

I'm not sure that it adds any value. Perhaps we should consider dropping this point altogether. It may make more sense to have a generic "Why don't my endpoints appear?" link off to the documentation website with some troubleshooting guidelines.

ramonsmits:
Resolves #755

No connectivity:
image

No data:
image

### DIFF
--- a/src/ServicePulse.Host/app/modules/monitoring/js/monitored_endpoints.controller.js
+++ b/src/ServicePulse.Host/app/modules/monitoring/js/monitored_endpoints.controller.js
@@ -11,7 +11,9 @@
         rx,
         $filter,
         smallGraphsMinimumYAxis,
-        connectivityNotifier) {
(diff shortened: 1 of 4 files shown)


WilliamBZA:
Are we going to add a Connecting/loading screen for while it's trying to connect to the server?

sergioc:
There's already a loading indicator which can keep showing until there's data

Issue #744, opened 2019-04-24
Failed msg group name missing in list of last 10 completed retry requests


sergioc:
<img width="705" alt="image" src="https://user-images.githubusercontent.com/493451/56661626-74185e80-66a2-11e9-97fc-b8e7ce10ecec.png">

***************

*Technical information if reporting a bug:*

- **ServicePulse version:** 1.19.0
- **ServiceControl version:** 3.6.9

WojcikMike:
@sergioc I am almost sure that we display lack of title if you retry a single message. Can you confirm that? Retry a group and later a single message and see what is presented in the 10 completed retry requests

sergioc:
Seems to be the case:

<img width="983" alt="image" src="https://user-images.githubusercontent.com/493451/56880037-7dffe000-6a5b-11e9-8a11-5a8b5fe58884.png">

PR #873, opened 2020-04-27
Displaying a static text for a retries consisting of selection or a s…

WojcikMike:
…ingle message

Solves: https://github.com/Particular/ServicePulse/issues/744

image

The group name is empty when retry is related to a selection of messages. And this is what I tried to represent here. 
I am waiting for a good name for such scenario:


WojcikMike:
We will not pursue that solution, but instead change a ServiceControl to fill the name with something of value (like exception type)

PR #877, opened 2020-04-29
Show the number of instances with connectivity problems on the monitoring overview screen

ramonsmits:
image

Depends on https://github.com/Particular/ServiceControl/pull/1989

### DIFF
--- a/src/ServicePulse.Host/app/modules/monitoring/views/monitored_endpoints.html
+++ b/src/ServicePulse.Host/app/modules/monitoring/views/monitored_endpoints.html
@@ -81,8 +81,13 @@
                                         <span class="warning" ng-if="endpoint.isScMonitoringDisconnected">
                                             <i class="fa pa-monitoring-lost endpoints-overview" uib-tooltip="Unable to connect to monitoring server"></i>
                                         </span>
-                                        <span class="warning" ng-if="endpoint.isStale">
+                                        <span class="warning" ng-if="endpoint.isStale && (endpoint.connectedCount==null || endpoint.connectedCount==0)">
                                             <i class="fa pa-endpoint-lost endpoints-overview" uib-tooltip="Unable to connect to instance"></i>
+                                        </span>
+                                        <span class="warning" ng-if="endpoint.isStale && (endpoint.connectedCount>0)">
+                                            <a ng-href="{{getDetailsUrl(endpoint)}}&tab=instancesBreakdown">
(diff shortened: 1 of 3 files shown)


src/ServicePulse.Host/ServicePulse.Host.csproj
```diff
@@ -36,4 +36,10 @@
     <EmbeddedResource Include="Hosting\Help.txt" />
   </ItemGroup>
 
+  <ItemGroup>
```
WilliamBZA, line 39: Does this need to be explicitly added? I think it should automatically get embedded based on line 31.

src/ServicePulse.Host/app/modules/monitoring/js/monitored_endpoints.controller.js
```diff
@@ -81,7 +81,7 @@
                     if (endpoint.error) {
                         connectivityNotifier.reportFailedConnection();
                         if ($scope.endpoints) {
-                            $scope.endpoints.forEach((item) => item.isScMonitoringDisconnected = true);
+                            $scope.endpoints.forEach((item) => {item.isScMonitoringDisconnected = true;item.disconnectedCount=10;});
```
WilliamBZA, line 84:
```suggestion
                            $scope.endpoints.forEach((item) => {
                                item.isScMonitoringDisconnected = true;
                                item.disconnectedCount=10;
                            });
```

src/ServicePulse.Host/app/modules/monitoring/views/monitored_endpoints.html
```diff
@@ -81,9 +81,14 @@ <h1>Endpoints overview</h1>
                                         <span class="warning" ng-if="endpoint.isScMonitoringDisconnected">
                                             <i class="fa pa-monitoring-lost endpoints-overview" uib-tooltip="Unable to connect to monitoring server"></i>
                                         </span>
-                                        <span class="warning" ng-if="endpoint.isStale">
+                                        <span class="warning" ng-if="endpoint.isStale && (endpoint.connectedCount==null || endpoint.connectedCount==0)">
```
WilliamBZA, line 84:
```suggestion
                                        <span class="warning" ng-if="endpoint.isStale && endpoint.connectedCount">
```

Type coalescing in JavaScript will convert the check to:

* Does endpoint.connectedCount have a defined value?
* Is the value non-null?
* Is the value non-zero?
* Is the value a non-empty string?

WilliamBZA:
This does mean that if the user is on an old version of SC, they won't see the current warning icon at all.

src/ServicePulse.Host/ServicePulse.Host.csproj
ramonsmits: @WilliamBZA strange.. that must be removed....Pretty sure I did that

PR #931, opened 2020-09-14
CustomChecks view - infinite scroll and initial load race condition fix

tmasternak:
Connected to https://github.com/Particular/ServiceControl/issues/2086

### Problem

```javascript
var page = 1;

$scope.loadMoreResults = function () {
    if ($scope.loadingData) {
        return;
    }

    $scope.loadingData = true;
    load(page++);
};

$scope.dismiss = function (row) {
    serviceControlService.dismissCustomChecks(row);
};

var notifier = notifyService();
notifier.subscribe($scope, reloadData, 'CustomChecksUpdated');

function reloadData() {
    page = 1;
    $scope.loadingData = true;
    $scope.model = { data: [], total: 0 };
    $scope.disableLoadingData = false;
    load(page);
}
```

It is possible for `loadMoreResults` to be triggered by the UI (by the infinite scroll) just before the `reloadData` is called via the notification service. In such a case, there are 2 outstanding requests that will both update the list of custom checks. If that happens during the initial page load the first page will be duplicated. If it happens further in the lifecycle of the page the view will end up showing the data from the first page and some further page (depending on how far a user scrolled and how much data was available).

### Solution

This PR introduces `reloadCount` that is incremented before each reload request. The `load` function will apply the data returned from the backend only if the `reloadCount` did not change between the request was made and the response was returned.

### DIFF
--- a/src/ServicePulse.Host/app/js/views/custom_checks/customChecks.controller.js
+++ b/src/ServicePulse.Host/app/js/views/custom_checks/customChecks.controller.js
@@ -10,6 +10,7 @@
         $scope.model = { data: [], total: 0 };
         $scope.loadingData = false;
         $scope.disableLoadingData = false;
+        $scope.reloadCount = 0;
 
         var page = 1;
 
@@ -19,7 +20,7 @@
             }
 
             $scope.loadingData = true;
-            load(page++);
+            load(page++, $scope.reloadCount);
         };
 
         $scope.dismiss = function (row) {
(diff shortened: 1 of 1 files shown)


SzymonPobiega (approved):

src/ServicePulse.Host/app/js/views/custom_checks/customChecks.controller.js
```diff
@@ -34,12 +35,19 @@
...
+            $scope.reloadCount++;
+
+            load(page, $scope.reloadCount);
         }
 
-        function load(page) {
+        function load(page, reloadCount) {
             serviceControlService.getFailingCustomChecks(page).then(function (response) {
 
+                //If the load was called before last result reload it should be discarded
```
mikeminutillo, line 46: Should this decrement the page number as well?

mikeminutillo (approved):

src/ServicePulse.Host/app/js/views/custom_checks/customChecks.controller.js
tmasternak: I don't think so. Whenever `reloadData` is called the page number is [set back to 1](https://github.com/Particular/ServicePulse/pull/931/files#diff-97d95757170a4ea9d39b4fae0f079725R34). So the next load caused by the scroll will start from 2. All reloads caused by the scroll that were done when before latest refresh will get discarded.

PR #932, opened 2020-09-15
Events page size

### LINKED ISSUE #886: Allow users to specify the number of events per page
Currently the events page shows 20 events per page. It would be nice if users could specify the page number instead of having a hardcoded value.

WojcikMike:
fixes #886 

image


sergioc (commented):
The dropdown is too crammed over the pager. Can it be aligned with the left side of the list?

WilliamBZA (approved):

WojcikMike:
@sergioc how about this? 
image

sergioc (approved):

PR #940, opened 2020-10-01
Endpoint filtering for monitoring view

timbussmann:
Adds a text input field that allows filtering endpoints by their name. In addition the filter can also be provided via search param in the URL so that filters can be shared and saved via bookmarks.

This is a draft PR, missing proper styling and more.

### DIFF
--- a/src/ServicePulse.Host/app/css/particular.css
+++ b/src/ServicePulse.Host/app/css/particular.css
@@ -3023,6 +3023,17 @@ a.remove-endpoint:hover i {
     border-radius: 4px;
 }
 
+.filter-input {
+    width: 280px;
+    vertical-align: top;
+    margin-top: 2px;
+    margin-left: 10px;
+    padding-right: 10px;
+    padding-left: 10px;
+    border: 1px solid #aaa;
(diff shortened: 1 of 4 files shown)


timbussmann:
this seems to work fine in combination with the SC-connection embedded in the URL query string.

timbussmann:
I moved the filter input to the same location as on the heartbeats view. I couldn't reuse the existing classes out of the box as they contains some specific magin settings and other stuff related to the position on the heartbeats view.

image

I can finish the styling next week, however I'm not 100% convinced about the location. Since this filters by endpoint name, does it make sense to have the filter input on the exact opposite side of the screen? This feels a bit counter-intuitve, thoughts @sergioc ?

timbussmann:
tried to tweak the styling. This has been a CSS nightmare due to get the correct "right-alignment" of the time scale and the filter input due to to all the styles applied to the nav elements (used for the timerange selectors).

image


@sergioc please review and feel free to fix the CSS ;)

seanfarmar (approved):
LGTM

sergioc (approved):

PR #942, opened 2020-10-06
Add ordering functionality to monitoring view

timbussmann:
Adds the ability to order by endpoint name or the metrics. The button is a toggle button and switches between asc/desc when you click it. 

The ordering works by the provided average value of the given metric. This means that not the endpoint with the highest spike would show up at the top but the endpoint with the highest average value. There is currently no other reliably comparable value available to sort by, we'd have to add that to SC I think. But average might be a fairly good value to sort by because the average changes much more slowly than the most current data point which makes the sorting quite a bit more stable and not jumping forth and back constantly on every bump in the graph.

The UI needs some tweaking, I just added the buttons as a first spike. It definitely causes issues with the length of the retry column header, I think we might have to shorten the column value if we keep the approach of adding the sorting on the column level.

image

(PR currently depends on #940)


seanfarmar (approved):
Loads of CCS stuff I can't review but looks good to me :-)

timbussmann:
I think I got the main functionality working but I think this needs a lot of CSS love and stuff. Main issues I saw:
* When the sort icon is added, the margin/padding is weird and moves the text incorrectly
* the tooltip text sometimes hardly shows up or flickers
* the click event isn't fired on every position, e.g. it seems when clicking directly on the text it doesn't account for a click

@sergioc let me know if you can help me out with this

src/ServicePulse.Host/app/modules/monitoring/views/endpoint_details.html
```diff
@@ -111,8 +111,8 @@ <h1 class="righ-side-ellipsis col-lg-max-10" uib-tooltip="{{endpointName}}">
                     </div>
                     <div class="col-xs-6 no-side-padding scheduled-retries-rate-values">
                         <div class="row">
-                            <span class="metric-digest-header" uib-tooltip="Scheduled retry rate: The number of messages per second scheduled for retries (immediate or delayed).">
-                                Scheduled Retries Rate
+                            <span class="metric-digest-header" uib-tooltip="Retries rate: The number of messages per second scheduled for retries (immediate or delayed).">
```
timbussmann, line 114: maybe an alternative would be "Scheduled retries:", that's also short but implies delayed retries (scheduled). There is a certain consistency using rate given the other columns though.

timbussmann:
this got a few more changes and tweaks but should be good now, do you want to review again @seanfarmar ?

timbussmann:
@sergioc I noticed some layouting issues on the monitoring endpoint details page, not sure this is related to the changes made in this PR though?

image

timbussmann:
discussed the styling error with @sergioc and it seems this is likely to come from the changes made in #940. This can be addressed outside of this PR, merging this for now.

PR #945, opened 2020-10-09
Update monitoring and heartbeat icons

sergioc:
Heartbeat and monitoring icons are now easier to distinguish one from the other. Icons for endpoint connectivity loss and monitoring connectivity loss are also now better represented.

image


timbussmann (approved):
looks fine to me. On the last row on the screenshot, the alignment of the two icons seems a bit off but not sure that needs to be addressed.

sergioc:
fixed alignment

seanfarmar (approved):
LGTM

PR #946, opened 2020-10-09
Add unit tests for passing filter value via url query

timbussmann:
adding some unit tests to the filter query feature that was added in #940


seanfarmar (approved):
LGTM

Issue #639, opened 2018-06-06
Add a Select All button to Service Pulse


outofcoolnames:
Describe your problem / suggestion / idea / feedback here.

When browsing https://<psp>/#/failed-messages/groups/ it would be useful to have a select all feature. Often, when opening message failures for a specific endpoint, there may be a large number of messages where you may want to archive/retry 250 out of 260 failures. By adding a select all feature, it would all unticking a handful of message failures. Alternatively, could you point me to some API documentation allowing retries by endpoint and failure type.
***************

*Technical information if reporting a bug:*

- **ServicePulse version:** 
- **ServiceControl version:** 
- **Browser version:**

WojcikMike:
Thanks @outofcoolnames for raising it. Can you describe in more detail the business case when you would like to use such a feature? The problem that I see with proposed implementation is that we are using infinite scroll so for the most part 260 items (in your example) will not be displayed on a page but probably we could find a way around it. The bigger problem is that a user needs to find 10 messages out of 260 to exclude them which is not a simple task, and is error-prone. 

Maybe it is a classification problem (user wants to retry/archive all errors send from a given endpoint, or of a given type etc.). I would like to understand better when you are in need to do such retry/archive.

outofcoolnames:
@WojcikMike , I get around the infinite scoll by scrolling to the bottom until all items are loaded. Lets say I'm in the list where I've selected "Failed message groups" and grouped by "Endpoint Address". I may have a mixture of failures in the list. An even more useful solution in this instance would be a search filter where I could type "failure message" and the list would be filtered, I could then archive/ retry these. 
You may say that this functionality is available when I've selected "Failed message groups" and grouped by "Exception Type and Stack Trace". However, when opening a failure in this list, there may be multiple endpoints returned. I may not want to retry / archive all of these. A search filter would be useful here too to filter by endpoint

WojcikMike:
If I understand correctly the actual need is to make operations on a group of certain message types (or exceptions types) in a context on 1 endpoint. Which at the moment you can't do. Let me think about it as I am a bit reluctant implementing select all button to solve this problem. Easily you might want to retry/archive 30% of failed messages in which case select all doesn't solve the problem.

sergioc:
@outofcoolnames we've just [released ServicePulse 1.27](https://github.com/Particular/ServicePulse/releases/tag/1.27.0) and added a button to select all messages. Note that this only applies to messages that were already loaded in the UI.

[...]

PR #947, opened 2020-10-09
Select all messages and toggle if messages are selected

seanfarmar:
related to #639


src/ServicePulse.Host/app/js/views/failed_messages/controller.js
```diff
@@ -109,6 +109,22 @@
...
+            if(vm.selectedIds.length == vm.failedMessages.length || vm.selectedIds.length > 0) {
+                selectAll = false;
+            }
+            vm.failedMessages.forEach(function(item) {
+                if (selectAll == false) {
+                    vm.selectedIds = vm.selectedIds.filter(i => i === item.id );
+                    item.selected = false;
+                } else {
+                    item.selected = true;
+                    vm.selectedIds.indexOf(item.id) >= 0 ? false : vm.selectedIds.push(item.id);
```
timbussmann, line 123: if we clear the `selectedIds` collection before looping through `failedMessages`, shouldn't this ensure that `vm.selectedIds.indexOf(item.id) >= 0` is never true and we could just push without checking?

timbussmann (commented):
some ideas to consider

sergioc (changes requested):
Ideally I'd prefer a checkbox instead of a button to stick to conventions, but for now let's roll with the button for the sake of simplicity. 

Only the label `Select None` should be renamed to `Clear selection`.

sergioc (approved):

timbussmann (approved):

PR #948, opened 2020-10-13
Monitoring endpoint grouping

timbussmann:


sergioc (approved):

JSCProjects:
Great functionality! Would also be nice if we also can group on "-". For viewing purposes tabs could also be an option!

timbussmann:
@JSCProjects are you using both `.` and `-` in combination or just one or the other?

JSCProjects:
We are using both, because we started using "-" and after a getting a Cogin Queue Explorer license we saw it was capable to group it when using "." then we started to use "." and "-" is legacy, but the grouping convention of naming endpoints is the same.

timbussmann:
thanks for the detail. Supporting both in combination might be a bit tricky with the current implementation as it also does "reconstruct" the name be combining the splitted segments again, and in this case it would need to know a lot more to do that properly. Doing `.` or `,` alone would be easier to support. But I can forward this as a feature request for a further improvement 👍

PR #972, opened 2020-10-22
don't drop task in StaticFileContext.SendAsync

adamralph:


src/ServicePulse.Host/Owin/Microsoft/StaticFileContext.cs
```diff
@@ -218,8 +218,7 @@ public Task SendAsync()
             var readStream = fileInfo.CreateReadStream();
             var copyOperation = new StreamCopyOperation(readStream, response.Body, length, request.CallCancelled);
             var task = copyOperation.Start();
-            _ = task.ContinueWith(resultTask => readStream.Close(), TaskContinuationOptions.ExecuteSynchronously);
-            return task;
+            return task.ContinueWith(resultTask => readStream.Close(), TaskContinuationOptions.ExecuteSynchronously);
```
danielmarbach, line 221: I think this is an unnecessary optimization that complicates things for now added value. I would just switch this to async-await style. That is more readable than continuations and is well aligned with https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-await-over-continuewith

This code is anyway not exactly on the hot path as far as I'm concerned

adamralph: 👍 I'll amend the commit to do that.

src/ServicePulse.Host/GlobalSuppressions.cs
```diff
@@ -6,3 +6,4 @@
 using System.Diagnostics.CodeAnalysis;
 
 [assembly: SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1600:Elements should be documented", Justification = "Not a public API")]
+[assembly: SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task", Justification = "console app")]
```
danielmarbach, line 9: Even if it is console app we never suppress configureawait in any production code and I see no reason to deveat from that. Even more so because middleware pieces like this tend to get borrowed into other code pieces

adamralph: Fair enough, I'll fix.

adamralph:
@danielmarbach I amended the first commit.

danielmarbach (approved):

PR #976, opened 2020-10-28
Provide shortened type names on message hierarchies

### LINKED ISSUE #970: Unknown Message Type listed in Monitoring when multiple message types
Similar to #604, I am seeing Unknown listed as the message type after upgrading to 1.27.0. However it looks like it shows unknown only if the message type has multiple message types listed via inheritance (the hover over icon shows this correctly)

image

***************

- **ServicePulse version:1.27.0 
- **ServiceControl version:4.7.1
- **Browser version:Firefox latest

timbussmann:
fixes #970 

"unknown" is shown as messages with a hierarchy don't have a `shortName` property at the moment. Instead of checking for the shortName to exist and fallback to the typeName, it seems more appropriate to also shorten type names on hierarchies too. Replaced the more complex fold with a map/join combination.

### DIFF
--- a/src/ServicePulse.Host.Tests/tests/js/services/services.messageTypeParser.spec.js
+++ b/src/ServicePulse.Host.Tests/tests/js/services/services.messageTypeParser.spec.js
@@ -70,6 +70,16 @@ describe('messageTypeParser', function () {
         expect(sut.assemblyName).toEqual('Shared');
     });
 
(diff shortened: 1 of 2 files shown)


seanfarmar (approved):
LGTM

WilliamBZA (approved):
Added a unit test for it

PR #984, opened 2020-11-17
Using BodyURl to get body for a single message. Due to this change SP…

WojcikMike:
… will work with SC v5 and with previous versions


WilliamBZA (changes requested):
It's really rich of me to ask for this, but any chance of a test?

WilliamBZA (approved):

PR #985, opened 2020-11-17
replace absolute paths with relative

rlaveycal:
There are absolute path references in links that stop the UI from working if hosted in IIS at a non-root location.


WojcikMike (approved):

WojcikMike:
Thanks @rlaveycal for fixing this for us. I will take a look at some automatic way for checking it. As this bug sneaks in more often than it should.

PR #1031, opened 2021-04-12
Healthcheck notifications

tmasternak:


SzymonPobiega:
I can't really read or review JS. I can try running it and clicking here and there. Would that be ok @tmasternak ?

SzymonPobiega:
Looks good. The only thing that looks a bit odd is this:

image

but I am not sure we can do anything with it. Certainly not with my CSS/HTML skills 😂

SzymonPobiega (approved):

src/ServicePulse.Host/app/modules/configuration/js/notifications/emailnotifications.controller.js
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        $scope.save = (notificationsForm) => {
+            if (notificationsForm.$valid) {
+                notificationsService.updateSettings($scope.settings).then(
+                    () => {
+                        toastService.showInfo('Eamil settings updated.');
+                        $uibModalInstance.dismiss('saved');
+                    },
+                    () => toastService.showError('Failed not update email settings.'));
```
danielmarbach, line 23: Is this yoda style ;)?

```suggestion
                    () => toastService.showError('Failed to update the email settings.'));
```

src/ServicePulse.Host/app/modules/configuration/views/notificationsemailmodal.html
```diff
@@ -0,0 +1,56 @@
...
+            </div>
+            <div class="row"></div>
+            <div class="form-group">
+                <label for="account">Authentication password</label>
+                <input type="password" id="password" name="password"
+                    ng-model="settings.authentication_password" class="form-control" />
+            </div>
+            <div class="row"></div>
+            <div class="form-group">
+                <input type="checkbox" id="enableSSL" name="enableSSL"
```
danielmarbach, line 34: As far as I recall TLS is superseeded SSL and email also has STARTTLS. Should this be renamed to something like EnableEncryption?

In outlook this is called Encryption method

image

src/ServicePulse.Host/app/modules/configuration/views/notificationsemailmodal.html
```diff
@@ -0,0 +1,56 @@
...
+            <div class="row"></div>
+            <div class="form-group">
+                <label for="from">From address</label>
+                <input type="from" id="from" name="from"
+                    ng-model="settings.from" class="form-control" />
+            </div>
+            <div class="row"></div>
+            <div class="form-group">
+                <label for="to">To address</label>
+                <input type="to" id="to" name="to"
```
danielmarbach, line 47: Is this validated or is this one of the https://mentalized.net/journal/2019/08/02/email-validation-is-hard/?

tmasternak: I've added browser-based validation to both email fields and the port fields

src/ServicePulse.Host/app/modules/configuration/views/notificationsemailmodal.html
tmasternak: I've changed the label (and all types involved) to `EnableTLS`

mikeminutillo (approved):
I'm not terribly confident reviewing js but it looks good to me.

It might make sense to have the test connection button on the edit. That requires a bit of a re-architectue of the back-end, and it's probably not worth it for this first release.

Does ServicePulse have a minim required ServiceControl version, because this will increase it. What happens in this UI with an older version of ServiceControl?

tmasternak:
> It might make sense to have the test connection button on the edit. That requires a bit of a re-architectue of the back-end, and it's probably not worth it for this first release.

I had a chat with Sergio about that and we decided that the current location is good enough

> Does ServicePulse have a minim required ServiceControl version because this will increase it. What happens in this UI with an older version of ServiceControl?

It looks that [this](https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host/app/js/services/factory.shareddata.js#L36) is the current minimum which looks quite outdated. I tried to spike an approach in which the tab hides when the ServiceControl version is less than 4.17 (we already have code to do version parsing etc.). That being said there seems to be more re-work needed around Angular controllers for the configuration section. I propose to create a separate issue for that.

PR #1049, opened 2021-05-11
Add warnings for negative critical time

sergioc:


aleksandr-samila:
Some plumbing was added

aleksandr-samila:
Updated the sources to highlight few more negative numbers. Should be ready for the final review

src/ServicePulse.Host/app/js/app.constants.js
```diff
@@ -1,7 +1,7 @@
 window.defaultConfig = {
     default_route: '/dashboard',
     version: '1.2.0',
-    service_control_url: 'http://localhost:33333/api/',
-    monitoring_urls: ['http://localhost:33633/'],
+    service_control_url: 'http://localhost:49200/api/',
```
WojcikMike, line 4: This change should be reverted

WojcikMike (approved):
Just reverse the change to the app.contants.js. Then you can merge it

PR #1053, opened 2021-05-11
Added general title for retried messages without a group.

aleksandr-samila:
Closes https://github.com/Particular/ServicePulse/issues/744

### DIFF
--- a/src/ServicePulse.Host/app/js/views/failed_groups/failed-groups-view.html
+++ b/src/ServicePulse.Host/app/js/views/failed_groups/failed-groups-view.html
@@ -29,8 +29,7 @@
                                 <div class="col-sm-12 no-side-padding">
                                     <div class="row box-header">
                                         <div class="col-sm-12 no-side-padding">
-                                            <p class="lead break" ng-if="group.originator != null"> {{group.originator}}</p>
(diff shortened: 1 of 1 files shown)


sergioc (approved):

src/ServicePulse.Host/app/js/views/failed_groups/failed-groups-view.html
```diff
@@ -29,7 +29,8 @@ <h6>
                                 <div class="col-sm-12 no-side-padding">
                                     <div class="row box-header">
                                         <div class="col-sm-12 no-side-padding">
-                                            <p class="lead break"> {{group.originator}}</p>
```
WojcikMike, line 32: What is written below is correct. But a shorter way of doing that would be:
<p class="lead break"> {{group.originator || 'Selection of individual message(s)'}}</p>

WojcikMike:
@aleksandr-samila above I posted another way of implementing the same thing (a bit shorter). It is up to you, which one you choose so feel free to merge it

aleksandr-samila:
@WojcikMike I applied your suggestion because I found it more elegant. Thanks.

PR #1055, opened 2021-05-12
Ability to add notes to failed message groups

sergioc:


sergioc:
@aleksandr-samila @WojcikMike I've added some markup and styling for failed message group notes. When no note is present, only the button "Add note shows". Otherwise everything else shows but not the "Add note" button.

sergioc:
Adding/editing notes should happen in a textarea inside a "Add note" or "Edit note" modal.

WojcikMike (approved):

Issue #271, opened 2016-02-11
Age of items should show GMT + Browser time when hovering


ramonsmits:
Age of items should show GMT + Browser time in the tooltip (title attribute) when hovering the mouse.

![image](https://cloud.githubusercontent.com/assets/152998/12974743/9638b90e-d0b6-11e5-9b76-945316fcb2db.png)

GMT should be shown as that is the timestamps used in log files.

Browser/Local time should be shown for the user to set the event in its own context of time.

PR #1056, opened 2021-05-13
Added gmt time

aleksandr-samila:
Closes https://github.com/Particular/ServicePulse/issues/271

### DIFF
--- a/src/ServicePulse.Host/app/js/directives/moment.js
+++ b/src/ServicePulse.Host/app/js/directives/moment.js
@@ -26,7 +26,7 @@
                 attrs.$observe('date', function(value) {
                     if (value && attrs.date !== minDate) {
                         m = moment(attrs.date);
-                        element.attr('title', m.format('LLLL') + " (local)\n" + m.utc().format('LLLL') + " (GMT)");
+                        element.attr('title', m.format('LLLL') + " (local)\n" + m.utc().format('LLLL') + " (UTC)");
                         updateText();
                         updateLoop();
                     } else {
--- a/src/ServicePulse.Host/app/js/directives/moment.js
+++ b/src/ServicePulse.Host/app/js/directives/moment.js
@@ -26,7 +26,7 @@
                 attrs.$observe('date', function(value) {
                     if (value && attrs.date !== minDate) {
                         m = moment(attrs.date);
-                        element.attr('title', m.format('LLLL'));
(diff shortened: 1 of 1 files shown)


aleksandr-samila:
image

sergioc (approved):

src/ServicePulse.Host/app/js/directives/moment.js
```diff
@@ -26,7 +26,7 @@
                 attrs.$observe('date', function(value) {
                     if (value && attrs.date !== minDate) {
                         m = moment(attrs.date);
-                        element.attr('title', m.format('LLLL'));
+                        element.attr('title', m.format('LLLL') + " (local)\n" + m.utc().format('LLLL') + " (GMT)");
```
WojcikMike, line 29: Can we change it to UTC? Overall that expresses the same intent. But UTC express time zones while GMT is a time zone with UTC+0

https://www.timeanddate.com/time/gmt-utc-time.html

aleksandr-samila: you mean just to change a label? yep, why not

sergioc: Check the original issue for the rationale to use GMT (to match timestamps in logs). Not sure if UTC is a good idea.

aleksandr-samila: Yesterday I found this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toGMTString - it's not recommended. And in any case it's just a label, because there is `utc()` function that is used and there is no `gmt()` one.

And physically - the difference between these two systems is a few seconds - nothing to discuss. Just need to make a decision.

WojcikMike (approved):

src/ServicePulse.Host/app/js/directives/moment.js
WojcikMike: @sergioc the original issue is incorrect. Meaning, the logs states:

2020-10-29 07:20:57.8920|228|Info|ServiceControl.Audit.Infrastructure ...

So the printed output does not display what timezone it uses. As such when expressing a universal time it should express in UTC - which is a standard and not GMT - which is the name of the timezone UTC+0. It is a common mistake but using GMT is not accurate.

ramonsmits: Agreed, UTC is the better name and was also my intend.

PR #1087, opened 2021-06-24
Signing with KeyVault

SeanFeldman:
This PR converts the signing of the installer and NuGet packages to use a certificate stored in Key Vault, which enables us to use an EV code signing certifcate.

### DIFF
--- a/src/Setup/ServicePulse.aip
+++ b/src/Setup/ServicePulse.aip
@@ -1,14 +1,12 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<DOCUMENT Type="Advanced Installer" CreateVersion="9.9" version="13.5" Modules="enterprise" RootPath="." Language="en" Id="{EB6913C3-A2D9-484A-9A49-E3D640F2A96C}">
-  <COMPONENT cid="caphyon.advinst.msicomp.ProjectOptionsComponent">
-    <ROW Name="HiddenItems" Value="AppXAppDetailsComponent;AppXCapabilitiesComponent;AppXDependenciesComponent;AppXProductDetailsComponent;AppXVisualAssetsComponent;AppXAppDeclarationsComponent;AppXUriRulesComponent"/>
-  </COMPONENT>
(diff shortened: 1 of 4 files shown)


bording (approved):

src/Setup/ServicePulse.aip
```diff
@@ -1,14 +1,12 @@
...
-    <ROW Name="HiddenItems" Value="AppXAppDetailsComponent;AppXCapabilitiesComponent;AppXDependenciesComponent;AppXProductDetailsComponent;AppXVisualAssetsComponent;AppXAppDeclarationsComponent;AppXUriRulesComponent"/>
-  </COMPONENT>
+<DOCUMENT Type="Advanced Installer" CreateVersion="9.9" version="18.3" Modules="enterprise" RootPath="." Language="en" Id="{EB6913C3-A2D9-484A-9A49-E3D640F2A96C}">
   <COMPONENT cid="caphyon.advinst.msicomp.MsiValidationComponent">
     <ROW Table="File" Column="Sequence" MinValue="1" MaxValue="2147483647" Description="Sequence with respect to the media images; order must track cabinet order." Options="0" ColumnType="3" ColumnIndex="7" ColumnSize="4" MsiKey="File#Sequence"/>
     <ROW Table="Media" Column="LastSequence" MinValue="0" MaxValue="2147483647" Description="File sequence number for the last file for this media." Options="0" ColumnType="3" ColumnIndex="1" ColumnSize="4" MsiKey="Media#LastSequence"/>
   </COMPONENT>
   <COMPONENT cid="caphyon.advinst.msicomp.MsiPropsComponent">
     <ROW Property="AI_BITMAP_DISPLAY_MODE" Value="0"/>
+    <ROW Property="AI_CURRENT_YEAR" Value="2021" ValueLocId="-"/>
```
DavidBoike, line 9: Since we're already doing some text replacement in this file as part of the release, should the value here be `replace-current-year` so it will never go out of date? I assume this is used for a copyright notice or something.

DavidBoike (approved):

src/Setup/ServicePulse.aip
bording: That's out of scope for these changes, so I'd rather not get into anything like that.

PR #1185, opened 2022-04-26
Strip comments from message body payloads

### LINKED ISSUE #1184: JSON bodies with comments are not displayed
While JSON does not officially support comments in the style of `// comment`, many JSON serializers allow the use of comments without breaking.

This means that if a JSON message with a comment in the payload is sent to the error queue, the body will be stored with the comment. ServicePulse will fail to render the body though as comments are not supported by the browser's deserializer.

WilliamBZA:
Fixes #1184

Removes comments of the style:

`// comment`
`/* comment */`

from the body of messages that are received from ServiceControl so that they can be rendered correctly.

This will also remove comments from XML messages as they are sent to SP as JSON anyway and will be subject to the same problems.

I don't know that this is testable given the reliance on angularjs as the `$http` wrapper.

### DIFF
--- a/src/ServicePulse.Host/app/js/services/services.service-control.js
+++ b/src/ServicePulse.Host/app/js/services/services.service-control.js
@@ -146,7 +146,7 @@
                 method: 'GET',
                 transformResponse: function(defaults, transform) {
                     // Remove any comments from the body before deserializing
-                    return defaults.replace(/\\"|"(?:\\"|[^"])*"|(\/\/.*|\/\*[\s\S]*?\*\/)/g, (m, g) => g ? "" : m);
+                    return defaults.replace(/\\"|"(?:\\"|[^"])*"|(\/\/.*|\/\*[\s\S]*?\*\/)/g, function (m, g) { return g ? "" : m });
                 }
               });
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/app/js/services/services.service-control.js
```diff
@@ -141,11 +141,14 @@
...
-                return {
-                    data: response.data
-                };
-            });
+            return $http({
+                url: url,
+                method: 'GET',
+                transformResponse: function(defaults, transform) {
+                    // Remove any comments from the body before deserializing
+                    return defaults.replace(/\\"|"(?:\\"|[^"])*"|(\/\/.*|\/\*[\s\S]*?\*\/)/g, function (m, g) { return g ? "" : m });
```
andreasohlund, line 149: OH MY GOD

andreasohlund (approved):

src/ServicePulse.Host/app/js/services/services.service-control.js
WilliamBZA: here

PR #1202, opened 2022-06-07
Make sure the built URI contains the endpoint name

mauroservienti:
This fixes a bug in the ServiceInsight command line search functionalities. If The command line doesn't contain an endpoint name, SI selects the ServiceControl node that causes the search to fail and instead all messages are displayed.

### DIFF
--- a/src/ServicePulse.Host/app/js/views/message/controller.js
+++ b/src/ServicePulse.Host/app/js/views/message/controller.js
@@ -183,6 +183,7 @@
 
         vm.debugInServiceInsight = function () {
             var messageId = vm.message.message_id;
+            var endpointName = vm.message.receiving_endpoint.name;
             var dnsName = connectionsManager.getServiceControlUrl().toLowerCase();
 
             if (dnsName.indexOf("https") === 0) {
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/app/js/views/message/controller.js
```diff
@@ -191,7 +192,7 @@
                 dnsName = dnsName.replace("http://", "");
             }
 
-            $window.open("si://" + dnsName + "?search=" + messageId);
+            $window.open("si://" + dnsName + "?search=" + messageId + "&endpointname=" + endpointName);
```
ramonsmits, line 195: Don't these values need escaping or is that done automatically by `$window.open`?

ramonsmits (approved):
Approving under the assumption that the comment regarding value escaping will be verified.

src/ServicePulse.Host/app/js/views/message/controller.js
mauroservienti: it's done automatically IIRC, also I don't think there can be many strange chars in an endpoint name.

mauroservienti:
I converted this to a draft because it needs more validation:

- validate if `$window.open` escapes content automatically
- if it does ServiceInsight needs to unescape the passed values to get back the original endpoint name

mauroservienti:
We did some research and it turns out that `$window.open` escapes the string. ServiceInsight decodes the string back upon receiving it https://github.com/Particular/ServiceInsight/blob/4ab2148d0cc0e2c59cb0e03942e19e1f9b90eff2/src/ServiceInsight/Models/CommandLineOptions.cs#L94

PR #1224, opened 2022-07-07
Adding average labels to graphs in monitoring module

cquirosj:
In the monitoring module, when hovering a given graph, show the corresponding average value on a label outside the sparkline.


src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.largeGraph.js
```diff
@@ -163,23 +168,103 @@
...
+                            var firstAverageLine = drawAverage(firstSeries, attrs.firstSeriesColor, attrs.firstSeriesFillColor );
+
+                            var secondAverageLine = null;
 
                             if (secondSeries) {
-                                drawAverage(secondSeries, attrs.secondSeriesColor, attrs.secondSeriesFillColor);
+                                secondAverageLine = drawAverage(secondSeries, attrs.secondSeriesColor, attrs.secondSeriesFillColor);
                             }
+
+                            chart.on("mouseover", function() {
```
cquirosj, line 200: @WojcikMike question: Do you think the closure functions I used here for the mouseover and mouseout events could create a memory leak? I am asking because the code of this ui.particular.largeGraph.js class redraws the whole SVG constantly by destroying all the elements in the SVG each time. 

scope.$watch('firstDataSeries', function is going to create a new chart variable every time it runs so in theory the chart instances from previous runs 


In the past to prevent possible memory leaks I created a function signature in the code sacrificing the readability but ensuring no new closure instances are created and kept referenced in the stack.

While checking the number of handlers register in the event using DevTools I see the number of registered closures remains constant, although I don't know if DevTools will refresh this in real-time:

image

cquirosj:
@sergioc this first version is working for the large graph. I would like to pair for implementing the same on the smaller graphs. I also have some concerns about the browser features I used that I would like to review with you.

src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.largeGraph.js
WojcikMike: The question sounds reasonable. And I would say that it should be causing memory leaks. I googled for a bit and found this (https://stackoverflow.com/questions/25059979/memory-release-of-d3-js-listener-inside-an-angularjs-directive). Som response suggest that modern browsers remove listener when an element is removed from DOM. Maybe that is why dev tools, do not show it.

cquirosj:
@sergioc could you test the average label in /monitoring/views/monitored_endpoints_list_grouped.html? I couldn't switch to the grouped view.

src/ServicePulse.Host/app/modules/monitoring/js/directives/ui.particular.largeGraph.js
cquirosj: I ran version 1.31.3, and the code from this branch and the memory usage behavior remains identical.

sergioc (approved):

WojcikMike (approved):

PR #1252, opened 2022-07-26
Show vanilla JS spike on events page

WilliamBZA:


src/ServicePulse.Host/app/js/dashboard/spMoment.ts
```diff
@@ -0,0 +1,27 @@
+import moment from 'moment';
+
+export default class EventLogItemComponent extends HTMLElement {
+    date: moment.Moment | null = null;
+
+    constructor() {
+        super();
+
+        // Wait for slots to be populated
+        let observer = new MutationObserver((changes, observer) => {
```
WilliamBZA, line 10: This web component shows how sub-web components would have to be loaded and displayed. This observer mechanism is again only required because of a lack of shadow dom.

src/ServicePulse.Host/app/js/dashboard/spMoment.ts
```diff
@@ -0,0 +1,27 @@
+import moment from 'moment';
+
+export default class EventLogItemComponent extends HTMLElement {
```
WilliamBZA, line 3: Currently done via an angularjs directive https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host/app/js/directives/moment.js

src/ServicePulse.Host/app/js/dashboard/spMoment.ts
```diff
@@ -0,0 +1,27 @@
...
+            this.innerText = this.date.fromNow();
+
+            this.startUpdateLoop();
+        });
+
+        observer.observe(this, { childList: true});
+    }
+
+    startUpdateLoop() {
+        setInterval(() => {
```
WilliamBZA, line 23: Don't need to cancel if this is an MPA as a full reload on navigate will clear this anyway

src/ServicePulse.Host/app/js/dashboard/eventLogItems.ts
```diff
@@ -0,0 +1,13 @@
+import EventLogItem from "./eventLogItem";
+
+const scu = 'http://localhost:33333/api/';
+
+export default class EventLogItems {
+    getEventLogItems() : Promise<EventLogItem[]> {
+        var url = scu + 'eventlogitems';
+
+        return fetch(url).then<EventLogItem[]>(response => {
```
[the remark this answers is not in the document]
kbaley: Would you recommend using vanilla fetch or something like axios?

kbaley (commented):
This looks like a feasible option and has the advantage of not requiring specialized knowledge of a framework. Should we be concerned that the code would evolve into our own internal framework that requires specialized knowledge to use?

src/ServicePulse.Host/app/js/dashboard/eventLogItems.ts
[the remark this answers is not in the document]
WilliamBZA: I'm not terribly opposed. I haven't used axios, but have used fetch plentifully. So I'm not sure that axios adds anything that we'd be missing for ServicePulse.

A quick search shows

- Easier cancellation (definitely relevant for a SPA, but maybe less so for a MPA)
- Progress bars for large transfers (not relevant for ServicePulse _yet_)
- Support for IE (Not needed for ServicePulse)
- One less step in json transformation (with fetch you have to call `response.json()` to get the json result)

None of these feel like killer features to me. I'm also a little wary since we've had at least [1](https://github.com/Particular/ServicePulse/issues/1184) bug recently that was awkward to fix because we were using non-standard ways of retrieving data. I could be convinced, but my default is to stick with standards until they no longer work.

WilliamBZA:
> Should we be concerned that the code would evolve into our own internal framework that requires specialized knowledge to use?

Yes. We should.

PR #1259, opened 2022-07-27
React/Angular JS hybrid

kbaley:
A spike to see how React could work in a hybrid mode with ServicePulse

All the important bits are in the app/React folder (except the public folder which contains stuff copied verbatim from the AngularJS app).

For development work, there's the `npm run dev-both` script which launches the AngularJS app in port 8080 and the React app in port 3000


src/ServicePulse.Host/app/react/src/Footer.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+function Footer() {
+  return (
+
+    <footer className="footer ng-scope">
```
kbaley, line 6: This is just copied from the AngularJS index page. I didn't put any smarts into it

src/ServicePulse.Host/app/react/src/Last10Events.js
```diff
@@ -0,0 +1,110 @@
+import { getAllByPlaceholderText } from '@testing-library/react';
+import React from 'react';
+
+function Last10Events() {
+  const scu = 'http://10.211.55.3:33333/api/';
```
kbaley, line 5: Should be localhost but I'm running ServiceControl in a VM locally

src/ServicePulse.Host/app/react/src/NavBar.js
```diff
@@ -0,0 +1,74 @@
+function NavBar() {
+  return (
```
kbaley, line 2: Nothing interesting here. Just copy/pasted HTML

src/ServicePulse.Host/app/react/src/SystemStatus.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+function SystemStatus() {
+  return (
```
kbaley, line 4: Copy-pasted

PR #1300, opened 2022-10-25
Moving angular related code into angular directory

WojcikMike:


WojcikMike:
Closing as we are changing approach

PR #1320, opened 2022-11-22
Refactor license

SzymonPobiega:
Try to combine the license warnings and inject the license-related props via `import`. Please take a look and let's discuss tomorrow.


jpalac (approved):
Looks good - I have done testing with various license.

- I got rid of some of the keys that we are no longer using.
- I've wrapped the license in a method and put it into the onMount method will all the other fetches. Also allows us to fetch it on a timer if we want to, but since angular is not doing that then I don't think there's any point until we have everything in vue.
- also applied toast styling changes to this branch (by accident) - it now doesn't have rounded corners and the close button and icon are at the top.

src/ServicePulse.Host/vue/src/App.vue
```diff
@@ -51,53 +47,33 @@ import {
...
-    isPlatformTrialExpired.value ||
-    isInvalidDueToUpgradeProtectionExpired.value
-  );
-});
+function getLicense() {
+  useLicense(serviceControlUrl.value).then((lic) => {
+    license.value = lic.value;
+  });
+}
+//setInterval(() => getLicense(), 60000);
```
tmasternak, line 72:
```suggestion
```

src/ServicePulse.Host/vue/src/assets/main.css
```diff
@@ -15,7 +36,7 @@
     /* background-image: url('../../../angular/app/img/toast-danger.svg') !important;
     background-position-y: 30px;
     background-size: 28px 28px; */
```
tmasternak, line 38:
```suggestion
```

tmasternak (approved):

WojcikMike (approved):

PR #1329, opened 2022-11-25
Fix release build

SzymonPobiega:
Attempt to port the changes in the CI build config to the Release build config


jpalac (approved):
Is there a reason we can't update this to also run on `windows-2022` - the same way the CI version has been updated?

tmasternak (approved):

PR #1350, opened 2022-12-14
Update set output

soujay:
update set output

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           dotnet-version: 5.0.x
       - name: Add msbuild to PATH
-        uses: microsoft/setup-msbuild@v1.0.3
+        uses: microsoft/setup-msbuild@v1.3
       - name: AngularJs - Install Host npm packages
         run: npm run load
         working-directory: src/ServicePulse.Host/angular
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         with:
           dotnet-version: 5.0.x
       - name: Add msbuild to PATH
-        uses: microsoft/setup-msbuild@v1.0.2
+        uses: microsoft/setup-msbuild@v1.0.3
```
DavidBoike, line 25: I don't know why they don't use 3-number format but 1.0.3. still uses old node.
```suggestion
        uses: microsoft/setup-msbuild@v1.3
```

.github/workflows/release.yml
```diff
@@ -16,20 +16,20 @@ jobs:
...
         with:
-          fetch-depth: 0    
+          fetch-depth: 0
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v3.0.3
         with:
           dotnet-version: 5.0.x
       - name: Add msbuild to PATH
-        uses: microsoft/setup-msbuild@v1.0.2
+        uses: microsoft/setup-msbuild@v1.0.3
```
DavidBoike, line 27: Here too.
```suggestion
        uses: microsoft/setup-msbuild@v1.3
```

DavidBoike (changes requested):

DavidBoike (approved):

PR #1382, opened 2023-02-18
Vue - Failed message details (StackTrace)

TravisNickels:


src/ServicePulse.Host/vue/src/views/MessageView.vue
```diff
@@ -0,0 +1,119 @@
...
+    .then((response) => {
+      if (response.status === 404) {
+        failedMessage.value = { notFound: true };
+      } else {
+        failedMessage.value = { error: true };
+      }
+      return response.json();
+    })
+    .then((data) => {
+      var message = data;
```
WilliamBZA, line 23: What are these values if the response is a 404? Will this fail?

src/ServicePulse.Host/vue/src/views/MessageView.vue
```diff
@@ -0,0 +1,119 @@
...
+      console.log(err);
+      return;
+    });
+}
+
+function togglePanel(panelNum) {
+  if (failedMessage.value.notFound || failedMessage.value.error) return false;
+
+  failedMessage.value.panel = panelNum;
+  return false;
```
WilliamBZA, line 39:
```suggestion
  if (!failedMessage.value.notFound && !failedMessage.value.error) {
    failedMessage.value.panel = panelNum;
  }
  return false;
```

2 returns feels weird given that they're both return false.

soujay (approved):

src/ServicePulse.Host/vue/src/views/MessageView.vue
TravisNickels: I accidentally made [this change](https://github.com/Particular/ServicePulse/pull/1384/commits/7f765854141863e510d6492c6e8d0f707da2c545#diff-eab870483f4a11e1169e4d2d0e5e24f4bee9d0220699a8550d7c8ee2438471c2R79-R81) in the other PR for the headers and the body that is branched off from this PR.

src/ServicePulse.Host/vue/src/views/MessageView.vue
TravisNickels: These values are null because if the response status is 404, it skips the `then` that assigns the data values.  However, I have not been able to generate an error status to test if that gets caught.  This does assume that if the status is anything other than a 404 it sets error to true and should be caught, unless the status is `200 ok` and then it should assign the data to the object and overwrite the failedMessage.value.  I might need to update this to be more specific about `200 ok` status.
image
image

WilliamBZA (approved):

PR #1393, opened 2023-02-22
Add Pending retries

WilliamBZA:
Adds the start for the pending retries screen

Make sure to set `showPendingRetry: true` in the `app.constants.js` file in the vue folder

image


src/ServicePulse.Host/vue/public/js/app.constants.js
```diff
@@ -3,5 +3,5 @@ window.defaultConfig = {
     version: '1.2.0',
     service_control_url: 'http://localhost:33333/api/',
     monitoring_urls: ['http://localhost:33633/'],
-    showPendingRetry: false
-};
+    showPendingRetry: false,
```
WilliamBZA, line 6: Change this to true to be able to see the changes in your dev environment

WilliamBZA, line 6: The build is failing because of this change
```suggestion
    showPendingRetry: true
```

src/ServicePulse.Host/vue/src/components/ConfirmDialog.vue
```diff
@@ -0,0 +1,66 @@
+<script setup>
```
soujay, line 1: I looked through vue and found a few controls that we could use.
https://vuejsexamples.com/tag/dialog/

soujay (approved):

PR #1397, opened 2023-02-23
Adds forward slash to the connection strings if they are missing

TravisNickels:
This is related to a support case that was raised about getting a 404 error on the connection and dashboard screens if a trailing `/` is not at the end of the connection strings in the configuration page.

Here is the issue [[2.0] ServicePulse unable to connect to ServiceControl if connection URL lacks a trailing forward slash](https://github.com/Particular/KeepTheLightsOn/issues/1188)

### DIFF
--- a/src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.js
+++ b/src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.js
@@ -92,7 +92,7 @@ export function updateServiceControlUrls(newServiceControlUrl, newMonitoringUrl)
   if (newServiceControlUrl.value.match(regex) === null){
     newServiceControlUrl.value += "/";
   }
-  if (newMonitoringUrl.value.match(regex) === null){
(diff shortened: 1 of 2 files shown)


src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.js
```diff
@@ -80,19 +80,25 @@ export function useServiceControlUrls() {
 }
 
 export function updateServiceControlUrls(newServiceControlUrl, newMonitoringUrl) {
-  if (!newServiceControlUrl) {
+  const regex = /\/$/gm;  // checks for a trailing forward slash
```
WilliamBZA, line 83: Do we really need regex for this? Can't a simple if check suffice?

I.e. `if (!newServiceControl.value.endsWith('/'))`

soujay (approved):

WilliamBZA (approved):

PR #1416, opened 2023-03-07
Vue - failed message edit and retry dialog

TravisNickels:


src/ServicePulse.Host/vue/src/components/failedmessages/EditRetryDialog.vue
```diff
@@ -0,0 +1,87 @@
...
+        <div class="modal-content">
+          <div class="modal-container">
+            <div class="modal-header">
+              <div class="modal-title">
+                <h3>Edit and retry message</h3>
+              </div>
+            </div>
+            <div class="modal-body">
+              <div class="row">
+                <div isolate-click class="col-sm-12 no-side-padding">
```
WilliamBZA, line 43:
```suggestion
                <div class="col-sm-12 no-side-padding">
```
`isolate-click` is an angular directive that does nothing here, it can be deleted.

src/ServicePulse.Host/vue/src/components/failedmessages/EditRetryDialog.vue
```diff
@@ -0,0 +1,87 @@
...
+                  <div class="nav tabs msg-tabs">
+                    <h5 :class="{ active: panelFoo === 1 }" class="nav-item" @click="togglePanel(1)"><a href="#">Headers</a></h5>
+                    <h5 :class="{ active: panelFoo === 2 }" class="nav-item" @click="togglePanel(2)"><a href="#">Message body</a></h5>
+                  </div>
+                  <!-- <pre isolate-click v-if="panel === 0">{{ failedMessage.exception?.message }}</pre>
+                                <pre isolate-click v-if="panel === 1">{{ failedMessage.exception.stack_trace }}</pre> -->
+                  <table isolate-click class="table" v-if="panel.number === 1">
```
WilliamBZA, line 50:
```suggestion
                  <table class="table" v-if="panel.number === 1">
```

`isolate-click` is an angular directive that does nothing here, it can be deleted.

src/ServicePulse.Host/vue/src/components/failedmessages/EditRetryDialog.vue
```diff
@@ -0,0 +1,87 @@
...
+                      </tr>
+                    </tbody>
+                  </table>
+                  <!-- <div isolate-click class="alert alert-info" v-if="panel === 2 && failedMessage.headersNotFound">Could not find message headers. This could be because the message URL is invalid or the corresponding message was processed and is no longer tracked by ServiceControl.</div> -->
+                  <!-- <pre isolate-click v-if="panel === 3 && !failedMessage.messageBodyNotFound">{{ failedMessage.messageBody }}</pre> -->
+                  <!-- <div isolate-click class="alert alert-info" v-if="panel === 3 && failedMessage.messageBodyNotFound">Could not find message body. This could be because the message URL is invalid or the corresponding message was processed and is no longer tracked by ServiceControl.</div> -->
+                  <!-- <flow-diagram v-if="failedMessage.conversationId" conversation-id="{{failedMessage.conversationId}}" v-show="panel === 4"></flow-diagram> -->
+                </div>
+              </div>
+              <pre isolate-click v-if="panel.number === 2">{{ settings.message.messageBody }}</pre>
```
WilliamBZA, line 66:
```suggestion
              <pre v-if="panel.number === 2">{{ settings.message.messageBody }}</pre>
```
`isolate-click` is an angular directive that does nothing here, it can be deleted.

src/ServicePulse.Host/vue/src/components/failedmessages/EditMessageHeader.vue
```diff
@@ -0,0 +1,70 @@
...
+const settings = defineProps({
+  header: Object,
+});
+
+let origHeaderValue = undefined;
+const header = ref(settings.header);
+
+const headerValue = computed(() => settings.header.value);
+watch(headerValue, async (newValue) => {
+  header.value.isChanged = newValue !== origHeaderValue ? true : false;
```
WilliamBZA, line 13:
```suggestion
  header.value.isChanged = newValue !== origHeaderValue;
```

Truthy will make sure this is true without the explicit true/false stuff

src/ServicePulse.Host/vue/src/components/failedmessages/EditMessageHeader.vue
```diff
@@ -0,0 +1,70 @@
...
+      &nbsp;
+      <i class="fa fa-exclamation-triangle" uib-tooltip="This is a sensitive message header that if changed can the system behavior. Proceed with caution."></i>
+    </span>
+    <span v-if="header.isChanged">
+      &nbsp;
+      <i class="fa fa-pencil" tooltip="Edited"></i>
+    </span>
+  </td>
+  <td>
+    <input :class="{ 'header-removed': header.isMarkedAsRemoved }" class="form-control" :disabled="header.isLocked" v-model="header.value" />
```
WilliamBZA, line 49:
```suggestion
    <input :class="{ 'header-removed': header.isMarkedAsRemoved }" class="form-control" :disabled="header.isLocked" :value="header.value" />
```

src/ServicePulse.Host/vue/src/components/failedmessages/EditRetryDialog.vue
```diff
@@ -0,0 +1,233 @@
...
+  showEditAndRetryConfirmation.value = true;
+}
+
+function confirmCancel() {
+  if (message.value.isBodyChanged) {
+    showCancelConfirmation.value = true;
+    return;
+  }
+
+  if (message.value.headers.find((header) => header.isChanged === true)) {
```
WilliamBZA, line 50:
```suggestion
  if (message.value.headers.some((header) => header.isChanged)) {
```

src/ServicePulse.Host/vue/src/components/failedmessages/EditRetryDialog.vue
```diff
@@ -0,0 +1,233 @@
...
+  message.value.isBodyChanged = false;
+}
+
+function findHeadersByKey(key) {
+  return message.value.headers.find((header) => header.key === key);
+}
+
+function getContentType() {
+  var header = findHeadersByKey("NServiceBus.ContentType");
+  return header ? header.value : null;
```
WilliamBZA, line 69: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```suggestion
  return header?.value;
```

src/ServicePulse.Host/vue/src/components/failedmessages/EditRetryDialog.vue
```diff
@@ -0,0 +1,233 @@
...
+}
+
+function isContentTypeSupported(contentType) {
+  return contentType === "application/json" || contentType === "text/xml";
+}
+
+function getMessageIntent() {
+  var intent = findHeadersByKey("NServiceBus.MessageIntent");
+  return intent ? intent.value : null;
```
WilliamBZA, line 78: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```suggestion
  return intent?.value;
```

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -62,25 +65,26 @@ function getConfiguration() {
     })
     .then((data) => {
       configuration.value = data;
-      return;
+      return getEditAndRetryConfig();
```
WilliamBZA, line 68: Can this be done in parallel to the first data load?

E.g. can we make both requests at the same time?


Something like:

```JavaScript
function getConfiguration() {
  return promise.all([useFetchFromServiceControl("configuration")
    .then((response) => response.json())
    .then((data) => {
      configuration.value = data;
    }), getEditAndRetryConfig()]);
```

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -302,6 +306,35 @@ function exportMessage() {
...
+  showEditRetryModal.value = false;
+  useShowToast("info", "Info", `Retrying the edited message ${id} ...`);
+  return startRefreshInterval();
+}
+
+function startRefreshInterval() {
+  refreshInterval = setInterval(() => {
+    loadFailedMessage();
+  }, 5000);
+}
```
WilliamBZA, line 330: Should we clear the interval if it exists just in case to prevent accidental memory leaks?

```suggestion
function startRefreshInterval() {
  stopRefreshInterval();

  refreshInterval = setInterval(() => {
    loadFailedMessage();
  }, 5000);
}
```

src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.js
```diff
@@ -44,3 +44,18 @@ export function useRetryMessages(ids) {
     return;
   });
 }
+
+export function useRetryEditedMessage(ids, editedMessage) {
+  var payload = { message_body: editedMessage.value.messageBody, message_headers: editedMessage.value.headers };
+  return usePostToServiceControl("edit/" + ids, payload)
+    .then(async (response) => {
+      if (response.ok === false) {
```
WilliamBZA, line 52:
```suggestion
      if (!response.ok) {
```

WilliamBZA (changes requested):

src/ServicePulse.Host/vue/src/components/failedmessages/EditMessageHeader.vue
TravisNickels: `v-model` is a two-way binding for form inputs.  By changing this to `:value` it breaks the reactivity of the input field.  I would have to add `:input` as well and call another function to get the reactivity back.  However, this has highlighted a couple of things for me:
 - I had to fetch the message data again from the ServiceControl API to reset the values on a cancel request, which might not be necessary if we only are using `:value` and `:input`. [Here is the code](https://github.com/Particular/ServicePulse/pull/1416/files#diff-045b53b9f0f278f3c9f435926328841394e39d567c341feed34754807d148295R316)
 -  Also, I watch the computed header value [here](https://github.com/Particular/ServicePulse/pull/1416/files#diff-cd3f7081646b1c0eb2421a2e3516af2c7a6eb92a42f0a73739fdb6899d5afb31R11-R13), but I think I read somewhere that getting a value from `computed()` is like a snapshot of the value and shouldn't be mutated.  However, we do state [here](https://github.com/Particular/Platform/blob/main/documentation/service-pulse.md#reactivity) that using `computed()` to watch for changes in a given value is a best practice.  So, I'm not sure if I am doing this right by binding the model to an input component like this?

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
TravisNickels: We could do it this way, but I would definitely have to update some other things.  In the `getEditAndRetryConfig()` I set the edit and retry configuration to `configuration.value.edit`, but if the regular configuration returns after the edit and retry config, which sets the root object of the configuration (`configuration.value`), then the edit and retry config will be lost.  I could set the regular configuration to something like `configuration.value.message` and the edit and retry config to `configuration.value.edit`, which should make it so that it wouldn't matter what order the requests finish.

src/ServicePulse.Host/vue/src/components/failedmessages/EditMessageHeader.vue
WilliamBZA: Ok, got it :+1:

PR #1427, opened 2023-03-10
Poll more frequently when operations are underway

WilliamBZA:
Changes the polling frequency based on whether or not a retry/delete/restore operation is underway.

Changes from polling every 5 seconds to every second.


WilliamBZA:
I've added the polling changes to the individual pages too @soujay @TravisNickels

soujay (approved):

TravisNickels (approved):

PR #1446, opened 2023-03-15
Don't allow spaces in comments

WilliamBZA:
Validates that a comment isn't just a bunch of spaces. There must be at least some text in there.


WilliamBZA:
Replaced by https://github.com/Particular/ServicePulse/pull/1449

PR #1447, opened 2023-03-15
Use routerlink for edit of navigation instead of a tag

WilliamBZA:
This prevents the full page refresh on navigation


TravisNickels:
This updates the url but the `loadFailedMessage()` function is never called again so the data does not get updated.

TravisNickels (approved):

PR #1488, opened 2023-05-15
Allow displaying logical endpoint

WilliamBZA:
This adds a dropdown to the filter section which allows the user to show only the endpoints' logical names instead of always physical.

image

It also includes some summary information about how many endpoints are beating and how many used to be beating.

image

image

Configuration still allows the user to disable individual instances.

image

Documentation PR: https://github.com/Particular/docs.particular.net/pull/6069

### DIFF
--- a/src/ServicePulse.Host/angular/app/js/.eslintrc.js
+++ b/src/ServicePulse.Host/angular/app/js/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
     "env": {
         "browser": true,
-        "es6": false,
+        "es6": true,
         "node": false,
         "jquery": true
     },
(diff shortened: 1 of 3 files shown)


WilliamBZA:
I've updated the PR to save the display type in cookies and to hide the display dropdown on the configuration tab

soujay (approved):

src/ServicePulse.Host/angular/app/js/.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
     "env": {
         "browser": true,
-        "es6": false,
+        "es6": true,
```
boblangley, line 4: If I remember right from how you explained it: This is OK because we don't support IE anymore. Correct?

boblangley (approved):

src/ServicePulse.Host/angular/app/js/.eslintrc.js
WilliamBZA: Correct

PR #1516, opened 2023-06-01
Use RepoRootDir property instead of SolutionDir

bording:
This aligns the Setup project with the approach used in ServiceControl, where the required value is set in the project file, so it will always be available regardless of how the project is built.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Build Windows installer
         env:
           AZURE_KEY_VAULT_CLIENT_SECRET: ${{ secrets.AZURE_KEY_VAULT_CLIENT_SECRET }}
-        run: dotnet build src/Setup -property:SolutionDir=..\..\..\..\ --configuration Release
+        run: dotnet build src/Setup --configuration Release
       - name: Publish artifacts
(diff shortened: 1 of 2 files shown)


tmasternak (approved):

WilliamBZA:
Can this be tested somehow? It seems that at some point after [this](https://github.com/Particular/ServicePulse/pull/1483) was made the `SolutionDir` variable started being undefined. Without actually exercising changes to the release process, how can we be confident that we won't be breaking it again in the future?

tmasternak:
@WilliamBZA it know of 2 possible options:
1. Merge this PR and after backporting to `release-1.36` burnt one patch version to verify the `release` workflow
2. Copy the installer building parts of the `release` workflow to `ci` and test if the package is properly built.

WilliamBZA:
That's fine for this specific PR, I guess, but this isn't the first time a release process has been broken. It would be nice for it to be more reliable, which to me means some way to test it

bording:
> That's fine for this specific PR, I guess, but this isn't the first time a release process has been broken. It would be nice for it to be more reliable, which to me means some way to test it

I've been talking with David to brainstorm some ways that we could have most of the release workflow being run along with the normal CI workflow, which should improve things on this front.

PR #1532, opened 2023-06-17
Backport edit retry multiple edits

soujay:
backport Disable "Edit & Retry" on an already edited and retried message #1530 -

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
+++ b/src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
@@ -400,7 +400,7 @@ onUnmounted(() => {
           <div class="row">
             <div class="col-sm-12 no-side-padding">
               <div class="active break group-title">
-                <h1 class="message-type-title">{{ failedMessage.message_type }} TESTING</h1>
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -400,7 +400,7 @@ onUnmounted(() => {
           <div class="row">
             <div class="col-sm-12 no-side-padding">
               <div class="active break group-title">
-                <h1 class="message-type-title">{{ failedMessage.message_type }}</h1>
+                <h1 class="message-type-title">{{ failedMessage.message_type }} TESTING</h1>
```
WilliamBZA, line 403: This does not seem correct

WilliamBZA (changes requested):

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
soujay: Yes,  that was test code and was removed in master .

mauroservienti:
Why are we backporting? Aren't apps li ServicePulse using GitHub flow (no release branches)?

Issue #1469, opened 2023-03-23
Connection test fails if the url does not end with a slash (version 1.33.1)


ramonsmits:
#### Expected behavior

Testing connectivity to `http://zolder:33633` should be successful.

#### Actual behavior

Testing connectivity to `http://zolder:33633` fails, but if you add a slash (i.e. ``http://zolder:33633/` it succeeds.

#### Versions

Please list the version of the relevant packages or applications in which the bug exists.        

### Steps to reproduce

1. Open Configuration Tab
2. Select Connections sub tab
3. Enter value without a slash at the end like `http://zolder:33633`
4. Click Test
5. Test shows "CONNECTION FAILED"

soujay:
Will be released in SP 1.36.5

soujay:
The bug has been fixed in ServicePulse 1.37.0 and has been released.
https://github.com/Particular/ServicePulse/releases/tag/1.37.0

PR #1538, opened 2023-06-25
add trailing slash to monitoring url

soujay:
This PR fixes [Connection test fails if the url does not end with a slash  #1469](https://github.com/Particular/ServicePulse/issues/1469)

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/configuration/PlatformConnections.vue
+++ b/src/ServicePulse.Host/vue/src/components/configuration/PlatformConnections.vue
@@ -42,7 +42,12 @@ function testServiceControlUrl(event) {
 
 function testMonitoringUrl(event) {
   if (event) {
-    testingMonitoring.value = true;
+      testingMonitoring.value = true;
+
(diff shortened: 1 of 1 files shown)


WilliamBZA:
Should we remove https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.js#L108 as part of this PR?

soujay:
> Should we remove https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.js#L108 as part of this PR?

@WilliamBZA  - No, Looks like that function is called while updating the urls . So if there is no value in the monitoring url text box and the save button  is clicked it considers monitoring to be disabled and adds in a "!" and the "Test" button is disabled.

WilliamBZA:
> if there is no value in the monitoring url text box and the save button is clicked it considers monitoring to be disabled and adds in a "!" and the "Test" button is disabled.

Why wouldn't that be correct behaviour?

WilliamBZA (approved):

PR #1619, opened 2023-11-19
Basic graph for the endpoint list

soujay:
This PR includes the following changes
- Basic graph component with and without datapoints
- typo fixes


src/ServicePulse.Host/vue/src/components/monitoring/D3Graph.vue
```diff
@@ -0,0 +1,141 @@
...
+
+  var scaleX = d3
+    .scaleLinear()
+    .domain([0, numberOfPoints - 1])
+    .range([horizontalMargin, width - horizontalMargin]);
+
+    var area = d3
+        .area()
+        .x(function (d, i) {
+            console.log(d); //dummy call to get past eslint error
```
cquirosj, line 61: Could we try something to avoid leaving the console.log? Like using deconstructing for assigning the non-used parameter to a lowercase dash _ and then configuring lint options to allow for that pattern? https://eslint.org/docs/latest/rules/no-unused-vars#varsignorepattern https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern

cquirosj (changes requested):

src/ServicePulse.Host/vue/src/components/monitoring/D3Graph.vue
soujay: Agreed. I will work on it in a separate PR

cquirosj: Sounds good. Approving now. Let's add a task so we don't forget to configure ling and remove the console.log

cquirosj (approved):

PR #1620, opened 2023-11-20
Endpoint details graphs

soujay:
This PR includes the following changes

- Large graph in endpoint details
- Small graph in endpoint details
- Default stories for large graph
The PR  has some dummy code to get past the eslint errors as well as commented code for the average label. I will be working on addressing those two in separate PRs


src/ServicePulse.Host/vue/src/components/monitoring/D3LargeGraph.stories.js
```diff
@@ -0,0 +1,57 @@
...
+    plotwidth: 750,
+    plotheight: 200,
+    firstseriescolor: "#EA7E00",
+    firstseriesfillcolor: "#EADDCE",
+    secondseriescolor: "",
+    secondseriesfillcolor: "",
+  },
+};
+
+export const ThroughputRetries = {
```
cquirosj, line 34: I see the story data passes two different data series; however, the story doesn't showcase the distinct two different colors nor the average line. This is how it looks when I run it:

image

But I was expecting to see something like this. This screenshot uses different data, colors, and scale units, but you get the idea. It showcases two data series with different colors and shows the average line. 

image

src/ServicePulse.Host/vue/src/components/monitoring/D3LargeGraph.stories.js
```diff
@@ -0,0 +1,57 @@
...
+
+//stories
+
+export const Default = {
+
+    minimumyaxis: 10,
+    avglabelcolor: "#EA7E00",
+    isdurationgraph: false,
+    metricsuffix: "MSGS",
+    csclass: "large-graph pull-left",
```
cquirosj, line 19: I think we can encapsulate these classes and also remove the following properties: firstseriescolor: "#EA7E00",
    firstseriesfillcolor: "#EADDCE",
    secondseriescolor: "",
    secondseriesfillcolor: "",

Instead, let the graph component use CSS specificity to show the corresponding colors. For example, see how the static version uses CSS specificity:

image

All the different colors are already in the CSS in a way where it is expected only to set a css marker class, which establishes the specificity 

image

src/ServicePulse.Host/vue/src/components/monitoring/D3LargeGraph.stories.js
soujay: The datapoints for both the first and second series are the same and hence the graphs overlapped. I have modified that second data series so that the 2 graphs are seen more distinctly.

src/ServicePulse.Host/vue/src/components/monitoring/D3LargeGraph.stories.js
soujay: I agree that there are many redundancies that needs to be removed and code optimized. I was planning to do that once I got all the pieces of the graph component working .

cquirosj: Sounds good @soujay I will approve the PR based on that plan.

cquirosj (approved):

PR #1622, opened 2023-11-21
Vue License toast notification

TravisNickels:
This PR:

- Makes sure that license toast notifications are getting displayed
- Cleans up the license toast notifications for Vue
- Moves all toast related css to the toast.css file
- Removes duplicate css


src/ServicePulse.Host/vue/src/composables/serviceLicense.js
```diff
@@ -40,16 +40,19 @@ export const licenseStatus = reactive({
...
+        displayWarningMessage(newValue);
+      }
     }
-  });
+  );
 
   return getLicense(license)
     .then((lic) => {
-      license = lic;
+      Object.assign(license, lic);
```
soujay, line 55: Is there any specific reason, you had use the Object.assign?

soujay (approved):

cquirosj (approved):

src/ServicePulse.Host/vue/src/composables/serviceLicense.js
TravisNickels: Yes, because the assignment operator `=` creates a new reference to the original object, which is not reactive. If you want to create a reactive copy of an object, we should be using `Object.assign` or the spread operator `...`

https://stackoverflow.com/questions/65732144/vue-js-3-replace-update-reactive-object-without-losing-reactivity

PR #1636, opened 2023-12-06
Display monitoring not available 

soujay:
This PR fixes css and content for monitoring not available


TravisNickels:
I noticed that there are some lines on the top and bottom of the `no monitoring` component.  I'm fine with approving this PR, but just wanted to point that out.

image

TravisNickels (approved):

soujay:
@TravisNickels - When I increased the contrast on my screen, I was able to see those 2 lines. Thanks for bringing it up. I have pushed out a commit that removes those lines.

PR #1647, opened 2023-12-13
refresh interval for endpoint details and list page

soujay:
Add in the refresh interval for endpoint list and endpoint details page


cquirosj (approved):

TravisNickels:
The expected functionality from the angular page is when you select the period history the graph should update immediately with the appropriate data from the select period and then it should continue to update at the interval selected from the period history.  It just feels weird to select an option from the period and have to wait to see the graph update.

cquirosj:
> It just feels weird to select an option from the period and have to wait to see the graph update.

I agree; the user shouldn't have to wait especially if that is not what is happening in the AngularJS version. Let's take this PR as a stepping stone, but let's ensure the same expected functionality gets migrated to the vue version.

Issue #1559, opened 2023-08-29
Paging in the Events view is broken


tmasternak:
### Describe the bug

In the Events view (under `https://<host>/a/#/events`) the last page of the results is incorrect. It shows data both from the last page and the second to last page. E.g. if the page size is 20 and the total count of items is 61 then the last page shows 21 items instead of just 1.

saschanm:
Possibly related but with a slightly different presentation.

Since upgrading ServiceControl to v5 both the last 10 events on the dashboard and the events tab paging are showing in reverse order

Dashboard shows the first 10 events after upgrade (and within retention period - 14 days for us) rather than the most recent.
Have to go to last page in events tab to see most recent - and those seem to also include some earliest events from 14 days back.

This suggests maybe a combination of this bug and another (maybe best as a ServiceControl issue)

mikeminutillo:
Fixed by #1665

PR #1665, opened 2024-01-11
Migrate events view

mikeminutillo:
Fixes https://github.com/Particular/ServicePulse/pull/1665
Fixes https://github.com/Particular/ServicePulse/issues/1559

- Migrates the Events page to Vue
- Added Pagination and ItemsPerPage re-usable controls
- Added a DataView reusable control that knows how to load paginated data from ServiceControl, and autorefresh it.


src/ServicePulse.Host/vue/src/views/EventsView.vue
```diff
@@ -0,0 +1,184 @@
...
+import { licenseStatus } from "../composables/serviceLicense.js";
+import { useGetEventLogItems } from "../composables/serviceEventLogItems.js";
+import TimeSince from "../components/TimeSince.vue";
+import LicenseExpired from "../components/LicenseExpired.vue";
+
+const eventLogItems = ref([]);
+const eventLogItemsPerPage = ref(25);
+
+function getEventLogItems() {
+    return useGetEventLogItems().then((result) => {
```
PhilBastian, line 12: was discussed in the pulse and control channel that samples like this can be converted to async/await rather than callback syntax

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,177 @@
...
+}
+
+function stopRefreshTimer() {
+  if (refreshTimer.value) {
+    clearInterval(refreshTimer.value);
+  }
+}
+
+onMounted(() => {
+  stopRefreshTimer();
```
johnsimons, line 118: Do we need to stop?

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,177 @@
...
+  startRefreshTimer();
+  loadData();
```
johnsimons, line 120: Given the timer will load the data, do we need to load here?

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,177 @@
...
+
+const props = defineProps({
+  apiUrl: String,
+  itemKey: {
+    type: String,
+    default: "id",
+  },
+  itemsPerPageOptions: {
+    type: Array,
+    default: () => [],
```
johnsimons, line 13: I feel this could be defaulted. 
I don't really see a reason for all consumers to have to think about it, thoughts?

johnsimons (approved):
@mikeminutillo really nice work on the `DataView`

A few observations:
- We have "autorefresh" lists, but IMO, they do make sense for the Dashboard (last 10 items) but for pageable lists, IMHO it feels weird, thoughts?
- To me the organisation of the code feels a bit odd, I would add a subfolder in the views folder, named "events" and add to it the EventItemShort and EventLogItem. And also move Eventsview into it.
   To me "components" are for reusable things that are not really specific to the app that are shared everywhere

src/ServicePulse.Host/vue/src/views/EventsView.vue
```diff
@@ -0,0 +1,28 @@
...
+        <template #item="item">
+          <EventLogItem :eventLogItem="item" />
+        </template>
+      </DataView>
+    </div>
+  </template>
+</template>
+
+<style>
+/* TODO: Fix up styles */
```
johnsimons, line 23: what's left here?

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: I did not default it because I wanted the default to be not to show it at all. I defaulted the values and included a boolean flag to show or hide it so it works the same as the pagination.

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: The timer does not go off immediately so you would need to wait 5 seconds (or whatever the timeout is) for the data to initially appear.

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: Removed.

src/ServicePulse.Host/vue/src/components/DataView.vue
johnsimons: i see

src/ServicePulse.Host/vue/src/views/EventsView.vue
mikeminutillo: Converted

src/ServicePulse.Host/vue/src/views/EventsView.vue
mikeminutillo: I need to go through all of the styles and see what is appropriate to put in each location. It's a bit of a mess at the moment.

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,173 @@
...
+const showItemsPerPage = computed(() => {
+  return props.showItemsPerPage;
+});
+
+function changeItemsPerPage(value) {
+  itemsPerPage.value = value;
+  loadData();
+}
+
+function calculatePageNumbers() {
```
PhilBastian, line 51: I feel that this function should contain _all_ the logic for determining page numbers, and return an array of objects { label: (number or ...), setPage: function } which the template can do a single iteration over

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,173 @@
...
+  if (numberOfPages.value <= 10) {
+    return Array.from(Array(numberOfPages.value), (_, index) => index + 1);
```
PhilBastian, line 53: i can see what this is doing, but it's confusing and needs a comment. Also, supposedly faster to have Array(numberOfPages.value).fill(0).map((_, i) => i + 1)

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,173 @@
...
+    type: Array,
+    default: () => [20, 35, 50, 75],
+  },
+  itemsPerPage: {
+    type: Number,
+    default: 50,
+  },
+  autoRefresh: {
+    type: Number,
+    default: 0,
```
PhilBastian, line 17: i feel that the default should be null, since 0 is confusing and could be interpreted as being a constant refresh

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,173 @@
...
+      const data = await response.json();
+      items.value = data;
+    }
+  } catch (error) {
+    console.log(error);
+  }
+}
+
+function startRefreshTimer() {
+  if (props.autoRefresh > 0) {
```
PhilBastian, line 106: how does Vue handle prop refreshes (i.e. if the owning component decides to change the refresh rate)? I'm certain it won't go back through onMounted, which means this function won't be called again (and also the existing interval can't be cancelled)

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,173 @@
...
+
+onUnmounted(() => {
+  stopRefreshTimer();
+});
+</script>
+
+<template>
+  <slot name="data" v-bind="items"></slot>
+  <div class="row">
+    <div v-if="showItemsPerPage" class="pagination col-md-2">
```
PhilBastian, line 132: is this div going to be extracted to a component?

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,173 @@
...
+      <div class="dropdown">
+        <label class="control-label">Items Per Page:</label>
+        <button type="button" class="btn btn-default dropdown-toggle sp-btn-menu" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+          {{ itemsPerPage }}
+          <span class="caret"></span>
```
PhilBastian, line 137: if elements aren't going to contain anything then it's neater to self close <span />

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -0,0 +1,173 @@
...
+        <li class="page-item" :class="{ disabled: pageNumber === 1 }">
+          <a class="page-link" href="#" @click.prevent="previousPage">Previous</a>
+        </li>
+        <li v-if="pageNumber > 5 && numberOfPages > 10" class="page-item">
+          <a @click.prevent="setPage(1)" class="page-link" href="#">1</a>
+        </li>
+        <li v-if="pageNumber > 5 && numberOfPages > 10" class="page-item">
+          <a @click.prevent="setPage(pageNumber - 5)" class="page-link" href="#">...</a>
+        </li>
+        <li v-for="n in calculatePageNumbers()" class="page-item" :class="{ active: pageNumber == n }" :key="n">
```
PhilBastian, line 157: ===

src/ServicePulse.Host/vue/src/components/EventItemShort.vue
```diff
@@ -1,109 +1,25 @@
...
-                <time-since :dateUtc="eventLogItem.raised_at"></time-since>
-              </div>
-            </div>
-          </div>
-        </div>
-      </div>
-
-      <div class="row text-center">
+  <div class="events">
+    <DataView api-url="eventlogitems" :auto-refresh="5000" :itemsPerPage="10" :show-pagination="false">
```
PhilBastian, line 8: can change autorefresh value to be seconds rather than milliseconds for readability?

PhilBastian: itemsPerPage should probably be application wide for usability

src/ServicePulse.Host/vue/src/components/EventItemShort.vue
```diff
@@ -1,109 +1,25 @@
...
+      <template #data="items">
         <div class="col-12">
-          <a v-if="eventCount > 10" class="btn btn-default btn-secondary btn-all-events" href="/a/#/events">View all events</a>
+          <h6>Last 10 events</h6>
+          <EventLogItem v-for="item in items" :eventLogItem="item" :key="item.id" />
```
PhilBastian, line 12: point to note: Vue allows you to use for in or for of, but in raw JavaScript they do different things (in iterates through the keys of the object rather than the items in the enumerable). For this reason I'd favour sticking to for of

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
```diff
@@ -0,0 +1,186 @@
...
+import TimeSince from "../components/TimeSince.vue";
+
+defineProps({
+  eventLogItem: Object,
+});
+const router = useRouter();
+
+function navigateToEvent(eventLogItem) {
+  switch (eventLogItem.category) {
+    case "Endpoints":
```
PhilBastian, line 12: suggestion to create eventLogItemCategory as a frozen object containing all the possible values, so that spelling mistakes or changes become a non-issue (see flowdiagram for an example)

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
```diff
@@ -0,0 +1,186 @@
...
+      router.push("/configuration/endpoint-connection");
```
PhilBastian, line 13: these routes are essentially defined elsewhere (in index.js), so should be exported constants from a module

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
```diff
@@ -0,0 +1,186 @@
...
+      break;
+    case "CustomChecks":
+      window.location = "/a/#/custom-checks";
+      break;
+    case "EndpointControl":
+      window.location = "/a/#/endpoints";
+      break;
+    case "MessageFailures":
+      var newlocation = "/failed-messages";
+      if (eventLogItem.related_to && eventLogItem.related_to.length > 0 && eventLogItem.related_to[0].search("message") > 0) {
```
PhilBastian, line 26: first and second clause can be combined to eventLogItem.related_to?.length (don't need to check for >0 since 0 is falsy)

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
PhilBastian: same for other string comparisons below

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
```diff
@@ -0,0 +1,186 @@
...
+}
+</script>
+
+<template>
+  <div class="row box box-event-item">
+    <div class="col-12" @click="navigateToEvent(eventLogItem)">
+      <div class="row">
+        <div class="col-1">
+          <span class="fa-stack fa-lg">
+            <i class="fa fa-stack-2x" :class="iconClasses(eventLogItem)"></i>
```
PhilBastian, line 69: self close i elements

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
```diff
@@ -0,0 +1,186 @@
...
+  top: -8px;
+}
+
+.events-list .col-message p.lead {
+  padding-bottom: 2px;
+}
+
+.events-list .col-timestamp {
+  display: table-cell;
+  width: 200px;
```
PhilBastian, line 153: i suggest starting to change px sizing to em sizing when it's being touched by migrations or changes

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
```diff
@@ -0,0 +1,186 @@
...
+
+.btn.sp-btn-menu:active,
+.btn-default.sp-btn-menu:active,
+.btn-default.sp-btn-menu.active,
+.open > .dropdown-toggle.btn-default.sp-btn-menu {
+  background: none;
+  border: none;
+  color: #00a3c4;
+  text-decoration: underline;
+  -webkit-box-shadow: none;
```
PhilBastian, line 183: i don't think this is necessary - Vue has a feature to auto fill vendor specific styling options

src/ServicePulse.Host/vue/src/views/EventsView.vue
```diff
@@ -0,0 +1,32 @@
...
+
+.events {
+  margin-top: 30px;
```
PhilBastian, line 30: em sizing

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: Extracted

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
mikeminutillo: Applied

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: Added a `watch` for this change

src/ServicePulse.Host/vue/src/components/EventItemShort.vue
mikeminutillo: > can change autorefresh value to be seconds rather than milliseconds for readability?

Done

> itemsPerPage should probably be application wide for usability

Not sure what you mean by this but in this case it is hard-coded to show 10 for the dashboard.

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
mikeminutillo: Switched to named routes (which still uses magic strings but it's better). Given the relatively low number of pages, I wonder if we should have a `navigate` service or something `navigate.toEndpointConnection();`

src/ServicePulse.Host/vue/src/components/EventItemShort.vue
mikeminutillo: Applied

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: This was completely re-written

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
johnsimons: @mikeminutillo from previous experience, I would still make it composable, but make the links an exported object.
Example:
```js
const routeLinks = {
  dashboard: {
     root: "/",
  },
  configuration: {
     root: "/configuration",
     connection: "/configuration/connections",
  },
  ....
}

export default routeLinks;
```
and then:
```js
import routeLinks from "routeLinks"

<a href="routeLinks.configuration.root">Configuration</a>
```

src/ServicePulse.Host/vue/src/components/PaginationStrip.vue
```diff
@@ -43,51 +39,105 @@ function setPage(page) {
...
-      <li class="page-item" :class="{ disabled: pageNumber === numberOfPages }">
-        <a class="page-link" href="#" @click.prevent="nextPage">Next</a>
+      <li v-for="page of pages" class="page-item" :key="page.key">
+        <a class="page-link" @click="page.click" :class="page.class">{{ page.label }}</a>
       </li>
     </ul>
   </div>
 </template>
+
+<style>
```
johnsimons, line 134: `<style scoped>`
https://vuejs.org/api/sfc-css-features.html#scoped-css

mikeminutillo: Applied

src/ServicePulse.Host/vue/src/views/EventsView.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/views/EventsView.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
mikeminutillo: Done

PhilBastian (approved):

src/ServicePulse.Host/vue/src/components/EventLogItem.vue
mikeminutillo: Removed

PR #1670, opened 2024-01-17
SPIKE: change fontawesome from v4.4 to v6

PhilBastian:
v4.4 of fontawesome requires the entire css and fontset to be delivered with the bundled code on every page load. v6 allows us to only take the icons that are actually being used, and also delivers them as SVG.

This PR starts on the conversion (converting the footer icons, not the header icons or any others) as a pattern to follow.

The downside of this change is that the icons will change look slightly, which won't be reflected in the angularJS pages
v4.4
image

v6 (lack of padding needs to be fixed)
image

### DIFF
--- a/src/ServicePulse.Host/vue/package-lock.json
+++ b/src/ServicePulse.Host/vue/package-lock.json
@@ -8,6 +8,10 @@
       "name": "service-pulse",
       "version": "1.0.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


PhilBastian:
closing spike awaiting finalisation of AngularJS to Vue migration. Related issue remains open

PR #1676, opened 2024-01-19
Fix header change actions

mikeminutillo:
## Symptoms

When the [edit and retry](https://docs.particular.net/servicepulse/intro-editing-messages) functionality is enabled, clicking the button to delete or reset a header navigates to the dashboard. This prevents the user from completing the retry operation.

## Root cause

The actions are anchor tags which had an `href="#"` attribute. With the href in place the browser navigates away from the modal. This can also be solved with `.prevent` event modifiers but the href is not needed.

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/failedmessages/EditMessageHeader.vue
+++ b/src/ServicePulse.Host/vue/src/components/failedmessages/EditMessageHeader.vue
@@ -49,10 +49,10 @@ onMounted(() => {
     <input :class="{ 'header-removed': header.isMarkedAsRemoved }" class="form-control" :disabled="header.isLocked" v-model="header.value" />
   </td>
   <td>
-    <a v-if="!header.isLocked && !header.isMarkedAsRemoved" href="#" @click="markHeaderAsRemoved()">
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/vue/src/components/failedmessages/EditMessageHeader.vue
```diff
@@ -49,10 +49,10 @@ onMounted(() => {
     <input :class="{ 'header-removed': header.isMarkedAsRemoved }" class="form-control" :disabled="header.isLocked" v-model="header.value" />
   </td>
   <td>
-    <a v-if="!header.isLocked && !header.isMarkedAsRemoved" href="#" @click="markHeaderAsRemoved()">
+    <a v-if="!header.isLocked && !header.isMarkedAsRemoved" @click="markHeaderAsRemoved()">
```
PhilBastian, line 52: the general guidance on template generation is to not use anchors unless you're actually linking to something. These should be buttons, not anchors

PhilBastian (changes requested):
shouldn't use anchors for template items which aren't links

PhilBastian (approved):
approved based on conversation that the anchor issues will be fixed as a separate issue

PR #1680, opened 2024-01-22
Enable no-floating-promises

mikeminutillo:
Detects missing await on async function calls https://github.com/SebastienGllmt/eslint-plugin-no-floating-promise

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/failedmessages/MessageGroupList.vue
+++ b/src/ServicePulse.Host/vue/src/components/failedmessages/MessageGroupList.vue
@@ -109,7 +109,7 @@ async function saveDeleteNote(group, hideToastMessage) {
       useShowToast("info", "Info", "Note deleted succesfully");
     }
 
-    loadFailedMessageGroups(); //reload the groups
+    await loadFailedMessageGroups(); //reload the groups
   } else {
     noteSaveSuccessful.value = false;
     if (!hideToastMessage) {
@@ -129,19 +129,19 @@ async function saveNote(group) {
   if (result.message === "success") {
     noteSaveSuccessful.value = true;
     useShowToast("info", "Info", "Note updated successfully");
-    loadFailedMessageGroups(); //reload the groups
+    await loadFailedMessageGroups(); //reload the groups
   } else {
     noteSaveSuccessful.value = false;
     useShowToast("error", "Error", "Failed to update Note:" + result.message);
   }
 }
 
-function saveCreatedNote(group) {
-  saveNote(group);
+async function saveCreatedNote(group) {
+  await saveNote(group);
 }
 
-function saveEditedNote(group) {
-  saveNote(group);
+async function saveEditedNote(group) {
+  await saveNote(group);
 }
 
 function editNote(group) {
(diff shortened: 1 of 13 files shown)


src/ServicePulse.Host/vue/src/composables/serviceServiceControl.js
```diff
@@ -79,9 +79,9 @@ export const connections = reactive({
 });
 
 export function useServiceControl() {
-  onMounted(() => {
-    useServiceControlStats();
-    useServiceControlMonitoringStats();
+  onMounted(async () => {
+    await useServiceControlStats();
+    await useServiceControlMonitoringStats();
```
mikeminutillo, line 84: These should potentially be done in parallel with a `Promise.all`

src/ServicePulse.Host/vue/src/components/failedmessages/MessageGroupList.vue
```diff
@@ -129,19 +129,19 @@ async function saveNote(group) {
...
+    await loadFailedMessageGroups(); //reload the groups
   } else {
     noteSaveSuccessful.value = false;
     useShowToast("error", "Error", "Failed to update Note:" + result.message);
   }
 }
 
-function saveCreatedNote(group) {
-  saveNote(group);
+async function saveCreatedNote(group) {
```
PhilBastian, line 139: these ones could just return saveNote without having to be marked async

src/ServicePulse.Host/vue/src/components/failedmessages/MessageGroupList.vue
```diff
@@ -166,7 +166,7 @@ async function saveDeleteGroup(group) {
   // We've started a delete, so increase the polling frequency
   changeRefreshInterval(1000);
 
-  saveDeleteNote(group, true); // delete comment note when group is archived
+  await saveDeleteNote(group, true); // delete comment note when group is archived
```
PhilBastian, line 169: this is changing behaviour to wait for the delete note promise to complete before continuing. If this isn't necessary then it will affect performance

src/ServicePulse.Host/vue/src/components/failedmessages/MessageGroupList.vue
```diff
@@ -209,7 +209,7 @@ async function saveRetryGroup(group) {
   // We've started a retry, so increase the polling frequency
   changeRefreshInterval(1000);
 
-  saveDeleteNote(group, true);
+  await saveDeleteNote(group, true);
```
PhilBastian, line 212: this is changing behaviour to wait for the delete note promise to complete before continuing. If this isn't necessary then it will affect performance

src/ServicePulse.Host/vue/src/composables/serviceServiceControl.js
PhilBastian: definitely, if equivalent behaviour is desired

PhilBastian (commented):
many of these changes change the behaviour/timing of function execution, which could lead to bugs

mikeminutillo:
Closing for now. We can revisit this later if needed.

PR #1692, opened 2024-02-05
Apply pagination strip

mikeminutillo:
https://github.com/Particular/ServicePulse/pull/1665 added a pagination strip re-usable component. This PR applies the component everywhere in the application where pagination has been added to provide a consistent user experience.

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
+++ b/src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { onMounted, onUnmounted, ref } from "vue";
+import { onMounted, onUnmounted, ref, watch } from "vue";
 import { licenseStatus } from "../../composables/serviceLicense";
 import { connectionState } from "../../composables/serviceServiceControl";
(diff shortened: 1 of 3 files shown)


johnsimons (approved):
very nice to see soo much code being deleted

johnsimons:
@mikeminutillo the only comment is it seems we are repeating `const per page = 50;` in all the pages.
I'm wondering if that should be the component default.

mikeminutillo:
> @mikeminutillo the only comment is it seems we are repeating `const per page = 50;` in all the pages. I'm wondering if that should be the component default.

The component can't have a default unless the call to service control also passes the same value (i.e. the SC page default is 50). If either of those change we will get an inconsistent UI. I opted to make it explicit and always pass it to SC to guarantee consistency. It could maybe be extracted as an app-wide constant that gets imported in where it is used.

src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
```diff
@@ -45,6 +46,10 @@ const sortOptions = [
   },
 ];
 
+watch(pageNumber, () => {
```
PhilBastian, line 49: remove the curly brackets so this gets single lined like the file above

PhilBastian (approved):

PR #1694, opened 2024-02-05
Removing .js from imports

johnsimons:
This is quite a few files, but is just removing the `.js` from the imports, so that once we rename the files to `.ts` we don't need to revisit these imports.

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
+++ b/src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
@@ -1,9 +1,9 @@
 <script setup>
 import { ref, onMounted, onUnmounted } from "vue";
-import { licenseStatus } from "../../composables/serviceLicense.js";
-import { connectionState } from "../../composables/serviceServiceControl.js";
-import { useFetchFromServiceControl, usePatchToServiceControl } from "../../composables/serviceServiceControlUrls.js";
-import { useShowToast } from "../../composables/toast.js";
+import { licenseStatus } from "../../composables/serviceLicense";
+import { connectionState } from "../../composables/serviceServiceControl";
+import { useFetchFromServiceControl, usePatchToServiceControl } from "../../composables/serviceServiceControlUrls";
+import { useShowToast } from "../../composables/toast";
 import { useRetryMessages } from "../../composables/serviceFailedMessage";
 import { useDownloadFile } from "../../composables/fileDownloadCreator";
 import { useRoute, onBeforeRouteLeave } from "vue-router";
@@ -93,7 +93,7 @@ function loadPagedMessages(groupId, page, sortBy, direction) {
       messages.value = data;
     } catch (err) {
       console.log(err);
-      var result = {
+      const result = {
         message: "error",
       };
       return result;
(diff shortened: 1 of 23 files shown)


mikeminutillo (approved):

src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
```diff
@@ -147,7 +147,7 @@ function exportSelected() {
     let d = {};
 
     if (type == "array" || type == "object") {
-      for (let i in obj) {
+      for (const i in obj) {
```
PhilBastian, line 150: terrible variable name... I suggest key

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -37,7 +37,7 @@ async function loadFailedMessage() {
       failedMessage.value = { error: true };
     }
     const data = await response.json();
-    var message = data;
+    const message = data;
```
PhilBastian, line 40: I know this isn't new... but why?

src/ServicePulse.Host/vue/src/composables/serviceRedirects.js
```diff
@@ -37,15 +37,15 @@ async function getRedirects() {
...
+    const result = {
       message: response.ok ? "success" : "error:" + response.statusText,
       status: response.status,
       statusText: response.statusText,
     };
     return result;
   } catch (err) {
     console.log(err);
-    var result_1 = {
+    const result_1 = {
```
PhilBastian, line 48: i suspect this was from Mike using autoconvert to async await: can we just have this object returned rather than having a machine generated variable name?

src/ServicePulse.Host/vue/src/composables/serviceRedirects.js
```diff
@@ -68,7 +68,7 @@ export async function useUpdateRedirects(redirectId, sourceEndpoint, targetEndpo
...
```
PhilBastian, line 71: just return

src/ServicePulse.Host/vue/src/composables/serviceRedirects.js
```diff
@@ -90,7 +90,7 @@ export async function useCreateRedirects(sourceEndpoint, targetEndpoint) {
...
```
PhilBastian, line 93: just return

src/ServicePulse.Host/vue/src/composables/serviceRedirects.js
```diff
@@ -109,7 +109,7 @@ export async function useDeleteRedirects(redirectId) {
...
```
PhilBastian, line 112: just return

PhilBastian (approved):

src/ServicePulse.Host/vue/src/composables/serviceRedirects.js
johnsimons: I think there is an eslint about this, "unnecessary variable", can you check @PhilBastian ?

src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
johnsimons: we can address this in a different PR

PR #1695, opened 2024-02-05
Reformat with prettier and more eslint rules included

johnsimons:
So I did a complete reformat again of the whole code, not many changes from it.
I fixed all the eslint rules that were set as warnings, eg prefer-const, triple equality.
Finally I enabled a few more eslint rules to do with imports.

So lots of files changed but mostly just code style updated.


src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
```diff
@@ -141,18 +141,18 @@ function exportSelected() {
   }
 
   function parseObject(obj, propertiesToSkip, path) {
-    if (path == undefined) path = "";
+    if (path === undefined) path = "";
```
PhilBastian, line 144: should be comparing to == null rather than === undefined I would think?

src/ServicePulse.Host/vue/src/composables/serviceLicense.js
```diff
@@ -12,7 +12,7 @@ const upgradeProtectionExpired =
 const trialExpiring =
   '<div class="license-warning"><strong>Non-production development license expiring</strong><div>Your non-production development license will expire soon. To continue using the Particular Service Platform you\'ll need to extend your license.</div><a href="http://particular.net/extend-your-trial?p=servicepulse" class="btn btn-license-warning"><i class="fa fa-external-link-alt"></i> Extend your license</a><a href="#/configuration" class="btn btn-license-warning-light">View license details</a></div>';
 
-export var license = reactive({
+export let license = reactive({
```
PhilBastian, line 15: should be const

PhilBastian (approved):

src/ServicePulse.Host/vue/src/components/failedmessages/AllFailedMessages.vue
johnsimons: We are using the ["smart" mode](https://eslint.org/docs/latest/rules/eqeqeq#smart), and it seems to suggest `===`.
@PhilBastian is the "smart" mode the wrong setting?

src/ServicePulse.Host/vue/src/composables/serviceLicense.js
johnsimons: We reassign `license` later on.
Maybe we should not do that!

johnsimons:
@PhilBastian I have reverted the svg changes

PR #1697, opened 2024-02-06
Migrating composables and a few vue component

johnsimons:
Migrating composables and a few vue component

Started adding typings for REST api resources


src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts
```diff
@@ -1,29 +1,36 @@
...
+import { type Ref, ref } from "vue";
 
-export const serviceControlUrl = ref(null);
-export const monitoringUrl = ref(null);
+export const serviceControlUrl = ref<string>();
+export const monitoringUrl = ref<string>();
 
 export function useIsMonitoringDisabled() {
-  return monitoringUrl.value === null || monitoringUrl.value === "" || monitoringUrl.value === "!";
+  return monitoringUrl.value === undefined || monitoringUrl.value === "" || monitoringUrl.value === "!";
```
PhilBastian, line 7: i think we want this first clause to be monitoringUrl.value == null, since it depends on how the form is handled whether the clearing of a value will result in an empty string or null, whereas it starts above as undefined.

src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts
```diff
@@ -1,29 +1,36 @@
...
+  const response = await fetch(serviceControlUrl.value + suffix);
+  const data = await response.json();
+
+  return [response, data];
+}
+
+export function useFetchFromMonitoring(suffix: string) {
   if (useIsMonitoringDisabled()) {
-    return Promise.resolve(null);
+    return Promise.resolve(undefined);
```
PhilBastian, line 27: I really don't like returning undefined from functions: should be null

src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts
```diff
@@ -1,29 +1,36 @@
...
   return fetch(monitoringUrl.value + suffix);
 }
 
-export function usePostToServiceControl(suffix, payload) {
-  const requestOptions = {
+export function usePostToServiceControl(suffix: string, payload: object | undefined = undefined) {
+  const requestOptions: RequestInit = {
     method: "POST",
   };
   if (payload !== undefined) {
```
PhilBastian, line 36: will also not want to add these headers if payload is null, so check should be if payload != null (i.e. it was wrong in the old version)

src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts
```diff
@@ -72,7 +79,7 @@ export function useServiceControlUrls() {
     window.localStorage.setItem("scu", serviceControlUrl.value);
     console.debug(`ServiceControl Url found in QS and stored in local storage: ${serviceControlUrl.value}`);
   } else if (window.localStorage.getItem("scu")) {
-    serviceControlUrl.value = window.localStorage.getItem("scu");
+    serviceControlUrl.value = window.localStorage.getItem("scu") || undefined;
```
PhilBastian, line 82: || undefined doesn't make sense to me...? If the getItem doesn't find the key, won't it return undefined anyway?

src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts
```diff
@@ -115,16 +122,23 @@ export function updateServiceControlUrls(newServiceControlUrl, newMonitoringUrl)
   window.localStorage.removeItem("scu");
   window.localStorage.removeItem("mu");
 
-  const newSearch = "?scu=" + newServiceControlUrl.value + "&mu=" + newMonitoringUrl.value;
+  let newSearch = "?scu=" + newServiceControlUrl.value + "&mu=" + newMonitoringUrl.value;
```
PhilBastian, line 125: why?

src/ServicePulse.Host/vue/src/router/index.ts
```diff
@@ -109,7 +109,7 @@ const router = createRouter({
 });
 
 router.beforeEach((to, from, next) => {
-  document.title = to.meta.title || "ServicePulse";
+  document.title = `${to.meta.title} || "ServicePulse"`;
```
PhilBastian, line 112: this is changing the title from what it previously was

PhilBastian (changes requested):
in general i'd prefer to have the assignation to undefined and returning from functions of undefined changed to assigning/returning null, since using undefined is 'overloading' the default meaning of undefined as set by JavaScript of "this thing hasn't been defined, or this function has no return statement".

Comparisons to undefined should be comparisons to undefined or null (i.e. use the shortcut == null)

src/ServicePulse.Host/vue/src/composables/formatter.js
```diff
@@ -1,62 +0,0 @@
-import moment from "moment";
-
-const secondDuration = moment.duration(10 * 1000);
-const minuteDuration = moment.duration(60 * 1000);
-const hourDuration = moment.duration(60 * 1000); //this ensure that we never use minute formatting
-const dayDuration = moment.duration(24 * 60 * 60 * 1000);
-
-export function useFormatTime(value) {
```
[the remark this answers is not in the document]
johnsimons: Ah, I thought I could take a shortcut 😄 

So I converted all the functions now, but unsure about the function parameter types, I went with `value: string, decimals: number`. Is this right?

johnsimons (commented):
I think I addressed all comments @PhilBastian let me know

src/ServicePulse.Host/vue/src/components/TimeSince.vue
```diff
@@ -1,21 +1,18 @@
-﻿<script setup>
+﻿<script setup lang="ts">
 import { onMounted, onUnmounted, ref } from "vue";
 import moment from "moment";
 
-const props = defineProps({
-  dateUtc: String,
-  default: function () {
```
johnsimons, line 7: from reading the [doco](https://vuejs.org/guide/components/props.html#prop-validation), I think this was wrong. 
It should have been:
```js
{ 
  dateUtc: { type: String, default: ... },
}
```

src/ServicePulse.Host/vue/src/components/TimeSince.vue
```diff
@@ -1,21 +1,18 @@
...
-    return "0001-01-01T00:00:00";
-  },
-});
+const emptyDate = "0001-01-01T00:00:00";
+
+const props = withDefaults(defineProps<{ dateUtc: string }>(), { dateUtc: emptyDate });
 
-let interval = null;
+let interval: number | undefined = undefined;
```
johnsimons, line 9: I left this one as `undefined` @PhilBastian, mostly because the `window.clearInterval` takes a number or undefined

src/ServicePulse.Host/vue/src/composables/formatter.ts
```diff
@@ -0,0 +1,59 @@
...
+    return round(num, decimals);
+  }
+
+  const exp = Math.floor(Math.log(num) / Math.log(1000));
+
+  return round(num / Math.pow(1000, exp), decimals) + suffixes[exp - 1];
+}
+
+function round(num: number, decimals: number) {
+  return Math.round(num + Number(`e+${decimals}`)) + Number(`e-${decimals}`);
```
johnsimons, line 58: I hope I got this right.
I wasn't sure why it needs to be this complex!

src/ServicePulse.Host/vue/src/composables/serviceLicense.ts
```diff
@@ -12,19 +15,38 @@ const upgradeProtectionExpired =
...
   license_type: "",
   instance_name: "",
-  formattedInstanceName: "",
   trial_license: true,
-  license_status: "Unavailable",
-});
+  registered_to: "",
+  status: "",
+  license_status: LicenseStatus.Unavailable,
+  licenseEdition: computed(() => {
```
johnsimons, line 35: I decided to move the computed into the "empty/default" object, thoughts @PhilBastian ?

src/ServicePulse.Host/vue/src/composables/serviceSemVer.ts
```diff
@@ -41,7 +41,7 @@ export function useIsSupported(currentVersion: string, minSupportedVersion: stri
 }
 
 interface SemVer {
-  semver: string | undefined;
+  semver: string | null;
```
johnsimons, line 44: Going with `null`

src/ServicePulse.Host/vue/src/components/TimeSince.vue
johnsimons: anyway, I fixed it

src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts
```diff
@@ -148,11 +148,7 @@ function getParams() {
...
 
 function getParameter(params: Param[], key: string) {
-  if (params) {
-    return params.find((param) => {
-      return param.name === key;
-    });
-  }
-
-  return undefined;
+  return params.find((param) => {
```
PhilBastian, line 151: does this still need to be null aware? i.e. params?.find((param) => param.name === key)

PhilBastian (approved):

src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts
johnsimons: The assumption here is that params should either be empty array or not, but not null.
I also checked the usages of this and they seem to always pass an array.

PR #1700, opened 2024-02-07
Monitoring sorting columns

TravisNickels:
Added the column sorting functionality to the monitoring endpoint list for vue.


src/ServicePulse.Host/vue/src/components/monitoring/EndpointListSortableColumn.vue
```diff
@@ -1,19 +1,81 @@
...
+    required: true,
+    validator: (value) => ["name", "queueLength", "throughput", "retries", "processingTime", "criticalTime"].includes(value),
+  },
 });
+
+const monitoringStore = useMonitoringStore();
+const isActive = computed(() => props.isActive);
+const sortBy = computed(() => props.sortBy);
+const isAscending = ref(false);
+const sortIcon = ref("sort-down");
```
mikeminutillo, line 19: This could be a computed property off of `isAscending` and then you don't need to change it below

```suggestion
const sortIcon = computed(() => (isAscending.value ? "sort-up" : "sort-down"));
```

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListSortableColumn.vue
```diff
@@ -1,19 +1,81 @@
...
+  isActive: Boolean,
+  sortBy: {
+    type: String,
+    required: true,
+    validator: (value) => ["name", "queueLength", "throughput", "retries", "processingTime", "criticalTime"].includes(value),
+  },
 });
+
+const monitoringStore = useMonitoringStore();
+const isActive = computed(() => props.isActive);
```
mikeminutillo, line 16: You might consider passing in the active header as a `v-model` and then computing `isActive` off of that. If you do that then:
- You don't need a click handler on the parent
- The parent doesn't need to decide if this is active, the component itself can
- The parent can put a watch on the active column and update the monitoring store

This would make this a generic component that isn't unique to the EndpointList.
You _would_ need to update `activeColumn` in this component and you would need to expose the sort direction somehow. 

In this component:
```js

const activeColumn = defineModel({
  type: String,
  required: true,
});

const isActive = computed(() => activeColumn.value == props.sortBy);

// ...

function toggleSort() {
  activeColumn.value = props.sortBy;
  isAscending.value = isActive.value ? !isAscending.value : false;

  // ...
}
```

In the consumer component:
```js
<EndpointListSortableColumn 
  :sort-by="'queueLength'"
  v-model="activeColumn"
  v-tooltip
  title="Queue length: The number of messages waiting to be processed in the input queue(s) of the endpoint.">
  Queue Length
  <template #unit>(MSGS)</template>
</EndpointListSortableColumn
```

src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
```diff
@@ -30,23 +35,58 @@ onMounted(async () => {
...
       <div class="table-first-col">
-        <endpoint-list-sortable-column>Endpoint name</endpoint-list-sortable-column>
+        <EndpointListSortableColumn :is-active="activeColumn === 'name'" :sort-by="'name'" @click="updateActiveColumn('name')">Endpoint name</EndpointListSortableColumn>
       </div>
       <div class="table-col">
-        <endpoint-list-sortable-column v-tooltip title="Queue length: The number of messages waiting to be processed in the input queue(s) of the endpoint.">Queue Length<template #unit>(MSGS)</template></endpoint-list-sortable-column>
+        <EndpointListSortableColumn
+          :is-active="activeColumn === 'queueLength'"
+          :sort-by="'queueLength'"
+          @click="updateActiveColumn('queueLength')"
```
mikeminutillo, line 44: Needing to restate the column name 3 times could lead to subtle copy-paste issues. This might be addessed by the suggestion about using `v-model` below.

src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
```diff
@@ -30,23 +35,58 @@ onMounted(async () => {
     <!--Table headings-->
     <div class="table-head-row">
       <div class="table-first-col">
-        <endpoint-list-sortable-column>Endpoint name</endpoint-list-sortable-column>
+        <EndpointListSortableColumn :is-active="activeColumn === 'name'" :sort-by="'name'" @click="updateActiveColumn('name')">Endpoint name</EndpointListSortableColumn>
```
mikeminutillo, line 38: If you drop the colon, you don't need the extra quotes:

```suggestion
        <EndpointListSortableColumn :is-active="activeColumn === 'name'" sort-by="name" @click="updateActiveColumn('name')">Endpoint name</EndpointListSortableColumn>
```

mikeminutillo (approved):

PhilBastian (approved):

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListSortableColumn.vue
TravisNickels: This really helped me to look at this differently.  Thanks.  I was able to refactor a lot of stuff after reviewing this.

src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
```diff
@@ -30,23 +35,58 @@ onMounted(async () => {
...
       </div>
       <div class="table-col">
-        <endpoint-list-sortable-column v-tooltip title="Queue length: The number of messages waiting to be processed in the input queue(s) of the endpoint.">Queue Length<template #unit>(MSGS)</template></endpoint-list-sortable-column>
+        <EndpointListSortableColumn
+          :is-active="activeColumn === 'queueLength'"
+          :sort-by="'queueLength'"
+          @click="updateActiveColumn('queueLength')"
+          v-tooltip
+          title="Queue length: The number of messages waiting to be processed in the input queue(s) of the endpoint."
+          >Queue Length<template #unit>(MSGS)</template></EndpointListSortableColumn
```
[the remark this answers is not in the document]
TravisNickels: I'm having a little trouble with prettier.  the sortableColumn "Critical Time" is the only column that put each parameter on a new line.  When I tried to apply that to the other columns when I saved the file it would automatically set everything back.  Maybe I don't have it configured correctly.

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListSortableColumn.vue
```diff
@@ -1,19 +1,81 @@
...
+import { ref, computed } from "vue";
+import { useMonitoringStore } from "../../stores/MonitoringStore";
+
+const props = defineProps({
   unit: String,
+  isActive: Boolean,
+  sortBy: {
+    type: String,
+    required: true,
+    validator: (value) => ["name", "queueLength", "throughput", "retries", "processingTime", "criticalTime"].includes(value),
```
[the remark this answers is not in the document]
TravisNickels: > these values can be exported as properties of a frozen object from EndpointList.vue thus allowing them to be used similar to an enum throughout

I wasn't really sure what you meant by this.  I ended it up putting the frozen object in the EndpointList.vue file.  If I didn't understand this correctly, please let me know.

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListSortableColumn.vue
```diff
@@ -1,19 +1,81 @@
...
+
+.column-header-button div {
+  display: inline-block;
+}
+
+.sort-up,
+.sort-down {
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 8px;
```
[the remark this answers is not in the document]
TravisNickels: I was able to just completely remove the sizing.

src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
TravisNickels: I agree.  I wasn't happy with it either.  Changing this to a `v-model` worked a lot better.

cquirosj (approved):

src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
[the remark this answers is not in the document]
PhilBastian: no, prettier will single line below `printWidth` which is currently set to 250 in order to reduce the conflicts on the monitoring branch. Now that the componentisation release is complete, we can set this down to something sane, like 120

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListSortableColumn.vue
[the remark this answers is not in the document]
PhilBastian: yep, this was exactly what I meant. Then if you need those constants in any other module they can be exported from EndpointList.vue

PR #1701, opened 2024-02-07
Add NODE_OPTIONS environment variable to build.ps1

TravisNickels:
Adds the `NODE_OPTIONS = "--openssl-legacy-provider"` to the `.npmrc` file in the root directory for the angular app. This allows any `npm` commands run in the angular directory to use the `NODE_OPTIONS` environment variable, which allows compatibility with Node.js versions 17 and above.

The node-version in the ci.yml and release.yml have been pinned to version 21.6.x for now.  This version will need to be manually updated in future releases, as it is not picked up by dependabot.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,8 +27,6 @@ jobs:
           node-version: latest
       # Build AngularJS and Vue apps
       - name: Build AngularJS and Vue.js
-        #env:
-        #  NODE_OPTIONS: "--openssl-legacy-provider"
         run: .\build.ps1
         working-directory: src/ServicePulse.Host
       # .NET Build and run tests
@@ -42,7 +40,7 @@ jobs:
       - name: Run Javascript tests
         run: npm run test-firefox-ci
         working-directory: src/ServicePulse.Host.Tests
-        # Upload assets and packages
+      # Upload assets and packages
       - name: Upload assets
         uses: actions/upload-artifact@v4.3.1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,20 +22,27 @@ jobs:
         with:
           dotnet-version: 7.0.x
       - name: Set up Node.js
-        uses: actions/setup-node@v4.0.1
+        uses: actions/setup-node@v4.0.2
         with:
-            node-version: 16
-      - name: AngularJs - Install Host npm packages
-        run: npm run load
-        working-directory: src/ServicePulse.Host/angular
-      - name: AngularJs - Run lint
-        run: npm run lint
-        working-directory: src/ServicePulse.Host/angular
+          node-version: latest
+      # Build AngularJS and Vue apps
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -22,32 +22,50 @@ jobs:
         with:
           dotnet-version: 7.0.x
       - name: Set up Node.js
-        uses: actions/setup-node@v4.0.1
+        uses: actions/setup-node@v4
         with:
-            node-version: 16
+          node-version: latest
```
PhilBastian, line 27: do we want to use latest and allow the build failing to signal breaking node version changes, or would it be better to set the current latest version number and have each release responsible for attempting to upgrade to latest?

PhilBastian (approved):

.github/workflows/ci.yml
TravisNickels: I couldn't really find any documentation talking about needing to check the CI/CD workflows to make sure that their action's versions are up to date. I'm not really sure if there is a process in place for maintaining GitHub workflows and the versions. It feels like these versions could get stale pretty fast.  To me, it seems to make sense that we would want to make sure that our code is compatible with the latest versions of node, but I do acknowledge that having broken builds because of a node version update could be disruptive to the development cycle and dependabot PRs for bumping version in C# projects.

However, I'm also not opposed to having a set version to increase stability.  But maybe we could update the [releasing guidelines](https://github.com/Particular/Platform/blob/main/guidelines/releasing.md) or the [release workflow](https://github.com/Particular/Platform/blob/bce1aeb559c6147d542a02b482fd114b206cfd7f/guidelines/github-actions/release-workflow.md) documentation to indicate that when releasing a new version, the GitHub actions should be checked?

src/ServicePulse.Host/build.ps1
```diff
@@ -20,7 +20,9 @@ if ( $? -eq $false ) {
 cd ..
 
 cd $AngularSourceFolder
+$env:NODE_OPTIONS = "--openssl-legacy-provider"
```
mikeminutillo, line 23: I may have found a better way of achieving this, that will apply only to angular. I was trying to figure out what this setting actually does, to ensure that it will not have an impact on the final release. I did not figure that out, but I did find [this workaround](https://github.com/nodejs/node/issues/40455#issuecomment-1527619752).

To follow this, add a file named `.npmrc` in the root angular folder. Inside it, add the following:
```js
node-options="--openssl-legacy-provider"
```

Now, all npm commands in this folder will pass that to node. This has the benefit of being in the angular folder, so it will be removed when we move away from angular.

mikeminutillo (approved):

src/ServicePulse.Host/build.ps1
TravisNickels: I think this works better because it also works for the developer as well and not just for the workflows.  I have updated the PR to reflect the changes.

PR #1706, opened 2024-02-08
More vue converted to TypeScript

johnsimons:
I am not 100% happy with the generic `DataView` implementation.
I may raise an issue in vue to determine if there is a nicer way of doing that.

Apart from that, all the other conversions are straightforward.

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/DataView.vue
+++ b/src/ServicePulse.Host/vue/src/components/DataView.vue
@@ -2,7 +2,7 @@
 import { onMounted, onUnmounted, ref, watch } from "vue";
 // eslint-disable-next-line vue/prefer-import-from-vue
 import type { UnwrapRefSimple } from "@vue/reactivity";
-import { useTypedFetchFromServiceControl } from "../composables/serviceServiceControlUrls";
+import { useTypedFetchFromServiceControl } from "@/composables/serviceServiceControlUrls";
 import ItemsPerPage from "../components/ItemsPerPage.vue";
 import PaginationStrip from "../components/PaginationStrip.vue";
 
@@ -19,7 +19,7 @@ const props = withDefaults(
   { itemsPerPageOptions: () => [20, 35, 50, 75], itemsPerPage: 50, showPagination: true, showItemsPerPage: false }
 );
 
-const refreshTimer = ref<number>();
+let refreshTimer: number | undefined;
 const items = ref<T[]>([]);
 const pageNumber = ref(1);
 const itemsPerPage = ref(props.itemsPerPage);
@@ -46,14 +46,14 @@ async function loadData() {
 
 function startRefreshTimer() {
   if (props.autoRefreshSeconds) {
-    refreshTimer.value = setInterval(() => {
+    refreshTimer = setInterval(() => {
       loadData();
     }, props.autoRefreshSeconds * 1000);
   }
 }
 
 function stopRefreshTimer() {
-  clearInterval(refreshTimer.value);
+  clearInterval(refreshTimer);
 }
 
 onMounted(() => {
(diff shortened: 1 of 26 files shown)


src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -1,33 +1,26 @@
...
-    type: Boolean,
-    default: false,
-  },
-});
 
-const refreshTimer = ref();
-const items = ref([]);
+const props = withDefaults(
+  defineProps<{
+    itemsType: T[];
```
mikeminutillo, line 11: Would it make sense to turn this into a `v-model`? That way the caller would at least have access to the items if they needed them for some reason and the prop would not be completely irrelevant? Perhaps that would mean it is no longer necessary to pass the items to the slot. If we include total-count in the v-model then we would not need to pass anything in to the footer either.

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -1,33 +1,26 @@
...
+    itemsPerPageOptions?: number[];
+    itemsPerPage?: number;
+    autoRefreshSeconds: number;
+    showPagination?: boolean;
+    showItemsPerPage?: boolean;
+  }>(),
+  { itemsPerPageOptions: () => [20, 35, 50, 75], itemsPerPage: 50, showPagination: true, showItemsPerPage: false }
+);
+
+const refreshTimer = ref<number>();
```
mikeminutillo, line 22: Does this default to 0? Is it OK to call `clearInterval` on a 0?

mikeminutillo: Also, I wrote this as a `ref` but I'm wondering if it should be a `let`. We don't bind this to anything.

src/ServicePulse.Host/vue/src/components/ExclamationMark.vue
```diff
@@ -1,11 +1,16 @@
-﻿<script setup>
-const props = defineProps({
-  type: String,
-});
+﻿<script setup lang="ts">
+import { LicenseWarningLevel } from "@/composables/LicenseStatus";
+
+const props = withDefaults(
+  defineProps<{
+    type?: LicenseWarningLevel;
```
mikeminutillo, line 6: We definitely use this control in other contexts than only licenses. For instance, [we show it when we cannot connect to ServiceControl](https://github.com/Particular/ServicePulse/pull/1706/files#diff-bc13c6fff783e13d1577030cae7a3ae1c35d5e82b1fcab7ed639e3582bbf408cL23).

mikeminutillo (approved):

src/ServicePulse.Host/vue/src/components/DataView.vue
johnsimons: const is fine @mikeminutillo

johnsimons: it defaults to `undefined`

src/ServicePulse.Host/vue/src/components/ExclamationMark.vue
johnsimons: fixed

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: I meant should it not be a `ref` at all. i.e.

```js
let refreshTimer : number;
```

johnsimons: fixed

src/ServicePulse.Host/vue/src/components/DataView.vue
johnsimons: I am not 100% sure what you mean, @mikeminutillo .
Happy for you to refractor?

PR #1707, opened 2024-02-09
Remove d3 largegraph

PhilBastian:


src/ServicePulse.Host/vue/src/stores/FailedMessageStore.js
```diff
@@ -13,8 +13,8 @@ export const useFailedMessageStore = defineStore("FailedMessageStore", {
     },
   },
   getters: {
-    serviceControlId: (state) => state.failedMessagesList[0].id,
-    errorCount: (state) => state.failedMessagesList[0].count,
-    isFailedMessagesEmpty: (state) => state.errorCount === 0,
+    serviceControlId: (state) => state.failedMessagesList[0]?.id,
```
PhilBastian, line 16: I didn't investigate what was going on here, but this was throwing tonnes of errors into my console, so I made it capable of handling null values. There is something very wrong with the way this state is set/handled, since the UI flickers constantly.

src/ServicePulse.Host/vue/src/components/monitoring/LargeGraph.vue
```diff
@@ -0,0 +1,174 @@
+<script setup>
```
PhilBastian, line 1: it is worthwhile doing a diff between this file and D3LargeGraph rather than looking at it as a brand new file

src/ServicePulse.Host/vue/package.json
```diff
@@ -19,7 +19,6 @@
     "@vue-flow/core": "^1.29.1",
     "bootstrap": "^5.2.0",
     "bootstrap-icons": "^1.10.3",
-    "d3": "^7.6.1",
```
mikeminutillo, line 22: :tada:

src/ServicePulse.Host/vue/src/stores/FailedMessageStore.js
mikeminutillo: We fixed the flicker. Is this the same thing that was being fixed in https://github.com/Particular/ServicePulse/pull/1691 ?

src/ServicePulse.Host/vue/src/components/monitoring/graphLines.ts
```diff
@@ -12,19 +13,20 @@ export function useGraph(plotdata, minimumyaxis) {
...
     })();
     const xTick = 100 / (values.length - 1);
-    const coordinates = values.reduce((points, yValue, i) => [...points, [i * xTick, yValue]], []);
+    const coordinates = values.reduce((points: Coordinate[], yValue, i) => [...points, [i * xTick, yValue] as Coordinate], []);
     valuesPath.value = new Path().startAt(coordinates[0]).followCoordinates(coordinates.slice(1)).toString();
     valuesArea.value = new Path().startAt([0, 0]).followCoordinates(coordinates).lineTo([100, 0]).close().toString();
 
     average.value = plotData.average;
     //TODO: why is this called minimumYaxis when it's only used to determine the maxYaxis?
     // should the graph actually set the min y value rather than leave it at 0?
```
mikeminutillo, line 27: If all of your values are between 2 and 3, then the maxYaxis will be set to 10, so this is the _minimum_ that the max Y axis can be.

If all of your values are between 11 and 12, then the maxYaxis will be set to 12, which is above the minimum Y axis of 10.

src/ServicePulse.Host/vue/src/components/monitoring/LargeGraph.vue
```diff
@@ -0,0 +1,176 @@
...
+});
+const series1AverageLabelSuffix = computed(() => (props.isdurationgraph ? useFormatTime(series1.average.value.toString()).unit.toUpperCase() : props.metricsuffix ?? ""));
+const series2AverageLabelSuffix = computed(() => (props.isdurationgraph ? useFormatTime(series2.average.value.toString()).unit.toUpperCase() : props.metricsuffix ?? ""));
+//NOTE: using hard coded height of graph (200px - 10 for padding). To get it exact without hard coding a height value, we would need to perform measurement on the rendered SVG element, which we want to avoid
+const series1AverageLabelPosition = computed(() => `calc(${(series1.average.value / maxYaxis.value) * 190}px - 1em)`);
+const series2AverageLabelPosition = computed(() => `calc(${(series2.average.value / maxYaxis.value) * 190}px - 1em)`);
+</script>
+
+<template>
+  <div class="graph large-graph pull-left" :class="[hover ? 'hover' : '']" @mouseover="hover = true" @mouseout="hover = false">
```
mikeminutillo, line 75: You can use Vue-specific bindings for this. I am not sure if that optimizes anything but it looks like this:

```suggestion
  <div class="graph large-graph pull-left" :class="{ hover: hover }" @mouseover="hover = true" @mouseout="hover = false">
```

src/ServicePulse.Host/vue/src/components/monitoring/LargeGraph.vue
```diff
@@ -0,0 +1,176 @@
...
+    <div class="padding">
+      <svg class="y-axis">
+        <!-- 
+          Offsets explained:
+            - 60px inset from the left for the y-axis values to show
+            - 5px padding above and below to prevent the bottom and top y-axis value from being cut off by the edge of the SVG 
+              (font size 10px for class 'tick', offset half a character = 5px)
+            - transform 60px from the left and 5px down for the same reasons as above
+        -->
```
mikeminutillo, line 84: I was worried about these getting rendered in the final output, but they aren't :) 
[Vue strips them out of the production build](https://vuejs.org/api/application.html#app-config-compileroptions-comments).

src/ServicePulse.Host/vue/src/components/monitoring/LargeGraph.vue
```diff
@@ -0,0 +1,176 @@
...
+          </g>
+        </g>
+      </svg>
+      <svg class="data" :viewBox="`0 0 100 ${maxYaxis}`" preserveAspectRatio="none">
+        <g :class="props.firstseriestype">
+          <path :d="series1.valuesArea.value" class="graph-data-fill" />
+          <path :d="series1.valuesPath.value" vector-effect="non-scaling-stroke" class="graph-data-line" />
+          <path :d="series1.averageLine.value" vector-effect="non-scaling-stroke" class="graph-avg-line" />
+        </g>
+        <g :class="props.secondseriestype">
```
mikeminutillo, line 99: Should this have a `v-if` on it? It may not be set.

src/ServicePulse.Host/vue/src/components/monitoring/EndpointDetails.vue
```diff
@@ -466,18 +466,18 @@ onMounted(() => {
...
                   :isdurationgraph="true"
                   :firstdataseries="endpoint.metricDetails.metrics.criticalTime"
                   :seconddataseries="endpoint.metricDetails.metrics.processingTime"
                   :minimumyaxis="largeGraphsMinimumYAxis.processingCritical"
-                  :firstseriescolor="'#2700CB'"
-                  :firstseriesfillcolor="'#C4BCE5'"
-                  :secondseriescolor="'#258135'"
-                  :secondseriesfillcolor="'#BEE6C5'"
+                  :firstseriestype="'critical-time'"
+                  :firstseriesavgcolour="'#2700CB'"
```
mikeminutillo, line 476: Why pass in these colors and not make apply the type of graph as a class to the tooltip and style them with css?

mikeminutillo (approved):

src/ServicePulse.Host/vue/src/components/monitoring/EndpointDetails.vue
cquirosj: Good idea. We can do that in another PR, along with fixing using "color" instead of "colour" for consistency.

cquirosj (approved):

src/ServicePulse.Host/vue/src/components/monitoring/LargeGraph.vue
```diff
@@ -0,0 +1,176 @@
+<script setup lang="ts">
+import { ref, computed } from "vue";
+import { useFormatTime, useFormatLargeNumber } from "../../composables/formatter";
+import { useGraph } from "./graphLines";
+import type { PlotData } from "./PlotData";
+
+const props = defineProps<{
+  isdurationgraph: boolean;
+  metricsuffix?: string | undefined;
```
johnsimons, line 9: @PhilBastian, you don't actually need the ` | undefined`, because you have declared the prop as optional `?`.
So when you declare a prop as optional it will add automatically ` | undefined`.
But if you want to also support `null`, then you would include it, example:
```ts
 myVar?: string | null; // this will be `string | undefined | null`
```
I hope it makes sense?
BTW, I think the code base tends to use camelCase, in my IDE this code is highlighted as a spelling mistake.
image

PhilBastian: yep, was there is previous code so i didn't change it

PR #1714, opened 2024-02-13
Monitoring endpoint details: add Message Type pagination

mikeminutillo:
Adds pagination to message types on the monitoring endpoint details page

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/monitoring/EndpointDetails.vue
+++ b/src/ServicePulse.Host/vue/src/components/monitoring/EndpointDetails.vue
@@ -1,6 +1,6 @@
 ﻿<script setup>
 // Composables
-import { ref, watch, onMounted, onUnmounted } from "vue";
+import { computed, ref, watch, onMounted, onUnmounted } from "vue";
 import { useRoute, useRouter } from "vue-router";
 import { monitoringConnectionState, connectionState } from "../../composables/serviceServiceControl";
 import { useFormatTime, useFormatLargeNumber } from "../../composables/formatter";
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/vue/src/components/monitoring/EndpointDetails.vue
```diff
@@ -61,6 +62,18 @@ watch(monitoringStore.historyPeriod, (newValue) => {
...
 
+const paginatedMessageTypes = computed(() => {
+  return endpoint.value.messageTypes.slice((endpoint.value.messageTypesPage - 1) * endpoint.value.messageTypesItemsPerPage, endpoint.value.messageTypesPage * endpoint.value.messageTypesItemsPerPage);
+});
+
+watch(
+  () => endpoint.value.messageTypesPage,
+  () => {
+    const breakdownTabName = showInstancesBreakdown ? "instancesBreakdown" : "messageTypeBreakdown";
+    router.push({ name: "endpoint-details", params: { endpointName: endpointName }, query: { historyPeriod: historyPeriod.value.pVal, tab: breakdownTabName, pageNo: endpoint.value.messageTypesPage } });
```
PhilBastian, line 73: as per the cleanup branch, I'm not convinced we want every change of tab in browser history

PhilBastian (approved):

cquirosj (approved):

src/ServicePulse.Host/vue/src/components/monitoring/EndpointDetails.vue
mikeminutillo: Fixed this but the routing still does not behave with the history as expected. I may need to pair up with someone to figure this out, or it may be better to fix it after splitting other components out (to minimize amount of code in the page that is touching it).

cquirosj: Agreed, let's try to fix it after splitting other components out.

PR #1722, opened 2024-02-16
Deconstruct endpoint details

mikeminutillo:
Deconstructs the endpoint details screen into multiple components

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/monitoring/messageTypes.js
+++ b/src/ServicePulse.Host/vue/src/components/monitoring/messageTypes.js
@@ -1,0 +1,59 @@
+function shortenTypeName(typeName) {
+  return typeName.split(".").pop();
+}
+
+function parseTheMessageTypeData(messageType) {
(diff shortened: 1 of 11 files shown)


src/ServicePulse.Host/vue/src/components/monitoring/messageTypes.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  return {
+    ...messageType,
+    shortName: shortenTypeName(messageType.typeName),
+    tooltip,
+  };
+}
+
+export default class messageTypes {
```
johnsimons, line 50: Same casing as you would in c# 😄

cquirosj (approved):

src/ServicePulse.Host/vue/src/components/monitoring/messageTypes.js
johnsimons: So in this case `MessageTypes`

PR #1727, opened 2024-02-19
Presentation component tests samples (testing PaginationStrip)

cquirosj:
The goal of this PR is to showcase a more interesting set of tests for testing **presentation** components. 

The original PR: 
- https://github.com/Particular/ServicePulse/pull/1720

Included the testing setup for testing presentation components and only tested the very basic NoData.vue component. This PR introduces tests for the PaginationStrip

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/PaginationStrip.spec.ts
+++ b/src/ServicePulse.Host/vue/src/components/PaginationStrip.spec.ts
@@ -1,4 +1,4 @@
-import { expect, it, render, screen, describe } from "@component-test-utils";
+import { expect, it, render, screen, describe, userEvent } from "@component-test-utils";
 import paginationStrip from "./PaginationStrip.vue";
 
 describe("Previous page behavior", () => {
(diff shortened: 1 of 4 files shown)


src/ServicePulse.Host/vue/src/components/PaginationStrip.vue
```diff
@@ -115,7 +115,7 @@ const pages = computed(() => {
   <div v-if="showPagination" class="col align-self-center">
     <ul class="pagination justify-content-center">
       <li v-for="page of pages" class="page-item" :key="page.key">
-        <a class="page-link" @click="pageNumber = page.page" :class="page.class">{{ page.label }}</a>
+        <button :disabled="!!page.class?.disabled" :aria-label="page.label" class="page-link" @click="pageNumber = page.page" :class="page.class">{{ page.label }}</button>
```
cquirosj, line 118: @mikeminutillo, following Testing Library [Guiding Principle](https://testing-library.com/docs/guiding-principles/), which encourages writing tests that closely resemble how web pages are used (even for component tests). I changed this template to render a button instead of an anchor so that the "toBeDisabled()" assertion can be used and to avoid checking for the usage of a CSS class, which is an implementation detail. Do you see any problem with this? @PhilBastian I checked how the strip renders, and looks ok, although there is a perceptible difference.

src/ServicePulse.Host/vue/src/components/PaginationStrip.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+describe("Page buffer behavior", () => {
+  it.todo("Example 1");
+  it.todo("Example 2");
+});
+
+describe("Total count and items per page behavior", () => {
+  it.todo("Example 1");
+  it.todo("Example 2");
+});
```
cquirosj, line 45: @mikeminutillo could you suggest scenarios for testing the page bugger behavior? I can implement the tests.

mikeminutillo: The page buffer indicates how many pages to the left and right of the current page before showing an elipsis `...`.

- If the current page number is less than the page buffer there should be no left ellipsis
- If the current page number is greater than the page buffer there should be a left ellipsis
- If the current page is equal to the page buffer then should be no left ellipsis
Same for the right ellipsis, although that should be "current page is within page buffer of the last page.

src/ServicePulse.Host/vue/src/components/PaginationStrip.vue
mikeminutillo: That's fine. Phil and I discussed turning these into buttons. It is a systemic issue across the app. `<a>` tags are used without `href`. At the time we decided to leave it as is and do it across the board later.

PhilBastian: I have updated the styles to make this control look how it did in the old angular pages. Even before you changed it to be buttons it wasn't styling correctly

TravisNickels (approved):

PR #1728, opened 2024-02-19
Added typed links

johnsimons:
As part of the work I am doing for the throughput reporting, I need to add a new top nav link, and I noticed all our links are hard coded and not centralised.
So this is just a cleanup for better linking within the app.

So from a consumer perspective, you know take in `routerLinks`, and then you reference a link, eg:
```
<RouterLink :to="routeLinks.dashboard">
```

I have updated all hardcoded links to use this new way, with the exception of https://github.com/Particular/ServicePulse/blob/843554f03ee27cda1a5ae32140b53c4a5149a763/src/ServicePulse.Host/vue/src/composables/serviceLicense.ts#L10-L17 which should use templates!

### DIFF
--- a/src/ServicePulse.Host/vue/src/router/index.ts
+++ b/src/ServicePulse.Host/vue/src/router/index.ts
@@ -21,7 +21,7 @@ const routes = config.map((item) => {
 
 const router = createRouter({
   history: createWebHashHistory(window.defaultConfig.base_url),
-  routes: routes,
+  routes,
 });
 
 router.beforeEach((to, from, next) => {
--- a/src/ServicePulse.Host/vue/src/router/index.ts
+++ b/src/ServicePulse.Host/vue/src/router/index.ts
@@ -1,124 +1,31 @@
-import { createRouter, createWebHashHistory } from "vue-router";
-import DashboardView from "@/views/DashboardView.vue";
-import FailedMessagesView from "@/views/FailedMessagesView.vue";
-import EventsView from "@/views/EventsView.vue";
-import ConfigurationView from "@/views/ConfigurationView.vue";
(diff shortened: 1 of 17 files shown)


src/ServicePulse.Host/vue/src/router/index.ts
```diff
@@ -1,124 +1,31 @@
...
-        {
-          name: "endpoint-connection",
-          path: "endpoint-connection",
-          component: () => import("@/components/configuration/EndpointConnection.vue"),
-        },
-      ],
-    },
-  ],
-  strict: false,
+  routes: routes,
```
PhilBastian, line 24: can inline

src/ServicePulse.Host/vue/src/router/vue-router.d.ts
```diff
@@ -8,7 +8,6 @@ export {};
 
 declare module "vue-router" {
   interface RouteMeta {
-    // is optional
-    title?: string;
+    title: string;
```
PhilBastian, line 11: i haven't looked to see where RouteMeta is used, but would it make sense to have RouteItem extend RouteMeta?

PhilBastian (approved):

src/ServicePulse.Host/vue/src/router/vue-router.d.ts
johnsimons: I copied this from https://router.vuejs.org/guide/advanced/meta.html#TypeScript and didn't really look much into it.
This is just to make TS happy!

src/ServicePulse.Host/vue/src/router/index.ts
johnsimons: fixed

PR #1733, opened 2024-02-20
Refactor subIsActive to a common function

johnsimons:
I noticed we had this common functionality that was being reinvented on a bunch of screens, specifically screens with tabs.
This functionality selects the correct tab based on the current route selected.

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/PageHeader.vue
+++ b/src/ServicePulse.Host/vue/src/components/PageHeader.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { RouterLink, useRoute } from "vue-router";
+import { RouterLink } from "vue-router";
 import { computed } from "vue";
 import { connectionState, monitoringConnectionState, stats } from "../composables/serviceServiceControl";
 import { useIsMonitoringEnabled } from "../composables/serviceServiceControlUrls";
@@ -8,14 +8,7 @@ import ExclamationMark from "./ExclamationMark.vue";
 import { LicenseWarningLevel } from "@/composables/LicenseStatus";
 import { WarningLevel } from "@/components/WarningLevel";
 import routeLinks from "@/router/routeLinks";
-
-function subIsActive(input, exact) {
-  const paths = Array.isArray(input) ? input : [input];
-  const route = useRoute();
-  return paths.some((path) => {
-    return exact ? route.path.endsWith(path) : route.path.indexOf(path) === 0; // current path starts with this path string
-  });
-}
+import isRouteSelected from "@/composables/isRouteSelected";
 
 const displayWarn = computed(() => {
   return licenseStatus.warningLevel === LicenseWarningLevel.Warning;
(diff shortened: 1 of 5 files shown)


src/ServicePulse.Host/vue/src/components/PageHeader.vue
```diff
@@ -36,47 +29,47 @@ const displayDanger = computed(() => {
...
           </li>
-          <li :class="{ active: subIsActive('/a/#/endpoints') }">
+          <li>
             <a :href="routeLinks.heartbeats">
               <i class="fa fa-heartbeat icon-white" title="Heartbeats"></i>
               <span class="navbar-label">Heartbeats</span>
               <span v-if="stats.number_of_failed_heartbeats > 0" class="badge badge-important">{{ stats.number_of_failed_heartbeats }}</span>
             </a>
           </li>
-          <li v-if="useIsMonitoringEnabled()" :class="{ active: subIsActive('/a/#/monitoring') || subIsActive('/a/#/monitoring/endpoint') }">
```
johnsimons, line 52: I completely removed this code, given this would never be active in the vue side

src/ServicePulse.Host/vue/src/components/PageHeader.vue
```diff
@@ -36,47 +29,47 @@ const displayDanger = computed(() => {
...
           </li>
-          <li :class="{ active: subIsActive('/failed-messages') }">
+          <li :class="{ active: isRouteSelected(routeLinks.failedMessage.root) }">
             <RouterLink :to="routeLinks.failedMessage.root">
               <i class="fa fa-envelope icon-white" title="Failed Messages"></i>
               <span class="navbar-label">Failed Messages</span>
               <span v-if="stats.number_of_failed_messages > 0" class="badge badge-important">{{ stats.number_of_failed_messages }}</span>
             </RouterLink>
           </li>
-          <li :class="{ active: subIsActive('/a/#/custom-checks') }">
```
johnsimons, line 66: I completely removed this code, given this would never be active in the vue side

src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
```diff
@@ -0,0 +1,15 @@
+import { useLink, useRoute } from "vue-router";
+
+export default function isRouteSelected(path: string) {
+  const route = useRoute();
+
+  if (route.path === path) {
+    return true;
+  }
+
+  if (`${path}/` === route.path) {
```
johnsimons, line 10: This is to handle the scenarios when the child route path is an empty string ("").
Eg, `/configuration` vs `/configuration/`
See https://github.com/Particular/ServicePulse/blob/b27138fc4e28f82cacbf414dbc9d908c3b75663a/src/ServicePulse.Host/vue/src/router/routeLinks.ts#L26

src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
```diff
@@ -0,0 +1,15 @@
...
+    return true;
+  }
+
+  return useLink({ to: path }).route.value.name === route.name;
```
johnsimons, line 14: This one handles grandchildren routes, when we want to still select the correct tab, but the route is a child of the parent route.

src/ServicePulse.Host/vue/src/router/index.ts
```diff
@@ -1,22 +1,33 @@
...
   return { title: `${item.title} • ServicePulse` };
 }
 
-const routes = config.map((item) => {
-  return {
+const routes = config.flatMap<RouteRecordRaw>((item) => {
+  const result: RouteRecordRaw[] = [];
+  result.push({
     path: item.path,
+    name: item.path,
```
johnsimons, line 12: I added a name to the routes to support the lookup

src/ServicePulse.Host/vue/src/router/index.ts
```diff
@@ -1,22 +1,33 @@
...
     meta: meta(item),
-    alias: item.alias ?? [],
```
johnsimons, line 12: I found an issue with using aliases vs redirects. This was introduced in #1728.
The summary, is that redirects work better, so what I have done here is create redirects from aliases.

src/ServicePulse.Host/vue/src/components/PageHeader.vue
```diff
@@ -36,47 +29,47 @@ const displayDanger = computed(() => {
...
       <div id="navbar" class="navbar navbar-expand-lg">
         <ul class="nav navbar-nav navbar-inverse">
-          <li :class="{ active: subIsActive('/dashboard', true) }">
+          <li :class="{ active: isRouteSelected(routeLinks.dashboard) }">
             <RouterLink :to="routeLinks.dashboard">
               <i class="fa fa-dashboard icon-white" title="Dashboard"></i>
               <span class="navbar-label">Dashboard</span>
             </RouterLink>
           </li>
-          <li :class="{ active: subIsActive('/a/#/endpoints') }">
```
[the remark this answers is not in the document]
mikeminutillo: Keeping this here would allow a change in the `routeLinks.ts` when this screen is migrated to Vue

src/ServicePulse.Host/vue/src/components/PageHeader.vue
mikeminutillo: Same again

src/ServicePulse.Host/vue/src/components/PageHeader.vue
mikeminutillo: And again

src/ServicePulse.Host/vue/src/router/index.ts
mikeminutillo: Is this supposed to be the `.path` as well? This is replicating the `path` above.

mikeminutillo (approved):

PR #1736, opened 2024-02-21
Endpoint name group sorting

TravisNickels:
Adds sorting to the endpoint name column when the endpoint list is grouped by segments.  This also correctly displays the short name of the endpoint when the list is grouped.

### DIFF
--- a/src/ServicePulse.Host/vue/src/stores/MonitoringStore.js
+++ b/src/ServicePulse.Host/vue/src/stores/MonitoringStore.js
@@ -58,17 +58,14 @@ export const useMonitoringStore = defineStore("MonitoringStore", () => {
   async function updateEndpointList() {
     endpointList.value = await MonitoringEndpoints.useGetAllMonitoredEndpoints(historyPeriodStore.historyPeriod.pVal);
     if (!endpointListIsEmpty.value) {
-      sortEndpointList();
       updateGroupSegments();
-      if (endpointListIsGrouped.value) {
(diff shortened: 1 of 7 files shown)


src/ServicePulse.Host/vue/src/stores/MonitoringStore.js
```diff
@@ -105,6 +103,36 @@ export const useMonitoringStore = defineStore("MonitoringStore", () => {
...
+        // Sort each group's endpoints before sorting the group name
+        endpointListGroupA.sort(endpointShortNameComparator);
+        endpointListGroupB.sort(endpointShortNameComparator);
+
+        return isSortAscending.value ? groupNameA.localeCompare(groupNameB) : groupNameB.localeCompare(groupNameA);
+      };
+    }
+    // TODO: Determine how sorting should be handled for columns other than endpoint name
+
+    if (grouping.value.groupedEndpoints.length > 1) {
```
PhilBastian, line 129: since comparator isn't always set, need a check for comparator truthy here

PhilBastian (approved):

cquirosj (approved):

src/ServicePulse.Host/vue/src/stores/MonitoringStore.js
PhilBastian: grouping will always be set, but comparator is only set if "sortByProperty" is equal to "name"

PR #1740, opened 2024-02-21
Scope navbar styles

johnsimons:
Also fixed the issue with possible double up on exclamation-mark components


PhilBastian (changes requested):
There are other places that the navbar styles are used (monitoring definitely). The reference to navbar.css can be removed from main.css and imported into PageHeader.vue, but it needs to remain its own file

johnsimons:
> There are other places that the navbar styles are used (monitoring definitely).

Can you explain this, Why is monitoring using styles for the navbar?

PhilBastian:
> > There are other places that the navbar styles are used (monitoring definitely).
> 
> Can you explain this, Why is monitoring using styles for the navbar?

the historyperiodselector is using it.

johnsimons:
@PhilBastian I still don't understand why `historyperiodselector` uses navbar styles.

mikeminutillo:
Is this coincidental? If the navabar changed, would we also want the history period selector to change (and vice-versa)? If so, it may better to duplicate.

PhilBastian:
I think we would want them both to change. The styles it uses are for arranging, padding, underlining, hover styles
https://github.com/Particular/ServicePulse/blob/merge_monitoring_master/src/ServicePulse.Host/vue/src/components/monitoring/MonitoringHistoryPeriod.vue

johnsimons:
@PhilBastian @mikeminutillo, I may be assuming this wrong, but are we trying to reuse the same styles for these two components?
image

I find them quite different. Even if they share the same padding and spacing, those could be different per context. 

That said, I am happy to close this PR.

PhilBastian:
as it turns out, this change was essentially done in the monitoring branch (removing from main and importing into pageheader, but leaving it in the css file). 
This branch can be closed

PR #1741, opened 2024-02-22
Monitoring to typescript

PhilBastian:


src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
```diff
@@ -1,10 +1,10 @@
-﻿<script setup>
+﻿<script setup lang="ts">
 import { ref } from "vue";
 import SortableColumn from "../../components/SortableColumn.vue";
 import EndpointListRow from "./EndpointListRow.vue";
-import { useMonitoringStore } from "../../stores/MonitoringStore";
+import { type MonitoringStore, useMonitoringStore } from "@/stores/MonitoringStore";
 
-const monitoringStore = useMonitoringStore();
+const monitoringStore: MonitoringStore = useMonitoringStore();
```
johnsimons, line 7: That is weird, isn't the stored typed?

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
```diff
@@ -1,70 +1,65 @@
...
   <div class="table-first-col endpoint-name name-overview">
     <div class="box-header">
       <div class="no-side-padding lead righ-side-ellipsis endpoint-details-link">
-        <a @click="navigateToEndpointDetails($event, endpoint.name)" class="cursorpointer" v-tooltip :title="endpoint.name">
-          {{ isGrouped ? shortName : endpoint.name }}
+        <a @click="navigateToEndpointDetails(endpoint.name)" class="cursorpointer" v-tooltip :title="endpoint.name">
+          {{ shortName }}
         </a>
-        <span class="endpoint-count" v-if="endpoint.connectedCount || endpoint.disconnectedCount" v-tooltip :title="`Endpoint instance(s):` + endpoint.connectedCount || 0">({{ endpoint.connectedCount || 0 }})</span>
+        <span class="endpoint-count" v-if="endpoint.connectedCount || endpoint.disconnectedCount" v-tooltip :title="`Endpoint instance(s):${endpoint.connectedCount || 0}`">({{ endpoint.connectedCount || 0 }})</span>
```
johnsimons, line 62: I looked at https://github.com/Particular/ServicePulse/pull/1741/files#diff-5e11e94635a6c7dcf3c5e304785a3ae820793a350fa890e7e3c2b77927db7de7R10 and is not marked as nullable?

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
```diff
@@ -144,7 +139,7 @@ function formatGraphDecimal(input, deci) {
       <div class="no-side-padding">
         <SmallGraph :type="'critical-time'" :isdurationgraph="true" :plotdata="endpoint.metrics.criticalTime" :minimumyaxis="smallGraphsMinimumYAxis.criticalTime" :avglabelcolor="'#2700CB'" />
       </div>
-      <div class="no-side-padding sparkline-value" :class="{ negative: formatGraphDuration(endpoint.metrics.criticalTime).value < 0 }">
+      <div class="no-side-padding sparkline-value" :class="{ negative: parseFloat(formatGraphDuration(endpoint.metrics.criticalTime).value) < 0 }">
```
johnsimons, line 142: should this be computed?

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
```diff
@@ -1,70 +1,65 @@
...
-  } else {
-    input = {
-      points: [],
-      average: 0,
-      displayValue: 0,
-    };
-    return input;
+
+function formatGraphDecimal(input: EndpointValues, deci: number) {
+  input = input ?? {
```
johnsimons, line 42: `input` is not nullable?
Should it be?

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
```diff
@@ -1,70 +1,65 @@
...
-  if ($event.target.localName !== "button") {
-    router.push({ name: "endpoint-details", params: { endpointName: endpointName }, query: { historyPeriod: selectedPeriod.value.pVal } });
-  }
+function navigateToEndpointDetails(endpointName: string) {
+  router.push({ name: "endpoint-details", params: { endpointName: endpointName }, query: { historyPeriod: selectedPeriod.value.pVal } });
 }
 
-function formatGraphDuration(input) {
+function formatGraphDuration(input: EndpointValues) {
   if (input) {
```
johnsimons, line 34: `input` is not nullable?

src/ServicePulse.Host/vue/src/composables/serviceMonitoringEndpoints.ts
```diff
@@ -0,0 +1,163 @@
...
+ * @returns The details of the endpoint
+ */
+export function useGetEndpointDetails(endpointName: string, historyPeriod = 1) {
+  const data = ref<EndpointDetails | null>(null);
+  return {
+    data,
+    refresh: async () => {
+      if (!useIsMonitoringDisabled() && !monitoringConnectionState.unableToConnect) {
+        try {
+          const response = await useFetchFromMonitoring(`${`monitored-endpoints`}/${endpointName}?history=${historyPeriod}`);
```
johnsimons, line 80: I created a typed API for the main instance, have a look at https://github.com/Particular/ServicePulse/blob/41fb564ab1e17b28a288170f575be1a0910f4281/src/ServicePulse.Host/vue/src/composables/serviceServiceControlUrls.ts#L18, I think you could do the same here, thoughts?

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
```diff
@@ -0,0 +1,96 @@
...
+  connectedCount: number;
+}
+
+export interface EndpointValues {
+  points: number[];
+  average: number;
+}
+
+export interface EndpointValuesWithTime extends EndpointValues {
+  timeAxisValues: Date[];
```
johnsimons, line 19: I assume this does not come from the backend REST API? 
c# dates do not desecrialize into JS dates

johnsimons (approved):
@PhilBastian a few comments, but overall really good effort 👍

src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
cquirosj: That is correct. The store type is useful when passing the store to a function. When assigned to a const, like in a component, adding the type is not required.

src/ServicePulse.Host/vue/src/composables/formatter.ts
```diff
@@ -35,22 +35,22 @@ export function useGetDayDiffFromToday(value: string) {
   return Math.round(diff / 1000 / 60 / 60 / 24);
 }
 
-export function useFormatLargeNumber(value: string, decimals: number) {
+export function useFormatLargeNumber(value: number, decimals: number) {
   const suffixes = ["k", "M", "G", "T", "P", "E"];
 
   const num = Number(value);
```
johnsimons, line 41: This should not be required

src/ServicePulse.Host/vue/src/composables/formatter.ts
```diff
@@ -35,22 +35,22 @@ export function useGetDayDiffFromToday(value: string) {
...
 
   if (isNaN(num)) {
```
johnsimons, line 43: the type says is a number, do we need this?

src/ServicePulse.Host/vue/src/composables/serviceMonitoringEndpoints.ts
```diff
@@ -0,0 +1,163 @@
...
+export async function useGetAllMonitoredEndpoints(historyPeriod = 1) {
+  let endpoints: Endpoint[] = [];
+  if (!useIsMonitoringDisabled() && !monitoringConnectionState.unableToConnect) {
+    try {
+      const response = await useFetchFromMonitoring(`${`monitored-endpoints`}?history=${historyPeriod}`);
+      const data = response && (await response.json());
+      endpoints = data ?? [];
+      await addEndpointsFromScSubscription(endpoints);
+    } catch (error) {
+      console.error(error);
```
johnsimons, line 29: Consider not handling error here

src/ServicePulse.Host/vue/src/composables/serviceMonitoringEndpoints.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * @param historyPeriod - The history period value.  The default is (1)
+ * @returns A array of monitoring endpoint objects
+ */
+export async function useGetAllMonitoredEndpoints(historyPeriod = 1) {
+  let endpoints: Endpoint[] = [];
+  if (!useIsMonitoringDisabled() && !monitoringConnectionState.unableToConnect) {
+    try {
+      const response = await useFetchFromMonitoring(`${`monitored-endpoints`}?history=${historyPeriod}`);
```
johnsimons, line 24: Use a typed function instead

src/ServicePulse.Host/vue/src/composables/serviceMonitoringEndpoints.ts
```diff
@@ -0,0 +1,163 @@
...
+      const data = response && (await response.json());
+      endpoints = data ?? [];
+      await addEndpointsFromScSubscription(endpoints);
+    } catch (error) {
+      console.error(error);
+    }
+  } else {
+    await addEndpointsFromScSubscription(endpoints);
```
johnsimons, line 32: Consider using SC for abbreviation or the full word without abbreviation

src/ServicePulse.Host/vue/src/composables/serviceMonitoringEndpoints.ts
```diff
@@ -0,0 +1,163 @@
...
+  return {
+    data,
+    refresh: async () => {
+      if (!useIsMonitoringDisabled() && !monitoringConnectionState.unableToConnect) {
+        try {
+          const response = await useFetchFromMonitoring(`${`monitored-endpoints`}/${endpointName}?history=${historyPeriod}`);
+          const result: EndpointDetails = response && (await response.json());
+          data.value = result;
+        } catch (error) {
+          console.error(error);
```
johnsimons, line 84: consider not catching this if it makes sense

src/ServicePulse.Host/vue/src/composables/serviceMonitoringEndpoints.ts
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * @returns The count of disconnected endpoint
+ */
+export async function useGetDisconnectedEndpointCount() {
+  let disconnectedCount = 0;
+  try {
+    const response = await useFetchFromMonitoring(`${`monitored-endpoints`}/disconnected`);
+    return (response && ((await response.json()) as number)) ?? 0;
+  } catch (error) {
+    console.error(error);
```
johnsimons, line 100: consider not catching

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
```diff
@@ -0,0 +1,96 @@
...
+export interface EndpointMetrics {
+  [index: string]: EndpointValues;
+  queueLength: EndpointValues;
+  throughput: EndpointValues;
+  retries: EndpointValues;
+  processingTime: EndpointValuesWithTime;
+  criticalTime: EndpointValuesWithTime;
+}
+
+const defaultMetricData: EndpointValues = {
```
johnsimons, line 31: These defaults don't feel like belong here

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
```diff
@@ -0,0 +1,96 @@
...
+});
+
+export interface EndpointInstance {
+  name: string;
+  id: string;
+  isStale: boolean;
+  metrics: EndpointMetrics;
+}
+
+export interface ExtendedEndpointInstance extends EndpointInstance {
```
johnsimons, line 54: Consider using a better name for this interface?

src/ServicePulse.Host/vue/src/stores/FailedMessageStore.js
johnsimons, line 1: Still a js file

src/ServicePulse.Host/vue/src/stores/MonitoringStore.ts
```diff
@@ -85,33 +80,31 @@ export const useMonitoringStore = defineStore("MonitoringStore", () => {
...
-    } else {
-      comparator = (a, b) => {
-        const propertyA = getPropertyValue(a, `metrics.${sortByProperty}.average`);
-        const propertyB = getPropertyValue(b, `metrics.${sortByProperty}.average`);
-
-        return isSortAscending.value ? propertyA - propertyB : propertyB - propertyA;
-      };
-    }
+    const comparator = (() => {
+      if (sortByProperty === "name") {
```
johnsimons, line 84: This can actually be type-safe. The trick is to use [keyof operator](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html).
What it does is that it ensures the `sortByProperty` is a valid prop of `Endpoint`.
To do that, you need to change the type of the ref `const sortBy = ref<keyof Endpoint>("name");`
And the change this functions to be:
```ts
async function updateSort(newSortBy: keyof Endpoint = "name", newIsSortAscending = false) {
    sortBy.value = newSortBy;
    isSortAscending.value = newIsSortAscending;
    await updateEndpointList();
  }
``` 
Now, the consumer of `updateSort` will get an error if they try to pass a prop that does not exist.

All that said, I looked at the `EndpointList.vue` file, and it will be a bit tricky to get this to work properly.
Anyway, at least now you know, you could do it 😄

src/ServicePulse.Host/vue/src/components/monitoring/EndpointList.vue
PhilBastian: i figured this out yesterday but haven't made the change yet :)

src/ServicePulse.Host/vue/src/composables/formatter.ts
PhilBastian: NaN is a valid number

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
PhilBastian: i don't see any need to?

johnsimons: I don't know if there are any benefits in this case; maybe the question is, should `formatGraphDuration` be called on every redraw?

PhilBastian: oh i see what you were referring to now: not just this line, but this div, calls formatGraphDuration 3 times for the same value. Will extract

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
PhilBastian: removed || to 0

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
PhilBastian: removed this function in preference for the shared version

src/ServicePulse.Host/vue/src/composables/serviceMonitoringEndpoints.ts
PhilBastian: the general flow is that an error here shouldn't cause the site to fall over, just blank out the endpoint list until connection is re-established.

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
PhilBastian: where else? they're co-located with the exported function that uses them

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
PhilBastian: i went through this with Mike - Extended seems like the best name, since these are values added onto the data retrieved from SC; literally extending the existing data

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
johnsimons: The plan was only to have types that are "copies" from the REST API c# code types in the resources folder.
This does not seem to be a type from the backend, hence I think we can move these somewhere else.

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
johnsimons: To me, it seems this new type is to include data about the UI state, so I am wondering if it would be better to have a separate "state" type?

src/ServicePulse.Host/vue/src/resources/Endpoint.ts
PhilBastian: it's actually not used... I have kept the member in case it was meant to be used somewhere which has been lost in migration, but changed the type to string[]

mikeminutillo (approved):

PR #1742, opened 2024-02-23
Implement a simple re-usable tabs component

mikeminutillo:
Replaces the inline tabs with a simple re-usable tabs component

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/TabContainer.vue
+++ b/src/ServicePulse.Host/vue/src/components/TabContainer.vue
@@ -1,0 +1,35 @@
+<script setup>
+import { computed, inject, onBeforeMount, watch } from "vue";
+
+const addTab = inject("addTab");
+const activeTab = inject("activeTab");
+const updateTab = inject("updateTab");
+
+const props = defineProps({
+  id: {
+    type: String,
+    required: true,
+  },
+  header: {
+    type: String,
+    required: true,
+  },
+});
+
+watch(
+  () => Object.assign({}, props),
+  () => updateTab(props.id, props.header)
+);
+
+onBeforeMount(() => {
+  addTab(props.id, props.header);
+});
+
+const isActive = computed(() => activeTab.value === props.id);
+</script>
+
+<template>
+  <section v-if="isActive">
+    <slot />
+  </section>
+</template>
--- a/src/ServicePulse.Host/vue/src/components/TabsContainer.vue
+++ b/src/ServicePulse.Host/vue/src/components/TabsContainer.vue
@@ -1,0 +1,49 @@
+<script setup>
+import { computed, onMounted, provide, reactive } from "vue";
+import { useRoute, useRouter } from "vue-router";
+
+const route = useRoute();
+const router = useRouter();
+
+const state = reactive({
+  tabs: [],
+});
+
+const activeTab = computed({
+  get() {
+    return route?.query?.tab ?? state.tabs[0].id;
+  },
+  set(newValue) {
+    router.replace({ query: { ...route.query, tab: newValue } });
+  },
+});
+
+function selectTab(tabId) {
+  activeTab.value = tabId;
+}
+
+function addTab(tabId, header) {
+  state.tabs.push({ id: tabId, header });
+}
+
+function updateTab(tabId, header) {
+  state.tabs.find((t) => t.id === tabId).header = header;
+}
+
+provide("addTab", addTab);
+provide("updateTab", updateTab);
+provide("activeTab", activeTab);
+
+onMounted(() => {
+  selectTab(state.tabs[0].id);
+});
+</script>
+
+<template>
+  <div class="tabs">
+    <h5 v-for="tab in state.tabs" :key="tab.id" :class="{ active: tab.id === activeTab }">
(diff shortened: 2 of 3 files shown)


src/ServicePulse.Host/vue/src/components/TabContainer.vue
```diff
@@ -0,0 +1,35 @@
+<script setup>
```
cquirosj, line 1: Shouldn't this be typescript?

src/ServicePulse.Host/vue/src/components/TabsContainer.vue
```diff
@@ -0,0 +1,49 @@
+<script setup>
```
cquirosj, line 1:
```suggestion
<script setup lang="ts">
```

johnsimons (commented):
@mikeminutillo really good start, well done.
I have added a few inline comments, but the biggest one is whether you can consider making this navigatable as part of the routing.
What I mean by this is that if you look at the current usage of tabs, there is a mix.
Some of them are qualified URLs defined in the routing folder, [example](https://github.com/Particular/ServicePulse/blob/b27138fc4e28f82cacbf414dbc9d908c3b75663a/src/ServicePulse.Host/vue/src/views/FailedMessagesView.vue#L28C13-L58C18), while others are not, [example](https://github.com/Particular/ServicePulse/blob/b27138fc4e28f82cacbf414dbc9d908c3b75663a/src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue#L411-L439).

I actually like the ones that integrate with routing, because they are easier to navigate directly to.

I think in this PR, you are making them navigatable but using the querystring and hence you need the `tab.id`.
But now that we have all navigatable location in https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host/vue/src/router/routeLinks.ts, to me it would make sense to add these there too, so that in the future we could navigate directly to a specific tab from the app, and all the dev needs to do is refer to the `routeLinks.ts`. I hope this makes sense!

Happy to have a chat, and by the way, it would be nice to do this PR directly against master so I can start using this 😉 (I had plans to do this myself!)

src/ServicePulse.Host/vue/src/components/TabContainer.vue
```diff
@@ -0,0 +1,35 @@
...
+import { computed, inject, onBeforeMount, watch } from "vue";
+
+const addTab = inject("addTab");
+const activeTab = inject("activeTab");
```
PhilBastian, line 5: I don't like the concept of provide/inject for first level relations, but if we're going to then the injection key should be an exported symbol so that we can't have accidental key clashes

src/ServicePulse.Host/vue/src/components/TabContainer.vue
```diff
@@ -0,0 +1,35 @@
...
+    required: true,
+  },
+  header: {
+    type: String,
+    required: true,
+  },
+});
+
+watch(
+  () => Object.assign({}, props),
```
PhilBastian, line 20: it's not possible to just watch props.header?

src/ServicePulse.Host/vue/src/components/TabsContainer.vue
```diff
@@ -0,0 +1,49 @@
...
+const state = reactive({
+  tabs: [],
+});
+
+const activeTab = computed({
+  get() {
+    return route?.query?.tab ?? state.tabs[0].id;
+  },
+  set(newValue) {
+    router.replace({ query: { ...route.query, tab: newValue } });
```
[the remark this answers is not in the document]
johnsimons: I guess it is a preference thing. But it would be nice to make it consistent.
At the moment, we have a mix.

PhilBastian:
> @mikeminutillo really good start, well done. I have added a few inline comments, but the biggest one is whether you can consider making this navigatable as part of the routing. What I mean by this is that if you look at the current usage of tabs, there is a mix. Some of them are qualified URLs defined in the routing folder, [example](https://github.com/Particular/ServicePulse/blob/b27138fc4e28f82cacbf414dbc9d908c3b75663a/src/ServicePulse.Host/vue/src/views/FailedMessagesView.vue#L28C13-L58C18), while others are not, [example](https://github.com/Particular/ServicePulse/blob/b27138fc4e28f82cacbf414dbc9d908c3b75663a/src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue#L411-L439).
> 
> I actually like the ones that integrate with routing, because they are easier to navigate directly to.
> 
> I think in this PR, you are making them navigatable but using the querystring and hence you need the `tab.id`. But now that we have all navigatable location in https://github.com/Particular/ServicePulse/blob/master/src/ServicePulse.Host/vue/src/router/routeLinks.ts, to me it would make sense to add these there too, so that in the future we could navigate directly to a specific tab from the app, and all the dev needs to do is refer to the `routeLinks.ts`. I hope this makes sense!
> 
> Happy to have a chat, and by the way, it would be nice to do this PR directly against master so I can start using this 😉 (I had plans to do this myself!)

@johnsimons the monitoring task force has opted to not include this component in our scope. Feel free to use this branch for your own purposes and include it in the scope of your testing, otherwise it will be closed for a future componentisation TF to look at

PR #1747, opened 2024-02-27
Handle case where current page exceeds total page count

mikeminutillo:
This can happen if the number of items per page changes. i.e. If you have 2 pages of 25 items each, and you are on page 2, and then change the number of items per page to 75.


johnsimons (approved):
LGTM

PhilBastian (approved):

PR #1748, opened 2024-02-27
fix navigation items not showing as active for subroutes

PhilBastian:
a recent change to checking whether a navitem is active no longer worked for subroutes. This fix returns that functionality

### DIFF
--- a/src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
+++ b/src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
@@ -2,15 +2,7 @@ import { useLink, useRoute } from "vue-router";
 
 export default function isRouteSelected(path: string) {
   const route = useRoute();
-
-  if (route.path === path) {
-    return true;
-  }
-
-  if (`${path}/` === route.path) {
-    return true;
-  }
-
   const pathRoute = useLink({ to: path }).route.value;
+
   return route.matched.some((match) => match.name === pathRoute.name);
 }
--- a/src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
+++ b/src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
@@ -11,5 +11,6 @@ export default function isRouteSelected(path: string) {
     return true;
   }
 
-  return useLink({ to: path }).route.value.name === route.name;
+  const pathRoute = useLink({ to: path }).route.value;
+  return route.matched.some((match) => match.name === pathRoute.name);
 }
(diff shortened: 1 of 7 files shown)


johnsimons (changes requested):
I had a chat with @PhilBastian about these changes and explain where things fall apart if these are applied.
Phil is going to have a look and see what can be done to address this issue without introducing new bugs.

src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
```diff
@@ -11,5 +11,6 @@ export default function isRouteSelected(path: string) {
     return true;
   }
 
-  return useLink({ to: path }).route.value.name === route.name;
+  const pathRoute = useLink({ to: path }).route.value;
+  return route.matched.some((match) => match.name === pathRoute.name);
```
PhilBastian, line 15: i actually don't think the two checks above are required anymore

src/ServicePulse.Host/vue/src/router/config.ts
```diff
@@ -17,14 +18,15 @@ export interface RouteItem {
 const config: RouteItem[] = [
   {
     path: routeLinks.dashboard,
-    alias: ["/"],
+    alias: "/",
```
johnsimons, line 21: Should this one be a redirect too?

src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
johnsimons, line 15: So "matched" should contain the parent and the child route correct?
And if so, then do we still need the ors in for example https://github.com/Particular/ServicePulse/pull/1748/files#diff-c6d4b62c099bf184d1bfc0cbe5fa16dcad0c1039acaf718429ca4e8b3d0030b3R20

johnsimons (approved):
a few questions but overall looks good

src/ServicePulse.Host/vue/src/router/config.ts
PhilBastian: that would cause an infinite loop

src/ServicePulse.Host/vue/src/composables/isRouteSelected.ts
PhilBastian: they're different routes (one isn't an ancestor of the other)

johnsimons: I agree, it seems the "matched" contains the routes matched in order of parent -> child -> grandchild -> ...

PR #1753, opened 2024-02-28
Application tests setup

cquirosj:


src/ServicePulse.Host/vue/public/mockServiceWorker.js
```diff
@@ -0,0 +1,287 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker (2.2.2).
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ * - Please do NOT serve this file on production.
```
mikeminutillo, line 8: Does this need to be explicitly excluded from application packaging?

mikeminutillo (approved):

src/ServicePulse.Host/vue/public/mockServiceWorker.js
cquirosj: yes

PR #1757, opened 2024-02-29
Add default time unit

mikeminutillo:
Instead of 0, it should be 0 ms

Before:

image

After:

image

### DIFF
--- a/src/ServicePulse.Host/vue/src/composables/formatter.ts
+++ b/src/ServicePulse.Host/vue/src/composables/formatter.ts
@@ -11,7 +11,7 @@ export interface ValueWithUnit {
 }
 
 export function useFormatTime(value?: number): ValueWithUnit {
-  const time = { value: "0", unit: "" };
+  const time = { value: "0", unit: "ms" };
   if (value) {
     const duration = moment.duration(value);
 


src/ServicePulse.Host/vue/src/composables/formatter.ts
```diff
@@ -11,7 +11,7 @@ export interface ValueWithUnit {
 }
 
 export function useFormatTime(value?: number): ValueWithUnit {
-  const time = { value: "0", unit: "" };
+  const time = { value: "0", unit: "ms" };
```
cquirosj, line 14: I wonder why none of the if statements were working for this condition. https://github.com/Particular/ServicePulse/blob/772e86e803087e1fa30fa46577bddfdc77acc519/src/ServicePulse.Host/vue/src/composables/formatter.ts#L29-L31

PhilBastian: it isn't hitting this else statement because of the surrounding
``` cs
 if (value) {
   ...
```

PhilBastian (approved):

cquirosj (approved):

PR #1762, opened 2024-03-01
display connectedcount/total rather than just connectedcount

PhilBastian:
display connectedcount/total rather than just connectedcount and separate the display from the endpoint link

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
+++ b/src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
@@ -36,8 +36,10 @@ const criticalTimeGraphDuration = computed(() => formatGraphDuration(endpoint.va
         <RouterLink :to="routeLinks.monitoring.endpointDetails.link(endpoint.name, historyPeriod.pVal)" class="cursorpointer" v-tooltip :title="endpoint.name">
           {{ shortName }}
         </RouterLink>
-        <span class="endpoint-count" v-if="endpoint.connectedCount || endpoint.disconnectedCount" v-tooltip :title="`Endpoint instance(s):${endpoint.connectedCount}`">({{ endpoint.connectedCount }})</span>
       </div>
+      <span class="endpoint-count" v-if="endpoint.connectedCount || endpoint.disconnectedCount" v-tooltip title="Endpoint instance(s): (connected/total)">
+        ({{ endpoint.connectedCount }}/{{ endpoint.connectedCount + endpoint.disconnectedCount }})</span
+      >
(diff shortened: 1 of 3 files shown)


src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
```diff
@@ -36,8 +36,10 @@ const criticalTimeGraphDuration = computed(() => formatGraphDuration(endpoint.va
         <RouterLink :to="routeLinks.monitoring.endpointDetails.link(endpoint.name, historyPeriod.pVal)" class="cursorpointer" v-tooltip :title="endpoint.name">
           {{ shortName }}
         </RouterLink>
-        <span class="endpoint-count" v-if="endpoint.connectedCount || endpoint.disconnectedCount" v-tooltip :title="`Endpoint instance(s):${endpoint.connectedCount}`">({{ endpoint.connectedCount }})</span>
       </div>
+      <span class="endpoint-count" v-if="endpoint.connectedCount || endpoint.disconnectedCount" v-tooltip title="Endpoint instance(s): (connected/total)">
```
mikeminutillo, line 40: Does this mean if both are zero we will display nothing?

johnsimons:
After spending some time on this, to me, it feels like we have two features that maybe can be merged together.
So we have the "counter" and the "red icon".
image

To the counter, display the current "active" endpoints and the red icon displays whether all instances are "disconnected".

Which makes me wonder if it would be better to remove the icon and have something like:

1. `(10/10)`  render in green, with tooltip "All instance are online"
1. `(8/10)` render in warning color, with tooltip "Some instances are offline"
1. `(0/10)` render in red with tooltip "All instance are offline"

src/ServicePulse.Host/vue/src/components/monitoring/EndpointListRow.vue
PhilBastian: i believe in that scenario we wouldn't show the row at all

PhilBastian:
> After spending some time on this, to me, it feels like we have two features that maybe can be merged together. So we have the "counter" and the "red icon". 
> 
> To the counter, display the current "active" endpoints and the red icon displays whether all instances are "disconnected".
> 
> Which makes me wonder if it would be better to remove the icon and have something like:
> 
> 1. `(10/10)`  render in green, with tooltip "All instance are online"
> 2. `(8/10)` render in warning color, with tooltip "Some instances are offline"
> 3. `(0/10)` render in red with tooltip "All instance are offline"

I would suggest instead that we raise this as a new issue for later attention

mikeminutillo (approved):

PR #1763, opened 2024-03-01
Make Endpoint list default direction for name sorting ascending

mikeminutillo:
The default sort order for the endpoint list was descending. This change makes the default sort for the column ascending.

I also created a `SortInfo` interface and passed that around to bind to instead of having two separate states to pass around.


johnsimons:
@mikeminutillo I noticed you playing with the sorting, so I thought I mentioned it.
Have you considered passing a [comparer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#comparefn) as a parameter to it, so that you may make the "property name" typed. Because at the moment is a magic string!

PhilBastian (approved):

PR #1764, opened 2024-03-01
Fixing Vue to Angular links in the Dashboard and the Header views

tmasternak:


mauroservienti (approved):

johnsimons (commented):
@tmasternak I ended up creating an [alternative](https://github.com/Particular/ServicePulse/pull/1765) to this PR, and I have added comments to explain the reasons

src/ServicePulse.Host/vue/src/components/PageHeader.vue
```diff
@@ -22,7 +22,7 @@ const displayDanger = computed(() => {
   <nav class="navbar navbar-expand-lg navbar-inverse navbar-dark">
     <div class="container-fluid">
       <div class="navbar-header">
-        <a class="navbar-brand" href="/">
+        <a class="navbar-brand" :href="baseUrl + routeLinks.dashboard">
```
johnsimons, line 25: We want the responsibility of creating links to be in `routeLinks.ts`, if we need to fully qualify a url, you can use the `useLink` component.

src/ServicePulse.Host/vue/src/router/routeLinks.ts
```diff
@@ -38,7 +38,7 @@ const monitoringLinks = (root: string) => {
   };
 };
 
-const baseUrl = window.defaultConfig.base_url;
+export const baseUrl = window.defaultConfig.base_url;
```
johnsimons, line 41: We want the responsibility of creating links to be in `routeLinks.ts`, hence there shouldn't be a need to export this

src/ServicePulse.Host/vue/src/components/PageHeader.vue
johnsimons: `routeLinks.ts` is where all internal links are defined.
However, because we are still in a mix of Angular and Vue, we also define Angular links in there for easier of maintenance.

johnsimons:
Fixed in https://github.com/Particular/ServicePulse/pull/1765

PR #1765, opened 2024-03-03
Fixing Vue to Angular links in the Dashboard and the Header views

johnsimons:
Alternative to https://github.com/Particular/ServicePulse/pull/1764

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/PageHeader.vue
+++ b/src/ServicePulse.Host/vue/src/components/PageHeader.vue
@@ -22,9 +22,9 @@ const displayDanger = computed(() => {
   <nav class="navbar navbar-expand-lg navbar-inverse navbar-dark">
     <div class="container-fluid">
       <div class="navbar-header">
-        <a class="navbar-brand" href="/">
+        <RouterLink class="navbar-brand" :to="routeLinks.dashboard">
           <img alt="Service Pulse" src="@/assets/logo.svg" />
-        </a>
+        </RouterLink>
       </div>
 
       <div id="navbar" class="navbar navbar-expand-lg">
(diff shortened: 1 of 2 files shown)


src/ServicePulse.Host/vue/src/components/PageHeader.vue
```diff
@@ -22,9 +22,9 @@ const displayDanger = computed(() => {
   <nav class="navbar navbar-expand-lg navbar-inverse navbar-dark">
     <div class="container-fluid">
       <div class="navbar-header">
-        <a class="navbar-brand" href="/">
+        <RouterLink class="navbar-brand" :to="routeLinks.dashboard">
```
johnsimons, line 25: For internal links we need to start using `RouterLink` component

src/ServicePulse.Host/vue/src/views/DashboardView.vue
```diff
@@ -27,13 +29,13 @@ import { licenseStatus } from "@/composables/serviceLicense";
               <div class="col-12">
                 <div class="row">
                   <div class="col-4">
-                    <DashboardItem :counter="stats.number_of_failed_heartbeats" :url="'/a/#/endpoints'" :iconClass="'fa-heartbeat'">Heartbeats</DashboardItem>
+                    <DashboardItem :counter="stats.number_of_failed_heartbeats" :url="routeLinks.heartbeats" :iconClass="'fa-heartbeat'">Heartbeats</DashboardItem>
                   </div>
                   <div class="col-4">
-                    <DashboardItem :counter="stats.number_of_failed_messages" :url="'#/failed-messages'" :iconClass="'fa-envelope'">Failed Messages</DashboardItem>
+                    <DashboardItem :counter="stats.number_of_failed_messages" :url="useLink({ to: routeLinks.failedMessage.root }).href.value" :iconClass="'fa-envelope'">Failed Messages</DashboardItem>
```
johnsimons, line 35: This one is a bit more code than what we would expect!
This is because at the moment `DashboardItem` uses use a `a href` link instead of `RouterLink`.
Once we migrate all pages ut of Angular, we can change `DashboardItem` to use `RouterLink` and hence no need to use the `useLink` to give us a fully qualified url.

PhilBastian (approved):

PR #1769, opened 2024-03-05
Failed messages to typescript

cquirosj:
- [x] set AllowJs to false in tsconfig.app.json


src/ServicePulse.Host/vue/tsconfig.app.json
```diff
@@ -3,6 +3,7 @@
   "include": ["env.d.ts", "src/**/*.js", "src/**/*.ts", "src/**/*.vue", "./test/**/*.ts"],
   "exclude": ["node_modules"],
   "compilerOptions": {
+    "moduleResolution": "Node",
```
johnsimons, line 6: @cquirosj can you explain why we need to change this at all?

src/ServicePulse.Host/vue/tsconfig.node.json
```diff
@@ -8,7 +8,8 @@
     "composite": true,
     "noEmit": true,
     "module": "ESNext",
-    "moduleResolution": "Bundler",
-    "types": ["node"]
+    "moduleResolution": "Node",
```
johnsimons, line 11: Same, @cquirosj can you explain why we need to change this at all?

src/ServicePulse.Host/vue/vite.config.ts
```diff
@@ -1,9 +1,12 @@
-import { fileURLToPath, URL } from "node:url";
+import { fileURLToPath } from "node:url";
 import { defineConfig } from "vite";
 import vue from "@vitejs/plugin-vue";
 import path from "path";
 import checker from "vite-plugin-checker";
 
+const __filename = fileURLToPath(import.meta.url);
```
johnsimons, line 7: Why this change, what is this fixing?

johnsimons: These are not used anywhere in this file?

src/ServicePulse.Host/vue/vite.config.ts
```diff
@@ -69,7 +72,7 @@ export default defineConfig({
   },
   server: {
     headers: {
-      "Content-Security-Policy": createCSPOverrides(port, defaultUrls),
+      "content-security-policy": createCSPOverrides(port, defaultUrls),
```
johnsimons, line 75: Headers are case insensitive, why this change?

src/ServicePulse.Host/vue/vitest.config.ts
```diff
@@ -1,6 +1,11 @@
 import { defineConfig } from "vitest/config";
 import viteConfig from "./vite.config.js";
 import path from "path";
+import { fileURLToPath } from "node:url";
+
+const __filename = fileURLToPath(import.meta.url);
```
johnsimons, line 6: These are not used anywhere in the file?

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -57,18 +61,13 @@ function loadMessages() {
   return loadPagedMessages(groupId.value, pageNumber.value, "", "", startDate.toISOString(), endDate.toISOString());
 }
 
-async function loadGroupDetails(groupId) {
+async function loadGroupDetails(groupId: string | string[]) {
```
johnsimons, line 64: This function should not be taking a `string[]`, because we use that for string concatenation, which will produce something like `string[]`

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -57,18 +61,13 @@ function loadMessages() {
...
   groupName.value = data.title;
 }
 
-function loadPagedMessages(groupId, page, sortBy, direction, startDate, endDate) {
-  if (typeof sortBy === "undefined") sortBy = "modified";
-  if (typeof direction === "undefined") direction = "desc";
-  if (typeof page === "undefined") page = 1;
-  if (typeof startDate === "undefined") startDate = new Date(0).toISOString();
-  if (typeof endDate === "undefined") endDate = new Date().toISOString();
+function loadPagedMessages(groupId?: string | string[], page: number = 1, sortBy: string = "modified", direction: string = "desc", startDate: string = new Date(0).toISOString(), endDate: string = new Date().toISOString()) {
```
johnsimons, line 70: `groupId?: string | string[]` feels wrong to me given we are doing string concatenation.
`sortBy: string = "modified"` should these be keyed to the allowed types, I can't actually see a `modified` prop in `FailedMessageViewWithExtendedUIProperties`. I would recommend using types for these, eg `sortBy: keyof FailedMessageViewWithExtendedUIProperties = "last_modified",`, but I would checkin ServiceControl to see what sort columns are allowed.
Same for `direction`, this should be an enum.

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -12,22 +12,26 @@ import MessageList from "./MessageList.vue";
...
+const configuration: Ref<Configuration | null> = ref(null);
 
 const route = useRoute();
-const groupId = ref(route.params.groupId);
+const groupId: Ref<string | string[]> = ref(route.params.groupId);
 const groupName = ref("");
 const pageNumber = ref(1);
 const totalCount = ref(0);
 const cookies = useCookies().cookies;
 const selectedPeriod = ref("Deleted in the last 7 days");
```
johnsimons, line 31: The value options for this ref, should be an enum or a specific set of strings, eg `type Period = "All Deleted" | "Deleted in the last 2 Hours" | "Deleted in the last 1 Day" | "Deleted in the last 7 days";`

johnsimons: It looks like you already have an array with allowed values, so maybe you could use that.

johnsimons (changes requested):
A few comments so far

src/ServicePulse.Host/vue/vite.config.ts
cquirosj: the [key in the dictionary is lowercase](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f5925b6c95f53653f35ce3f83fcda1ba9b1e04a6/types/node/http.d.ts#L142). It was being highlighted in vscode

image

src/ServicePulse.Host/vue/vitest.config.ts
cquirosj: **__filename** it is used in line 7: const __dirname = path.dirname(**__filename**);
__dirname is used later in line 18
 replacement: path.resolve(**__dirname**, './src'),

src/ServicePulse.Host/vue/tsconfig.node.json
cquirosj: Had to pick one, Bundler was not a valid option

image

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
cquirosj: fixed

src/ServicePulse.Host/vue/tsconfig.app.json
cquirosj: https://github.com/Particular/ServicePulse/pull/1769/files#r1515342922

src/ServicePulse.Host/vue/vitest.config.ts
cquirosj: Running `vue-tsc --build` results in error if this was not addressed

image

src/ServicePulse.Host/vue/src/resources/FailedMessageView.ts
```diff
@@ -1,4 +1,5 @@
 import type EndpointDetails from "@/resources/EndpointDetails";
+import type Header from "@/resources/Header";
 
 export default interface FailedMessageView {
```
PhilBastian, line 4: I think we should rename this, it's not a view but rather a model

src/ServicePulse.Host/vue/src/resources/FailedMessageView.ts
```diff
@@ -18,6 +19,32 @@ export default interface FailedMessageView {
   edit_of: string;
 }
 
+export interface FailedMessageViewWithExtendedUIProperties extends FailedMessageView {
```
PhilBastian, line 22: ExtendedFailedMessage

src/ServicePulse.Host/vue/src/resources/Message.ts
```diff
@@ -0,0 +1,29 @@
...
+  is_system_message: boolean;
+  conversation_id: string;
+  headers: Header[];
+  status: string;
+  message_intent: string;
+  body_url: string;
+  body_size: number;
+  instance_id: string;
+}
+export interface MessageWithExtendedUIProperties extends Message {
```
PhilBastian, line 24: ExtendedMessage

src/ServicePulse.Host/vue/src/components/failedmessages/MessageList.vue
```diff
@@ -113,9 +108,11 @@ defineExpose({
                 <span class="metadata"><i class="fa fa-laptop"></i> Machine: {{ message.receiving_endpoint.host }}</span>
                 <span class="metadata" v-if="message.redirect"><i class="fa pa-redirect-source pa-redirect-small"></i> Redirect: {{ message.redirect }}</span>
                 <!-- for deleted messages-->
-                <span class="metadata" v-if="message.status == 'archived'"><i class="fa fa-clock-o"></i> Deleted: <time-since :date-utc="message.last_modified"></time-since></span>
-                <span class="metadata danger" v-if="message.status == 'archived' && message.delete_soon"><i class="fa fa-trash-o danger"></i> Scheduled for deletion: immediately</span>
-                <span class="metadata danger" v-if="message.status == 'archived' && !message.delete_soon"><i class="fa fa-trash-o danger"></i> Scheduled for deletion: <time-since class="danger" :date-utc="message.deleted_in"></time-since> </span>
+                <span class="metadata" v-if="message.status === FailedMessageStatus.Archived"><i class="fa fa-clock-o"></i> Deleted: <time-since :date-utc="message.last_modified"></time-since></span>
```
PhilBastian, line 111: NOTE: previously this was comparing to 'archived', and now will be comparing to 'Archived'. I'm thinking this potentially worked because it compared with == rather than ===. The FailedMessageStatus type should match what's returned from ServiceControl

src/ServicePulse.Host/vue/src/components/failedmessages/MessageList.vue
```diff
@@ -36,7 +37,8 @@ function numberDisplayed() {
   return props.messages.length;
 }
 
-function labelClicked($event, index) {
+function labelClicked($event: MouseEvent, index: number) {
+  //TODO: this functionality isn't consistent on including start/end items
```
PhilBastian, line 41: address this TODO as part of this PR or raise a separate issue?

src/ServicePulse.Host/vue/src/components/failedmessages/MessageList.vue
```diff
@@ -47,25 +49,18 @@ function labelClicked($event, index) {
       messages[x].selected = !messages[x].selected;
     }
 
-    clearSelection();
+    //clear selection
+    const selection = document.getSelection();
```
PhilBastian, line 53: This looks different to the old clearSelection because document.selection is not an official implementation (not recognised as being a valid property of document) and window.getSelection and document.getSelection are [equivalent](https://developer.mozilla.org/en-US/docs/Web/API/Document/getSelection), so only one needs to be checked. Likewise, empty() and removeAllRanges() are synonyms

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -32,29 +37,29 @@ async function loadFailedMessage() {
   try {
     const response = await useFetchFromServiceControl("errors/last/" + id.value);
     if (response.status === 404) {
-      failedMessage.value = { notFound: true };
+      failedMessage.value = { notFound: true } as FailedMessageViewWithExtendedUIProperties;
     } else if (response.status !== 200) {
-      failedMessage.value = { error: true };
+      failedMessage.value = { error: true } as FailedMessageViewWithExtendedUIProperties;
     }
```
PhilBastian, line 43: - should be checking response.ok rather than status 200
- should this be returning after this line, rather than continuing to try and read response.json which is guaranteed to throw an exception?
- should the error/notFound values be on a different type, rather than typecasting an object that doesn't match the interface to FailedMessageViewWithExtendedetc? I realise this would then require 'is' checks where these values are read, but that would definitely be clearer and potentially cleaner

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -32,29 +37,29 @@ async function loadFailedMessage() {
...
     const data = await response.json();
     const message = data;
-    message.archived = message.status === "archived";
-    message.resolved = message.status === "resolved";
-    message.retried = message.status === "retryIssued";
-    message.error_retention_period = moment.duration(configuration.value.data_retention.error_retention_period).asHours();
-    message.isEditAndRetryEnabled = configuration.value.edit.enabled;
+    message.archived = message.status === FailedMessageStatus.Archived;
```
PhilBastian, line 46: this needs to be checked if it's working - does the SC call return 'archived' or 'Archived'?

src/ServicePulse.Host/vue/vite.config.ts
johnsimons: I guess we can make it lowercase, but not required from the http specs

src/ServicePulse.Host/vue/tsconfig.node.json
johnsimons: I don't know what tool you are using, but bundler is valid, see https://www.typescriptlang.org/tsconfig#moduleResolution.
You may need to update your tooling.
Also if you look at the error message, it actually says it!
image

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -12,22 +12,26 @@ import MessageList from "./MessageList.vue";
...
+import { FailedMessageViewWithExtendedUIProperties } from "@/resources/FailedMessageView";
+import Configuration from "@/resources/Configuration";
+import { TYPE } from "vue-toastification";
 
 let pollingFaster = false;
-let refreshInterval = undefined;
+let refreshInterval: ReturnType<typeof setInterval>;
 const perPage = 50;
-const configuration = ref([]);
+const configuration: Ref<Configuration | null> = ref(null);
```
PhilBastian, line 23: this works without importing Ref
```suggestion
const configuration = ref<Configuration | null>(null);
```

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -57,18 +61,13 @@ function loadMessages() {
   return loadPagedMessages(groupId.value, pageNumber.value, "", "", startDate.toISOString(), endDate.toISOString());
 }
 
-async function loadGroupDetails(groupId) {
+async function loadGroupDetails(groupId: string) {
   const response = await useFetchFromServiceControl(`archive/groups/id/${groupId}`);
```
PhilBastian, line 65: useTypedFetchFromServiceControl? Although mostly throwaway, it would enable us to specify what we expect to be returned from this endpoint, and will also enable consistent error handling

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -115,10 +115,10 @@ function loadPagedMessages(groupId, page, sortBy, direction, startDate, endDate)
   return loadDelMessagesPromise;
 }
 
-function updateMessagesScheduledDeletionDate(messages) {
+function updateMessagesScheduledDeletionDate(messages: FailedMessageViewWithExtendedUIProperties[]) {
   //check deletion time
-  messages.forEach((message) => {
-    message.error_retention_period = moment.duration(configuration.value.data_retention.error_retention_period).asHours();
+  messages.forEach((message: FailedMessageViewWithExtendedUIProperties) => {
```
PhilBastian, line 120: explicit type shouldn't be required here, since messages is a known type

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -144,9 +144,9 @@ function isAnythingSelected() {
 
 async function restoreSelectedMessages() {
   changeRefreshInterval(1000);
-  const selectedMessages = messageList.value.getSelectedMessages();
+  const selectedMessages = messageList.value.getSelectedMessages() as FailedMessageViewWithExtendedUIProperties[]; //TODO: remove this cast once messageList has been converted to typescript
   selectedMessages.forEach((m) => (m.restoreInProgress = true));
-  useShowToast("info", "Info", "restoring " + selectedMessages.length + " messages...");
+  useShowToast(TYPE.INFO, "Info", "restoring " + selectedMessages.length + " messages...");
```
PhilBastian, line 149: use string interpolation

src/ServicePulse.Host/vue/src/components/failedmessages/LastTenOperations.vue
```diff
@@ -1,17 +1,17 @@
...
+import type { Ref } from "vue";
 import NoData from "../NoData.vue";
-import { useFetchFromServiceControl } from "../../composables/serviceServiceControlUrls";
+import { useTypedFetchFromServiceControl } from "../../composables/serviceServiceControlUrls";
 import TimeSince from "../TimeSince.vue";
+import type HistoricRetryOperation from "@/resources/HistoricRetryOperation";
+import RecoverabilityHistoryResponse from "@/resources/RecoverabilityHistoryResponse";
 
-const historicOperations = ref([]);
+const historicOperations: Ref<HistoricRetryOperation[]> = ref([]);
```
PhilBastian, line 10:
```suggestion
const historicOperations = ref<HistoricRetryOperation[]>([]);
```

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -12,13 +12,18 @@ import ConfirmDialog from "../ConfirmDialog.vue";
...
+import Configuration, { EditAndRetryConfig } from "@/resources/Configuration";
+import { TYPE } from "vue-toastification";
+import { FailedMessageStatus, FailedMessageViewWithExtendedUIProperties } from "@/resources/FailedMessageView";
+import Message from "@/resources/Message";
 
-let refreshInterval = undefined;
+let refreshInterval: ReturnType<typeof setInterval>;
 let pollingFaster = false;
-const panel = ref();
+const panel: Ref<number> = ref(1);
```
PhilBastian, line 22:
```suggestion
const panel = ref<number>(1);
```

PhilBastian: and below

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -71,51 +76,64 @@ async function getConfiguration() {
 }
 
 async function getEditAndRetryConfig() {
-  const response = await useFetchFromServiceControl("edit/config");
-  const data = await response.json();
-  configuration.value.edit = data;
+  const [, data] = await useTypedFetchFromServiceControl<EditAndRetryConfig>("edit/config");
+
+  editAndRetryConfiguration.value = data;
   return;
```
PhilBastian, line 82: not required

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -71,51 +76,64 @@ async function getConfiguration() {
...
 }
 
 function updateMessageDeleteDate() {
-  const countdown = moment(failedMessage.value.last_modified).add(failedMessage.value.error_retention_period, "hours");
-  failedMessage.value.delete_soon = countdown < moment();
-  failedMessage.value.deleted_in = countdown.format();
+  const countdown = moment(failedMessage.value?.last_modified).add(failedMessage.value?.error_retention_period, "hours");
```
PhilBastian, line 86: this line should be inside the if

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -71,51 +76,64 @@ async function getConfiguration() {
...
+  if (failedMessage.value) {
+    failedMessage.value.archiving = true;
+  }
 }
 
 async function unarchiveMessage() {
   changeRefreshInterval(1000); // We've started an unarchive, so increase the polling frequency
-  await useUnarchiveMessage([id.value]);
-  failedMessage.value.restoring = true;
+  await useUnarchiveMessage([id.value as string]);
```
PhilBastian, line 104: can id be made a string ref so that the casts aren't required at all usage points?

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -125,51 +143,65 @@ async function downloadHeadersAndBody() {
 }
 
 async function downloadBody() {
-  const response = await useFetchFromServiceControl("messages/" + failedMessage.value.message_id + "/body");
+  const response = await useFetchFromServiceControl("messages/" + failedMessage.value?.message_id + "/body");
```
PhilBastian, line 146: most things in this function require failedMessage.value != null, so can put this guard at the top and then remove all the null checks below

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -125,51 +143,65 @@ async function downloadHeadersAndBody() {
...
   if (response.status === 404) {
-    failedMessage.value.messageBodyNotFound = true;
+    if (failedMessage.value) {
+      failedMessage.value.messageBodyNotFound = true;
+    }
+
     return;
   }
 
   if (response.headers.get("content-type") === "application/json") {
```
PhilBastian, line 155: feels like these should be in a switch

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -125,51 +143,65 @@ async function downloadHeadersAndBody() {
...
     }
     return;
   }
 
   try {
-    failedMessage.value.messageBody = await response.text();
+    if (failedMessage.value) {
+      failedMessage.value.messageBody = await response.text();
+    }
   } catch {
```
PhilBastian, line 188: can be a single catch around all three possibilities

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -125,51 +143,65 @@ async function downloadHeadersAndBody() {
...
+function formatXml(xml: string) {
+  function createShiftArr(step: string) {
     let space = "";
     if (isNaN(parseInt(step))) {
       // argument is string
       space = step;
     } else {
       // argument is integer
-      for (let i = 0; i < step; i++) {
+      for (let i = 0; i < parseInt(step); i++) {
```
PhilBastian, line 204: space = ' '.repeat(parseInt(step))

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -244,45 +281,47 @@ function formatXml(xml) {
...
 
-function formatJson(json) {
+function formatJson(json: string) {
   return JSON.stringify(json, null, 2);
 }
 
-function togglePanel(panelNum) {
-  if (!failedMessage.value.notFound && !failedMessage.value.error) {
+function togglePanel(panelNum: number) {
+  if (!failedMessage.value?.notFound && !failedMessage.value?.error) {
```
PhilBastian, line 289: could this cause incorrect behaviour is failedMessage is null? Perhaps if (failedMessage.value && !failedMessage.value.notFound && !failedMessage.value.error)

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -244,45 +281,47 @@ function formatXml(xml) {
...
-    serviceControlUrl = serviceControlUrl.replace("https://", "");
+  if (url.indexOf("https") === 0) {
+    url = url.replace("https://", "");
   } else {
-    serviceControlUrl = serviceControlUrl.replace("http://", "");
+    url = url.replace("http://", "");
   }
 
-  window.open("si://" + serviceControlUrl + "?search=" + messageId + "&endpointname=" + endpointName);
+  window.open("si://" + url + "?search=" + messageId + "&endpointname=" + endpointName);
```
PhilBastian, line 306: string interpolation

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -364,29 +403,29 @@ onUnmounted(() => {
...
-                <span v-if="failedMessage.restoring" title="Message is being retried" class="label sidebar-label label-info metadata-label">Restoring...</span>
-                <span v-if="failedMessage.archiving" title="Message is being deleted" class="label sidebar-label label-info metadata-label">Deleting...</span>
-                <span v-if="failedMessage.archived" title="Message is being deleted" class="label sidebar-label label-warning metadata-label">Deleted</span>
-                <span v-if="failedMessage.resolved" title="Message was processed successfully" class="label sidebar-label label-warning metadata-label">Processed</span>
-                <span v-if="failedMessage.number_of_processing_attempts > 1" :title="'This message has already failed ' + failedMessage.number_of_processing_attempts + ' times'" class="label sidebar-label label-important metadata-label"
-                  >{{ failedMessage.number_of_processing_attempts }} Retry Failures</span
+                <span v-if="failedMessage?.retried" title="Message is being retried" class="label sidebar-label label-info metadata-label">Retried</span>
```
PhilBastian, line 413: i wonder if typescript's vue knowledge is enough to be able to put a v-if="failedMessage" on the surrounding div and then not have to null check all the usages of failedMessage

src/ServicePulse.Host/vue/vitest.config.ts
cquirosj: Yeah, I noticed the same, it builds fine. It was annoying seeing it every time I used vue-tsc --build so that is why I pushed that change

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
cquirosj: yeah, good catch, that is something I left behind when I was attempting using an union type for addressing the same thing you mentioned in this other comment https://github.com/Particular/ServicePulse/pull/1769#discussion_r1515400217

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
cquirosj: not required?

src/ServicePulse.Host/vue/tsconfig.node.json
```diff
@@ -9,6 +9,7 @@
     "noEmit": true,
     "module": "ESNext",
     "moduleResolution": "Bundler",
-    "types": ["node"]
+    "types": ["node"],
+    "lib": ["ES2022", "DOM", "DOM.Iterable"]
```
johnsimons, line 13: This is not required

src/ServicePulse.Host/vue/vitest.config.ts
johnsimons: This is picked up from Node automatically, we don't need to define them

src/ServicePulse.Host/vue/vite.config.ts
```diff
@@ -67,10 +70,7 @@ export default defineConfig({
       external: ["./js/app.constants.js"],
     },
   },
-  server: {
-    headers: {
-      "Content-Security-Policy": createCSPOverrides(port, defaultUrls),
```
johnsimons, line 72: Why is this now removed?

johnsimons (changes requested):

johnsimons:
@cquirosj As part of this PR, there should not be a need to modify either `vite.config.ts` or `tsconfig.*.ts` at all, are there any issues you are having building the project?

src/ServicePulse.Host/vue/vite.config.ts
cquirosj: @johnsimons  I removed it by mistake, I keep it removed only locally because it affects my workflow, but forgot to un-stage it during my last commit

src/ServicePulse.Host/vue/src/components/failedmessages/FlowDiagram.vue
```diff
@@ -1,48 +1,70 @@
...
+  isError: boolean;
+  sagaName: string;
+  link: {
+    name: string;
+    nodeName: string;
+  };
+  timeSent: string;
+  level?: number;
+  width?: number;
+  XPos?: number;
```
johnsimons, line 38: could be lowercase `x`?

src/ServicePulse.Host/vue/src/components/failedmessages/FlowDiagram.vue
```diff
@@ -1,48 +1,70 @@
...
   if (sagaHeader) {
-    sagaName = sagaHeader.value.split(", ")[0];
+    sagaName = sagaHeader.value?.split(", ")[0] ?? "";
   }
 
   const type = (() => {
-    if (message.headers.find((header) => header.key === "NServiceBus.MessageIntent").value === "Publish") return messageType.Event;
-    else if (message.headers.find((header) => header.key === "NServiceBus.IsSagaTimeoutMessage")?.value?.toLowerCase() === "true") return messageType.Timeout;
-    return messageType.Command;
+    if (message.headers.find((header) => header.key === "NServiceBus.MessageIntent")?.value === "Publish") return MessageType.Event;
```
johnsimons, line 65: I think MessageIntent is an enum in c# land, should we have a similar enumeration in ts?

src/ServicePulse.Host/vue/src/components/failedmessages/FlowDiagram.vue
```diff
@@ -90,12 +112,12 @@ function constructNodes(mappedMessages) {
         return group.reduce(
           ({ result, currentWidth, previousParent }, message) => {
             //position on current level needs to be based on parent Node, so see if one exists
-            const parentMessage = previousLevel?.find((plMessage) => message.parentId === plMessage.messageId && message.parentEndpoint === plMessage.receivingEndpoint);
+            const parentMessage = previousLevel?.find((plMessage) => message.parentId === plMessage.messageId && message.parentEndpoint === plMessage.receivingEndpoint) ?? null;
             //if the current parent node is the same as the previous parent node, then the current position needs to be to the right of siblings
             const currentParentWidth = previousParent === parentMessage ? currentWidth : 0;
-            const startX = parentMessage == null ? 0 : parentMessage.XPos - parentMessage.width / 2;
+            const startX = parentMessage == null ? 0 : (parentMessage.XPos as number) - (parentMessage.width as number) / 2;
```
johnsimons, line 118: I see now what you were talking about @PhilBastian.
In this case use the assertion operator, https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
```diff
@@ -1,60 +1,79 @@
...
 import TimeSince from "../TimeSince.vue";
 import LicenseExpired from "../../components/LicenseExpired.vue";
 import ServiceControlNotAvailable from "../ServiceControlNotAvailable.vue";
 import ConfirmDialog from "../ConfirmDialog.vue";
 import routeLinks from "@/router/routeLinks";
+import FailureGroupView from "@/resources/FailureGroupView";
+import { TYPE } from "vue-toastification";
+
+interface WorkflowState {
+  status: string;
```
johnsimons, line 20: Should this be typed? It looks like there are a specific set of allowed strings.

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
```diff
@@ -1,60 +1,79 @@
...
+}
+
+interface ExtendedFailureGroupView extends FailureGroupView {
+  index: number;
+  need_user_acknowledgement?: boolean;
+  workflow_state: WorkflowState;
+  operation_remaining_count?: number;
+  hover2?: boolean;
+  hover3?: boolean;
+  operation_start_time?: string;
```
johnsimons, line 33: camelCase

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
```diff
@@ -1,60 +1,79 @@
...
+const groupRestoreSuccessful = ref<boolean | null>(null);
+const selectedClassifier = ref<string | null>(null);
+const classifiers = ref<string[]>([]);
 
 async function getGroupingClassifiers() {
-  const response = await useFetchFromServiceControl("recoverability/classifiers");
-  const data = await response.json();
-  classifiers.value = data;
+  const [response, data] = await useTypedFetchFromServiceControl<string[]>("recoverability/classifiers");
+  if (response.ok) {
```
johnsimons, line 58: See https://particularsoftware.slack.com/archives/C05URQN9ZFS/p1709874145526279

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
```diff
@@ -144,38 +162,37 @@ function createWorkflowState(optionalStatus, optionalTotal, optionalFailed) {
...
+  if (group) {
+    // We're starting a restore, poll more frequently
+    changeRefreshInterval(1000);
+    undismissedRestoreGroups.value.push(group);
+
+    group.workflow_state = { status: "restorestarted", message: "Restore request initiated..." };
+    group.operation_start_time = new Date().toUTCString();
+
+    const result = await useRestoreGroup(group.id);
+    if (result.message === "success") {
```
johnsimons, line 182: This looks weird

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -364,29 +403,29 @@ onUnmounted(() => {
...
+                <span v-if="failedMessage?.restoring" title="Message is being retried" class="label sidebar-label label-info metadata-label">Restoring...</span>
+                <span v-if="failedMessage?.archiving" title="Message is being deleted" class="label sidebar-label label-info metadata-label">Deleting...</span>
+                <span v-if="failedMessage?.archived" title="Message is being deleted" class="label sidebar-label label-warning metadata-label">Deleted</span>
+                <span v-if="failedMessage?.resolved" title="Message was processed successfully" class="label sidebar-label label-warning metadata-label">Processed</span>
+                <span v-if="failedMessage?.number_of_processing_attempts ?? 0 > 1" :title="'This message has already failed ' + failedMessage?.number_of_processing_attempts + ' times'" class="label sidebar-label label-important metadata-label"
+                  >{{ failedMessage?.number_of_processing_attempts }} Retry Failures</span
```
PhilBastian: I'm not sure the logic is correct here (nothing to do with this PR). The retry failures would be the number_of_processing_attempts - 1, since the first processing attempt wouldn't be a retry

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
```diff
@@ -376,7 +415,7 @@ onUnmounted(() => {
                 <span v-if="failedMessage.archiving" title="Message is being deleted" class="label sidebar-label label-info metadata-label">Deleting...</span>
                 <span v-if="failedMessage.archived" title="Message is being deleted" class="label sidebar-label label-warning metadata-label">Deleted</span>
                 <span v-if="failedMessage.resolved" title="Message was processed successfully" class="label sidebar-label label-warning metadata-label">Processed</span>
-                <span v-if="failedMessage.number_of_processing_attempts > 1" :title="'This message has already failed ' + failedMessage.number_of_processing_attempts + ' times'" class="label sidebar-label label-important metadata-label"
+                <span v-if="failedMessage.number_of_processing_attempts ?? 0 > 1" :title="'This message has already failed ' + failedMessage.number_of_processing_attempts + ' times'" class="label sidebar-label label-important metadata-label"
                   >{{ failedMessage.number_of_processing_attempts }} Retry Failures</span
```
PhilBastian, line 419: I'm not sure the logic is correct here (nothing to do with this PR). The retry failures would be the number_of_processing_attempts - 1, since the first processing attempt wouldn't be a retry

src/ServicePulse.Host/vue/src/resources/SortOptions.ts
```diff
@@ -0,0 +1,18 @@
+import FailedMessage from "./FailedMessage";
+import FailureGroup from "./FailureGroup";
+
+export type GroupType = FailedMessage | FailureGroup;
+export type GroupPropertyType = string | number | Date;
+
+export default interface SortOptions {
+  description: string;
+  dir: string;
```
johnsimons, line 9: Is this `SortDirection`?

PhilBastian: yep, fixed on incoming branch

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
PhilBastian: not the groupId ref, but the selectedPeriod ref

src/ServicePulse.Host/vue/src/components/failedmessages/MessageView.vue
PhilBastian: i don't know what i was referring to anymore...

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
PhilBastian: yep, pretty terrible:
```js
async function evaluateResponse(response: Response) {
  return {
    message: response.ok ? "success" : `error: ${response.statusText}`,
  };
}```

I'll see how painful it is to refactor all the usages of this

**Comment by cquirosj** [comment]
> @cquirosj As part of this PR, there should not be a need to modify either `vite.config.ts` or `tsconfig.*.ts` at all, are there any issues you are having building the project?

All changes made to config.ts files have been reverted.

File: src/ServicePulse.Host/vue/tsconfig.app.json [hunk 45]
```diff
@@ -15,6 +16,7 @@
     },
     "strict": true,
     "sourceMap": true,
-    "lib": ["ES2022", "DOM", "DOM.Iterable"]
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
+    "types": ["node"]
```
johnsimons, line 20: You can remove `types` @cquirosj

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -57,18 +62,12 @@ function loadMessages() {
   return loadPagedMessages(groupId.value, pageNumber.value, "", "", startDate.toISOString(), endDate.toISOString());
 }
 
-async function loadGroupDetails(groupId) {
-  const response = await useFetchFromServiceControl(`archive/groups/id/${groupId}`);
-  const data = await response.json();
+async function loadGroupDetails(groupId: string) {
+  const [, data] = await useTypedFetchFromServiceControl<FailureGroup>(`archive/groups/id/${groupId}`);
```
johnsimons, line 66: Unfortunately it may be too late, but it would be nice for all the URLs to be typed from `routeLinks`

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
```diff
@@ -127,57 +127,56 @@ function updateMessagesScheduledDeletionDate(messages) {
...
 
 function isRestoreInProgress() {
   return messages.value.some((message) => message.restoreInProgress);
 }
 
-function changeRefreshInterval(milliseconds) {
-  if (typeof refreshInterval !== "undefined") {
-    clearInterval(refreshInterval);
+function changeRefreshInterval(milliseconds: number) {
+  if (refreshInterval != null) {
```
johnsimons, line 175: it seems `refreshInterval` is defined as `number | undefined`

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
johnsimons: ping

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
johnsimons: ping

src/ServicePulse.Host/vue/src/components/failedmessages/FlowDiagram.vue
johnsimons: ping

cquirosj: https://github.com/Particular/ServicePulse/pull/1769/commits/8527ab283ca91408b6c6f80028064f398cdb8618

src/ServicePulse.Host/vue/src/components/failedmessages/DeletedMessageGroups.vue
cquirosj: https://github.com/Particular/ServicePulse/pull/1780/files

src/ServicePulse.Host/vue/src/components/failedmessages/AllDeletedMessages.vue
PhilBastian: yep, this check is for null or undefined

johnsimons (approved):

PR #1773, opened 2024-03-08
Migrate OrderBy component to TypeScript

TravisNickels:


mikeminutillo (approved):
Approved, although I made a few tweaks so I would appreciate another set of eyes

src/ServicePulse.Host/vue/src/components/failedmessages/OrderBy.vue
```diff
@@ -1,14 +1,22 @@
...
-});
+import SortOptions, { GroupType } from "@/resources/SortOptions";
+
+const emit = defineEmits<{
+  sortUpdated: [option: SortOptions];
+}>();
+
+const props = withDefaults(
+  defineProps<{
+    hideSort?: boolean;
```
PhilBastian, line 12: is there any reason to mark this as nullable when it has a default?

src/ServicePulse.Host/vue/src/components/failedmessages/OrderBy.vue
```diff
@@ -31,14 +39,18 @@ function loadSavedSortOption() {
     const sortBy = getSortOptions().find((sort) => {
       return sort.description.toLowerCase() === criteria.toLowerCase();
     });
-    return { sort: getSortFunction(sortBy.selector, direction), dir: direction, description: sortBy.description };
+    if (sortBy) {
+      sortBy.sort = getSortFunction(sortBy.selector, direction);
+      sortBy.dir = direction;
+      return sortBy;
+    }
```
PhilBastian, line 46: this is modifying the original referenced object, whereas the old code was not. I would prefer to have
```suggestion
    if (sortBy) {
      return { ...sortBy, sort: getSortFunction(sortBy.selector, direction), dir: direction }
    }
```

src/ServicePulse.Host/vue/src/components/failedmessages/OrderBy.vue
```diff
@@ -31,14 +39,18 @@ function loadSavedSortOption() {
...
   }
 
   return props.sortOptions[0];
 }
 
-function getSortFunction(selector, dir) {
-  return (firstElement, secondElement) => {
+function getSortFunction(selector: SortOptions["selector"], dir: string) {
+  return (firstElement: GroupType, secondElement: GroupType) => {
     if (dir === "asc") {
```
PhilBastian, line 54: the direction strings can be extracted to an enum

src/ServicePulse.Host/vue/src/resources/SortOptions.ts
```diff
@@ -0,0 +1,13 @@
+import FailedMessage from "./FailedMessage";
+import FailureGroup from "./FailureGroup";
+
+export type GroupType = FailedMessage | FailureGroup;
+export type GroupPropertyType = string | number | Date;
+
+export default interface SortOptions {
+  description: string;
+  dir: string;
```
PhilBastian, line 9: I would prefer 'direction'

PhilBastian (approved):

src/ServicePulse.Host/vue/src/components/failedmessages/OrderBy.vue
mikeminutillo: I believe it has to be nullable so that the caller does not have to include it, but giving it a default allows the dev tools to recognize that it will not be null.

src/ServicePulse.Host/vue/src/resources/SortOptions.ts
mikeminutillo: This change is better done after the other components are migrated to ensure we don't mess up any references

src/ServicePulse.Host/vue/src/components/failedmessages/OrderBy.vue
mikeminutillo: Done

src/ServicePulse.Host/vue/src/components/failedmessages/OrderBy.vue
mikeminutillo: Done

PR #1788, opened 2024-03-19
Update min width

TravisNickels:


PhilBastian (changes requested):
i don't think we want to do this: someone on a standard 1920x1080 screen would hit this min-width just by half-screening their browser. Can we discuss what you were hoping to achieve with this change?

PR #1790, opened 2024-03-22
Convert Custom Checks to Vue

mikeminutillo:
Converting custom checks to vuejs.

### DIFF
--- a/src/ServicePulse.Host/vue/src/components/CustomCheckView.vue
+++ b/src/ServicePulse.Host/vue/src/components/CustomCheckView.vue
@@ -13,8 +13,8 @@ async function dismissCustomCheck(id: string) {
 <template>
   <div class="row box box-warning box-no-click">
     <div class="col-sm-12 no-side-padding">
-      <div class="row">
-        <div class="col-xs-10">
+      <div class="custom-check-row">
+        <div>
           <div class="row box-header">
             <div class="col-sm-12 no-side-padding">
               <p class="lead pre-wrap">{{ customCheck.failure_reason }}</p>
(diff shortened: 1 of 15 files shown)


src/ServicePulse.Host/vue/src/components/CustomCheckView.vue
```diff
@@ -0,0 +1,41 @@
+<script setup lang="ts">
```
PhilBastian, line 1: should be called CustomChecks or CustomChecksList, not CustomChecksView

mikeminutillo: It's the view of a single custom check. I originally called it `CustomCheck` but that conflicted with the type in the `Resources` folder.

src/ServicePulse.Host/vue/src/components/DataView.vue
```diff
@@ -13,6 +17,7 @@ const props = withDefaults(
     autoRefreshSeconds: number;
     showPagination?: boolean;
     showItemsPerPage?: boolean;
+    apiParams?: object;
```
TravisNickels, line 20:
```suggestion
    apiParams?: AdditionalDataViewParameters;
```
Shouldn't this be an explicit type of `AdditionalDataViewParameters` instead of just `object`?  In the CustomChecksView.vue we pass an object of type `AdditionalDataViewParameters` to the `DataView` api-params component parameter.

TravisNickels (approved):

src/ServicePulse.Host/vue/src/components/DataView.vue
mikeminutillo: I won't need these changes to DataView after https://github.com/Particular/ServicePulse/pull/1793

PhilBastian (approved):

PR #1796, opened 2024-04-02
Add automated application tests for filtering monitoring endpoints

TravisNickels:
Adds testing for filtering monitoring endpoints

### DIFF
--- a/src/ServicePulse.Host/vue/test/specs/monitoring/filtering-endpoints.spec.ts
+++ b/src/ServicePulse.Host/vue/test/specs/monitoring/filtering-endpoints.spec.ts
@@ -1,32 +1,69 @@
 import { expect } from "vitest";
 import { it, describe } from "../../drivers/vitest/driver";
-import { fireEvent, screen } from "@testing-library/vue";
+import { waitFor } from "@testing-library/vue";
+import { enterFilterString } from "./actions/enterFilterString";
+import { endpointWithName } from "./questions/endpointWithName";
 import * as precondition from "../../preconditions";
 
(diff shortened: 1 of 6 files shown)


src/ServicePulse.Host/vue/test/specs/monitoring/filtering-endpoints.spec.ts
```diff
@@ -1,32 +1,319 @@
...
+    });
+
+    it("Example: The permalink does not have the filter parameter set and then a filter string is entered that matches only 1 endpoint", async ({ driver }) => {
+      //Arrange
+      await driver.setUp(precondition.serviceControlWithMonitoring);
+      await driver.setUp(precondition.monitoredEndpointsNamed(["Universe.Solarsystem.Earth.Endpoint1", "Universe.Solarsystem.Earth.Endpoint2", "Universe.Solarsystem.Earth.Endpoint3"]));
+
+      //Act
+      await driver.goTo("monitoring?filter=Endpoint1");
+      let foo = await driver.getRouter();
```
cquirosj, line 305: Although exposing the Vue router is very convenient, it ties the test to Vue, which we want to avoid. 

- We could code assertions against the window.location instead. 
- Another option is to continue using vue router but abstracted away, hidden in the driver, but never return the whole driver itself, make the driver have a method with a name like "CurrentPath" or something like that, that only returns a string.

TravisNickels: Using waitFor with window.location seems to work.

cquirosj (approved):

PR #1800, opened 2024-04-10
Refactor page header component

mikeminutillo:
This PR moves all of the nav item elements into their own components and then renders them based on a data-driven list.

It also shifts the CSS for active items to use the built-in vue classes

### DIFF
--- a/src/ServicePulse.Host/vue/src/assets/header-menu-item.css
+++ b/src/ServicePulse.Host/vue/src/assets/header-menu-item.css
@@ -6,3 +6,8 @@
 .navbar > .container-fluid > div {
   margin: 0 1em;
 }
+
+#navbar .router-link-active {
+  background: transparent !important;
+  border-bottom: 5px solid #00a3c4;
+}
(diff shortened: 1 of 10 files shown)


cquirosj (approved):

TravisNickels (approved):

johnsimons:
@mikeminutillo I had a look at this, and I wonder if there is a way to make this more like a top menu registration API instead.
We currently are using vue component to display the items in the menu, but IMO the menu style and layout should not be a concern of the user of this API.
Example in the component below I highlighted the areas of the code that the consumer of the API needs to provide (I forgot the icon!)
image
As you can see you have a data structure like:
```json
{ 
    title: "Failed Messages",
    link: "routeLinks.failedMessage.root",
    icon: "fa-envelope"
    renderBadge: maybe a callback here
}
```

So I am wondering if we could have some kind of "add things to topmenu" API:
```
import topmenuapi;
import dashboarregistration;
import failedmessageregistration;

topmenuapi.Add(dashboarregistration);
topmenuapi.Add(failedmessageregistration);
topmenuapi.Add({.....});
```
I hope you get what I am talking about!

mikeminutillo:
> I hope you get what I am talking about!

I do, and I see that as a potential subsequent fix. The goal here was to make it so that you can test the way that features interact with the menu and dashboard in isolation without knowing anything about the others. Changes to the Custom Checks and Heartbeats menu are disconnected from anything else in the top menu.

PR #1807, opened 2024-04-12
implement remove endpoint

PhilBastian:
implement remove endpoint


src/ServicePulse.Host/vue/src/stores/HeartbeatsStore.ts
```diff
@@ -107,6 +109,13 @@ export const useHeartbeatsStore = defineStore("HeartbeatsStore", () => {
     filterString.value = filter;
   }
 
+  async function deleteEndpoint(endpoint: Endpoint) {
+    useShowToast(TYPE.INFO, "Info", "Removing Endpoint");
```
PhilBastian, line 113: note: legacy implementation didn't have the toast notifications, but if the delete operation was stalled for some reason (e.g. DB lock) then the user wouldn't have any feedback

mikeminutillo: Should this use `autoRefresh.executeAndResetTimer()`? otherwise the delete call may happen, update the UI, then an auto-refresh call brings it back, then a subsequent removes it again.

mikeminutillo (approved):

TravisNickels (approved):

PR #1812, opened 2024-04-16
Adding ServicePulse version to all requests to ServiceControl instances

johnsimons:
This PR adds `Particular-ServicePulse-Version` header to all ServiceControl instance calls.
ServiceControl will use this information as part of https://github.com/Particular/ServiceControl/pull/3965.
The plan is to send this data as part of the report generation that customer produce.


mikeminutillo (approved):

mikeminutillo:
NOTE: This is a breaking change as older versions of ServiceControl do not explicitly allow this header.

i.e. https://github.com/Particular/ServiceControl/blob/release-5.0/src/ServiceControl/Infrastructure/WebApi/Cors.cs#L44

PR #1819, opened 2024-04-19
Fix to prevent flaky tests

cquirosj:
This is an attempt to prevent flaky application tests. Sometimes, the window object could have been disposed from the test environment by the time window.clearTimeout is invoked. It may be the case that this change is not enough, and invoking the clearTimeout function directly instead of a window.clearTimeout also fails. 

image

### DIFF
--- a/src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
+++ b/src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
@@ -3,7 +3,7 @@ export default function useAutoRefresh(refreshAction: () => Promise<void>, timeo
 
   function stopTimer() {
     if (refreshInterval !== null) {
-      clearTimeout(refreshInterval);
+      window?.clearTimeout(refreshInterval);
       refreshInterval = null;
     }
   }
--- a/src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
+++ b/src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
@@ -10,7 +10,7 @@ export default function useAutoRefresh(refreshAction: () => Promise<void>, timeo
 
   function startTimer() {
     stopTimer();
-    refreshInterval = window.setTimeout(() => {
+    refreshInterval = window?.setTimeout(() => {
       executeAndResetTimer();
     }, timeout);
   }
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
```diff
@@ -3,7 +3,7 @@ export default function useAutoRefresh(refreshAction: () => Promise<void>, timeo
 
   function stopTimer() {
     if (refreshInterval !== null) {
-      window.clearTimeout(refreshInterval);
+      clearTimeout(refreshInterval);
```
PhilBastian, line 6: this is an issue - setTimeout and clearTimeout are interpreted by typescript as the versions from Node rather than window

cquirosj: What is the issue? Will that affect the bundle generated by the build? What am I missing?

cquirosj, line 6:
```suggestion
      window?.clearTimeout(refreshInterval);
```

src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
```diff
@@ -3,7 +3,7 @@ export default function useAutoRefresh(refreshAction: () => Promise<void>, timeo
 
   function stopTimer() {
     if (refreshInterval !== null) {
-      window.clearTimeout(refreshInterval);
+      window?.clearTimeout(refreshInterval);
```
cquirosj, line 6: @PhilBastian how about now?

src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
mikeminutillo: They have different signatures so TypeScript will complain about `setTimeout` but not `window.setTimeout`. The final bundle is unaffected. This is a develop/build-time issue.

src/ServicePulse.Host/vue/src/composables/autoRefresh.ts
PhilBastian: for consistency do you want to put the same null propagation on the setTimeout call also? It shouldn't change the type of the refreshInterval handle, since it can already be null

PhilBastian (approved):

PR #1825, opened 2024-04-23
When editing a failed message copy from edited headers only keys and values

mauroservienti:
The ServiceControl `master` branch switched from `Newtonsoft.Json` to `System.Text.Json`, which is less permissive.

When ServicePulse retries an edited failed message, it posts to ServiceControl the edited body and edited headers alongside a bunch of unrelated header properties (that ServicePulse uses to pilot the edit dialog user interface). `Newtonsoft.Json` happily accepted the incoming headers with additional properties to an `IEnumerable<KeyValuePair<string, string>>`, discarding the additional properties. The most recent ServiceControl version expects a `Dictionary<string, string>` and `System.Text.Json` is not so flexible and throws a deserialization exception.

This PR changes how ServicePulse posts the header by only posting header keys and values without the additional properties.

- [x] Test on ServiceControl `master`
- [x] Test on ServiceControl 5.1.0
- [x] Test on ServiceControl 5.0.5
- [x] Test on ServiceControl 4.33.x

### DIFF
--- a/src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
+++ b/src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
@@ -23,10 +23,15 @@ export async function useRetryEditedMessage(
   id: string,
   editedMessage: Ref<{
     messageBody: string;
-    headers: unknown[];
+    headers: any[];
   }>
 ) {
-  const payload = { message_body: editedMessage.value.messageBody, message_headers: editedMessage.value.headers };
+  let editedHeaders: { [key: string]: string } = {};
+  for (let index = 0; index < editedMessage.value.headers.length; index++) {
+    const header = editedMessage.value.headers[index];
+    editedHeaders[header.key] = header.value;
+  }
+  const payload = { message_body: editedMessage.value.messageBody, message_headers: editedHeaders };
   const response = await usePostToServiceControl(`edit/${id}`, payload);
   if (!response.ok) {
     throw new Error(response.statusText);
--- a/src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
+++ b/src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
@@ -26,12 +26,13 @@ export async function useRetryEditedMessage(
     headers: any[];
   }>
 ) {
-  let editedHeaders: { [key: string]: string } = {};
-  for (let index = 0; index < editedMessage.value.headers.length; index++) {
-    const header = editedMessage.value.headers[index];
-    editedHeaders[header.key] = header.value;
-  }
-  const payload = { message_body: editedMessage.value.messageBody, message_headers: editedHeaders };
+  const payload = {
+    message_body: editedMessage.value.messageBody,
+    message_headers: editedMessage.value.headers.reduce((result, header) => {
+      const { key, value } = header as { key: string; value: string };
+      result[key] = value;
+      return result;
+    }, {} as { [key: string]: string }),
   const response = await usePostToServiceControl(`edit/${id}`, payload);
   if (!response.ok) {
     throw new Error(response.statusText);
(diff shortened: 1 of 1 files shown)


src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
```diff
@@ -23,10 +23,15 @@ export async function useRetryEditedMessage(
...
+    headers: any[];
   }>
 ) {
-  const payload = { message_body: editedMessage.value.messageBody, message_headers: editedMessage.value.headers };
+  let editedHeaders: { [key: string]: string } = {};
+  for (let index = 0; index < editedMessage.value.headers.length; index++) {
+    const header = editedMessage.value.headers[index];
+    editedHeaders[header.key] = header.value;
+  }
+  const payload = { message_body: editedMessage.value.messageBody, message_headers: editedHeaders };
```
PhilBastian, line 34:
```suggestion
  let editedHeaders: { [key: string]: string } = {};
  for (let index = 0; index < editedMessage.value.headers.length; index++) {
    const header = editedMessage.value.headers[index];
    editedHeaders[header.key] = header.value;
  }
  const payload = { message_body: editedMessage.value.messageBody, message_headers: editedHeaders };
```
const payload = {
    message_body: editedMessage.value.messageBody,
    message_headers: editedMessage.value.headers.reduce((result, header) => {
      const { key, value } = header as { key: string; value: string };
      result[key] = value;
      return result;
    }),
  };

PhilBastian: this is going to need to be conditionally done according to the ServiceControl version. 4.33 works with the old headers array

johnsimons: @PhilBastian from what I gather from the discussion, the old code used to pass extra props (which were ignored by JsonNet at deserialization time in SC), so all this code is doing is ensure those extra props are not being passed in, which as far as I understand is backwards compatible with prior versions of SC.
So I don't think we need to conditionally protect this code/

src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
```diff
@@ -23,10 +23,15 @@ export async function useRetryEditedMessage(
   id: string,
   editedMessage: Ref<{
     messageBody: string;
-    headers: unknown[];
+    headers: any[];
```
johnsimons, line 26: Please, give this a type!

src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
PhilBastian: old code sent an array to the API, containing objects with key and value properties. This code sends an object to the API, containing properties where the property names are the keys

src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
mauroservienti: The issue is that I have no idea how to do it 😆 any hint is welcome, I'm totally new to TypeScript

mauroservienti:
@PhilBastian, can you please have another look? In https://github.com/Particular/ServicePulse/pull/1825/commits/f714b50078b0c022419ed63557c24373ffbdb60f, I introduced a condition to post a different payload depending on the ServiceControl version.

mauroservienti:
Successfully tested against all supported ServiceControl versions and `master`

cquirosj (approved):

src/ServicePulse.Host/vue/src/composables/serviceFailedMessage.ts
PhilBastian: { key: string; value: string }[] should work... although that may require a typecast at callsites

PhilBastian (approved):

PR #1831, opened 2024-04-26
Use the X-Particular-Reason header to get the response StatusText from ServiceControl versions later than 5.1

mauroservienti:
HTTP/2 doesn't support the reason phrase when returning errors in an HTTP response. In ServiceControl, we then stopped using it and, for now, moved to a custom `X-Particular-Reason` header ([see, for example, this controller](https://github.com/Particular/ServiceControl/blob/da452ea82a2cf6aca420d73029c7861cc009c764/src/ServiceControl/MessageRedirects/Api/MessageRedirectsController.cs#L55-L58)). A proper solution is more complex and out of scope for now. 

This PR updates ServicePulse to use the new header.

### DIFF
--- a/src/Frontend/src/composables/serviceNotifications.ts
+++ b/src/Frontend/src/composables/serviceNotifications.ts
@@ -2,6 +2,8 @@ import { usePostToServiceControl, useTypedFetchFromServiceControl } from "./serv
 
 import type EmailNotifications from "@/resources/EmailNotifications";
 import type UpdateEmailNotificationsSettingsRequest from "@/resources/UpdateEmailNotificationsSettingsRequest";
+import { useIsSupported } from "@/composables/serviceSemVer";
+import { environment } from "@/composables/serviceServiceControl";
 
 export async function useEmailNotifications() {
   try {
@@ -33,8 +35,16 @@ export async function useUpdateEmailNotifications(settings: UpdateEmailNotificat
 export async function useTestEmailNotifications() {
   try {
     const response = await usePostToServiceControl("notifications/email/test");
+
+    let responseStatusText;
+    if (useIsSupported(environment.sc_version, "5.2.0")) {
+      responseStatusText = response.headers.get("X-Particular-Reason");
+    }else{
+      responseStatusText = response.statusText;
+    }
+
     return {
-      message: response.ok ? "success" : "error:" + response.statusText,
+      message: response.ok ? "success" : "error:" + responseStatusText,
     };
   } catch (err) {
     console.log(err);
--- a/src/Frontend/src/composables/serviceNotifications.ts
+++ b/src/Frontend/src/composables/serviceNotifications.ts
@@ -37,9 +37,9 @@ export async function useTestEmailNotifications() {
     const response = await usePostToServiceControl("notifications/email/test");
 
     let responseStatusText;
-    if (useIsSupported(environment.sc_version, "5.2.0")) {
+    if (useIsSupported(environment.sc_version, "5.2")) {
       responseStatusText = response.headers.get("X-Particular-Reason");
-    }else{
+    } else{
       responseStatusText = response.statusText;
     }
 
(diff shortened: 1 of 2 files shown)


src/ServicePulse.Host/vue/src/composables/serviceRedirects.ts
```diff
@@ -59,19 +69,35 @@ export async function useCreateRedirects(sourceEndpoint: string, targetEndpoint:
...
   };
 }
 
 export async function useDeleteRedirects(redirectId: string) {
   const response = await useDeleteFromServiceControl(`redirects/${redirectId}`);
+  
+  let responseStatusText;
+  if (useIsSupported(environment.sc_version, "5.2.0")) {
+    responseStatusText = response.headers.get("X-Particular-Reason");
+  }else{
```
danielmarbach, line 93:
```suggestion
  } else{
```

src/ServicePulse.Host/vue/src/composables/serviceRedirects.ts
```diff
@@ -59,19 +69,35 @@ export async function useCreateRedirects(sourceEndpoint: string, targetEndpoint:
     fromphysicaladdress: sourceEndpoint,
     tophysicaladdress: targetEndpoint,
   });
+
+  let responseStatusText;
+  if (useIsSupported(environment.sc_version, "5.2.0")) {
+    responseStatusText = response.headers.get("X-Particular-Reason");
+  }else{
```
danielmarbach, line 76:
```suggestion
  } else{
```

src/ServicePulse.Host/vue/src/composables/serviceRedirects.ts
```diff
@@ -46,10 +48,18 @@ export async function useUpdateRedirects(redirectId: string, sourceEndpoint: str
...
```
danielmarbach, line 55:
```suggestion
  } else{
```

src/ServicePulse.Host/vue/src/composables/serviceNotifications.ts
```diff
@@ -33,8 +35,16 @@ export async function useUpdateEmailNotifications(settings: UpdateEmailNotificat
 export async function useTestEmailNotifications() {
   try {
     const response = await usePostToServiceControl("notifications/email/test");
+
+    let responseStatusText;
+    if (useIsSupported(environment.sc_version, "5.2.0")) {
+      responseStatusText = response.headers.get("X-Particular-Reason");
+    }else{
```
danielmarbach, line 42:
```suggestion
    } else{
```

src/ServicePulse.Host/vue/src/composables/serviceNotifications.ts
```diff
@@ -33,8 +35,16 @@ export async function useUpdateEmailNotifications(settings: UpdateEmailNotificat
 export async function useTestEmailNotifications() {
   try {
     const response = await usePostToServiceControl("notifications/email/test");
+
+    let responseStatusText;
+    if (useIsSupported(environment.sc_version, "5.2.0")) {
```
danielmarbach, line 40: Does this take into account 5.2.0 and higher?

danielmarbach (approved):

src/ServicePulse.Host/vue/src/composables/serviceNotifications.ts
mauroservienti: Good catch

```suggestion
    if (useIsSupported(environment.sc_version, "5.2")) {
```

danielmarbach: I checked the logic and I think it would have worked before too since it checks every version you pass against the current version. I don' t know if it can deal with X.Y version only honestly.

danielmarbach: I would say revert this for now since we are always checking against X.Y.Z in other code paths.

mauroservienti:
Rebased to incorporate https://github.com/Particular/ServicePulse/pull/1822

danielmarbach:
Seems to be a low risk change. Merging

PhilBastian (commented):
- All of the let/if/else can be changed to const ternary assignments.
- All of the string concatenation should be string interpolation

danielmarbach:
@PhilBastian your wish is our command ;) https://github.com/Particular/ServicePulse/pull/1834

PR #1832, opened 2024-04-26
Change service pulse fetch to not assume it always talks to SC by pushing setting the versions to a higher level

danielmarbach:
This is necessary to allow having multiple versions of different platform pieces running without CORS issues.

Without the change when running monitoring on a version before 5.2.0 but one of the SC instances updated we will always try to send the version header.


mauroservienti:
Ah! Great catch 🎉

mauroservienti:
Is there a use case in which we also need to Audit instance(s) versions? And the headache is that we could be talking to multiple audit instances using different versions...The primary instance scatter/gather and forwarding APIs always forward incoming headers, IIRC. If the primary version is >5.2 but the audit instance is not, is there a chance we'll forward stuff so that CORS will block it, for example? Do we need to implement the same "filtering" logic in the scatter/gather and forwarding APIs?

mauroservienti:
Rebased to incorporate https://github.com/Particular/ServicePulse/pull/1822

danielmarbach:
> Is there a use case where we also need the Audit instance(s) versions? And the headache is that we could be talking to multiple audit instances using different versions...The primary instance scatter/gather and forwarding APIs always forward incoming headers, IIRC. If the primary version is >5.2 but the audit instance is not, is there a chance we'll forward stuff that CORS blocks, for example? Do we need to implement the same "filtering" logic in the scatter/gather and forwarding APIs?

We will further discuss this and explore during the multi instance upgrade scenarios

PR #1835, opened 2024-04-30
Revert custom ServicePulse version header

mikeminutillo:
Originally added in #1812
Follow up work was added in #1823 and #1832 to make this work.

Before adding this header, it is important that all instances of ServiceControl being accessed will accept it. That includes ServiceControl Audit instances as the primary instance will forward a copy of the request (including headers) over.

Work has been done to ServiceControl to support this, but this work is not yet released:
- https://github.com/Particular/ServiceControl/pull/4093
- https://github.com/Particular/ServiceControl/pull/4121


mauroservienti (approved):

danielmarbach (approved):
Approving under the assumption we slightly twist the reason in the PR

> Before adding this header, it is important that all instances of ServiceControl being accessed will accept it. That includes ServiceControl Audit instances as the primary instance will forward a copy of the request (including headers) over.

CORS is not the issue. We simply don't need that header, and there are better options like including service pulse version information in the throughput requests or the user agent if we need a more generally applicable solution.

cquirosj (approved):

PR #1836, opened 2024-04-30
Fixes long custom check failure reasons layout overflows

PhilBastian:
custom checks can have very long failure reasons which the row CSS was not handling, thus causing layout overflows. This change ensures the failure reason wraps instead of overflowing.


mauroservienti (approved):

src/Frontend/src/components/customchecks/CustomCheckView.vue
```diff
@@ -45,5 +45,14 @@ const store = useCustomChecksStore();
...
   justify-content: space-between;
   align-items: center;
+  gap: 1em;
+}
+
+.custom-check-row-detail {
+  min-width: 0;
+}
+
+.custom-check-row-detail .lead {
```
danielmarbach, line 55: The failure reason already has `<p class="lead pre-wrap">{{ customCheck.failure_reason }}</p>`

Which seems to indicate wrapping should have taken place. Would it make sense to change those existing styles instead?

PhilBastian: pre-wrap class was removed everywhere else except here (from memory this was because it wasn't based on flex sizing and therefore the styles didn't translate to to the new layout model used in the Vue version of SP). I will remove it from here also.
Text wrapping can't happen unless the parent limits the child size to not overflow. That's what this change is about

danielmarbach (approved):

cquirosj (approved):

PR #1841, opened 2024-05-03
Throughput Report V2

jpalac:


src/Frontend/src/views/ThroughputReportView.vue
```diff
@@ -1,32 +1,68 @@
...
 </script>
 
 <template>
-  <div class="container">
-    <div class="row">
-      <div class="col-sm-12">
-        <h1>Throughput Report</h1>
+  <ThroughputNotAvailable />
+  <ServiceControlNotAvailable />
+  <template v-if="connectionState.connected && useIsThroughputSupported()">
```
johnsimons, line 33: I am not sold on this pattern, it feels like this should use [named slots](https://vuejs.org/guide/components/slots.html#named-slots).
I understand this is replicating today's pattern, but I still think we should change it.
@PhilBastian thoughts?

My take on this is something like:
```html
<ThroughputSupport>
	<slot name="unsupported">
                This probably can default to render <ThroughputNotAvailable />
	</slot>
	<slot>
                .......
	</slot>
</ThroughputSupport>
```

src/Frontend/src/views/ThroughputReportView.vue
```diff
@@ -1,32 +1,68 @@
...
+        <div class="col-sm-12">
+          <div class="nav tabs">
+            <div>
+              <h5 class="nav-item" :class="{ active: isRouteSelected(routeLinks.throughputReport.root) }">
+                <RouterLink :to="routeLinks.throughputReport.endpoints.link">Endpoints</RouterLink>
+              </h5>
+              <h5 class="nav-item" :class="{ active: isRouteSelected(routeLinks.throughputReport.setup.link) }">
+                <RouterLink :to="routeLinks.throughputReport.setup.link">Setup</RouterLink>
+              </h5>
+              <!--  <h5 class="nav-item" :class="{ active: isRouteSelected(routeLinks.throughputReport.report.link) }">
```
johnsimons, line 50: We can delete this

src/Frontend/src/views/ThroughputReportView.vue
```diff
@@ -1,32 +1,68 @@
...
+                <RouterLink :to="routeLinks.throughputReport.report.link">Report</RouterLink>
+              </h5> -->
+            </div>
+            <div class="filter-group">
+              <div>
+                <p v-if="!reportState?.report_can_be_generated">{{ reportState?.reason }}</p>
```
johnsimons, line 56: Should the `v-if` be on the div above?

src/Frontend/src/views/throughputreport/SetupView.vue
```diff
@@ -43,64 +55,81 @@ function displayTransportNameForInstructions() {
...
+          <h5>Connection Status</h5>
+          <template v-if="testResults?.broker_connection_result?.connection_successful">
+            <div>
+              <p>Succesfully connected to {{ transportNameForInstructions() }} for throughput collection.</p>
+            </div>
+          </template>
+          <template v-else>
+            <div>
+              <p>Some errors were encountered while attemping to connect to {{ transportNameForInstructions() }} for throughput collection.</p>
+              <p>Please look at the Connection Setup for more details.</p>
```
johnsimons, line 96: Should there be a link here?

src/Frontend/src/views/throughputreport/ThroughputNotAvailable.vue
```diff
@@ -0,0 +1,21 @@
...
+import { useIsThroughputSupported, minimumSCVersionForThroughput } from "@/views/throughputreport/IsThroughputSupported";
+</script>
+
+<template>
+  <div class="sp-loader" v-if="connectionState.connecting && !connectionState.unableToConnect"></div>
+  <template v-if="!connectionState.unableToConnect && !useIsThroughputSupported()">
+    <div class="text-center monitoring-no-data">
+      <h1>System Throughput</h1>
+      <p>
+        The minimum version of ServiceControl required to enable the System Throughput feaure is
```
johnsimons, line 12: spelling

src/Frontend/src/views/throughputreport/ThroughputNotAvailable.vue
```diff
@@ -0,0 +1,21 @@
+﻿<script setup lang="ts">
+import { connectionState } from "@/composables/serviceServiceControl";
+import { useIsThroughputSupported, minimumSCVersionForThroughput } from "@/views/throughputreport/IsThroughputSupported";
+</script>
+
+<template>
+  <div class="sp-loader" v-if="connectionState.connecting && !connectionState.unableToConnect"></div>
+  <template v-if="!connectionState.unableToConnect && !useIsThroughputSupported()">
+    <div class="text-center monitoring-no-data">
+      <h1>System Throughput</h1>
```
johnsimons, line 10: "System" ?

src/Frontend/src/views/throughputreport/SetupView.vue
```diff
@@ -20,13 +28,17 @@ const transport = computed(() => {
...
   settingsInfo.value = settings;
 }
 
-function displayTransportNameForInstructions() {
+async function testConnection() {
+  const test = await throughputClient.test();
+  testResults.value = test;
+}
+
+function transportNameForInstructions() {
```
johnsimons, line 41: This probably should be a computed prop

src/Frontend/src/views/throughputreport/setup/ConnectionView.vue
```diff
@@ -0,0 +1,84 @@
...
+            </div>
+            <p>{{ item.description }}</p>
+          </li>
+        </ul>
+      </div>
+    </div>
+  </div>
+  <div class="row">
+    <h3>Test Connection results</h3>
+    <template v-if="props.settingsInfo?.broker_settings?.length! > 0">
```
johnsimons, line 59: Why don't we always display the diagnostics for the broker?

src/Frontend/src/views/throughputreport/throughputClient.ts
```diff
@@ -34,9 +34,32 @@ class ThroughputClient {
...
+  }
+
+  public async getMasks() {
+    //return await useFetchFromServiceControl(`${this.basePath}/settings/masks`);
+    const [, data] = await useTypedFetchFromServiceControl<string[]>(`${this.basePath}/settings/masks`);
+    return data;
+  }
+
+  public async updateMasks(data: string[]): Promise<void> {
+    const response = await usePostToServiceControl(`${this.basePath}/settings/masks/update`, data);
```
johnsimons, line 62: we don't need the return value

src/Frontend/src/views/throughputreport/throughputClient.ts
```diff
@@ -34,9 +34,32 @@ class ThroughputClient {
...
+      var fileName = "throughput-report.json";
+      const contentDisposition = response.headers.get("Content-Disposition");
+      if (contentDisposition != null) {
+        try {
+          fileName = contentDisposition.split("filename=")[1].split(";")[0].replaceAll('"', "");
+        } catch {
+          //do nothing
+        }
+      }
+      await useDownloadFileFromResponse(response, "application/json", fileName);
```
johnsimons, line 49: well if we are using header from the server, then we should also use `Content-Type`?

src/Frontend/src/views/throughputreport/throughputClient.ts
```diff
@@ -34,9 +34,32 @@ class ThroughputClient {
...
-    //useDownloadFile();
-    //const response = await useFetchFromServiceControl(`${this.basePath}/report/file`);
+  public async downloadReport(spVersion: string) {
+    const response = await useFetchFromServiceControl(`${this.basePath}/report/file?${spVersion}`);
+    if (response.status == 200) {
+      var fileName = "throughput-report.json";
+      const contentDisposition = response.headers.get("Content-Disposition");
+      if (contentDisposition != null) {
+        try {
+          fileName = contentDisposition.split("filename=")[1].split(";")[0].replaceAll('"', "");
```
johnsimons, line 44: I would use something like https://www.npmjs.com/package/@tinyhttp/content-disposition instead

src/Frontend/src/views/throughputreport/throughputClient.ts
```diff
@@ -34,9 +34,32 @@ class ThroughputClient {
     return data;
   }
 
-  public downloadReport(masks: string[]) {
-    //useDownloadFile();
-    //const response = await useFetchFromServiceControl(`${this.basePath}/report/file`);
+  public async downloadReport(spVersion: string) {
+    const response = await useFetchFromServiceControl(`${this.basePath}/report/file?${spVersion}`);
+    if (response.status == 200) {
```
johnsimons, line 39:
```suggestion
    if (response.ok) {
```

src/Frontend/src/views/ThroughputReportView.vue
PhilBastian: slots are for if the decision about what should be displayed needs to be made in the parent. If this fits that criterion then use slots, but currently the decision is being made here, in the child component. If the information is available here then why would you move the decision up to the parent?

src/Frontend/src/composables/fileDownloadCreator.ts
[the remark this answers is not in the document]
jpalac: 2 different methods doing almost the same thing - hence refactoring the thing that was the same. I don't see why it needs to be pushed to master on its own since the new method only applies to our code.

src/Frontend/src/views/throughputreport/setup/ConnectionView.vue
jpalac: Because if they are not using a broker then why display an empty heading.

src/Frontend/src/views/throughputreport/throughputClient.ts
jpalac: Our general view is to use fewer third party plugins so we have fewer dependencies that need updating. Is it worth introducing one for this?

src/Frontend/src/composables/fileDownloadCreator.ts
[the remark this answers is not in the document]
johnsimons: less code to review later on

src/Frontend/src/views/throughputreport/setup/ConnectionView.vue
johnsimons: not sure i understand, lets chat in zoom

src/Frontend/src/views/throughputreport/throughputClient.ts
johnsimons: i think so, no need for us to implement this, given it is not straight forward

johnsimons: > fewer third party plugins

btw, where is this documented?

src/Frontend/src/views/throughputreport/throughputClient.ts
jpalac: updated to use it

src/Frontend/src/views/throughputreport/throughputClient.ts
jpalac: General guidance on third party dependencies [here](https://github.com/Particular/Platform/blob/main/principles/design-principles.md#third-party-dependencies).
A note on [keeping external components to a minimum in SP](https://github.com/Particular/Platform/blob/main/documentation/service-pulse.md#external-components). If we are adding one then there should be a reason and it needs to be added to this document.

johnsimons: The first link seems to be very c# specific.
And the 2nd link doesn't really provide any reasons why!

src/Frontend/src/views/throughputreport/ThroughputSupported.vue
```diff
@@ -0,0 +1,59 @@
...
+          <a class="btn btn-default btn-primary" href="https://particular.net/downloads" target="_blank">Update ServiceControl to latest version</a>
+        </div>
+      </div>
+    </template>
+    <slot />
+  </ConditionalRender>
+</template>
+
+<style scoped>
+.monitoring-no-data {
```
PhilBastian, line 30: there is no need for it to be called "monitoring-no-data" since this isn't part of monitoring.

src/Frontend/src/views/throughputreport/ThroughputSupported.vue
```diff
@@ -0,0 +1,59 @@
...
+  margin: 60px auto 120px;
```
PhilBastian, line 31: can all the px sizing, unless it's for something like a border width, be changed to em sizing?

PR #1842, opened 2024-05-05
Bumping mininum supported Service Control version

cquirosj:


mauroservienti (approved):

danielmarbach (approved):

PhilBastian (approved):

src/Frontend/src/composables/serviceServiceControl.ts
```diff
@@ -183,12 +183,10 @@ export async function useServiceControlConnections() {
   return connections;
 }
 
-watch(environment, (newValue, oldValue) => {
-  if (newValue.is_compatible_with_sc !== oldValue.is_compatible_with_sc) {
-    if (!newValue.is_compatible_with_sc) {
-      useShowToast(TYPE.ERROR, "Error", `You are using Service Control version ${newValue.sc_version}. Please, upgrade to version ${newValue.minimum_supported_sc_version} or higher to unlock new functionality in ServicePulse.`);
-    }
-  }
+watch(() => environment.is_compatible_with_sc, (newValue, oldValue) => {
```
PhilBastian, line 186: oldValue isn't being used so doesn't need to be part of the function signature. 

The old version was very careful about false entries to this watch - was this a carryover from angular or is there a real risk of this code getting hit multiple times?

PhilBastian (approved):

src/Frontend/src/composables/serviceServiceControl.ts
cquirosj: This was carried over from Angular. The check between old value and new value was done all the time because the watcher was for the whole object, therefore triggering every time any property changed. With this new version, I had a breakpoint in the debugger, and it hit only once.

cquirosj: Actually, the problem was that the original watcher was lacking the ,deep:true parameter and it was not triggering when any property changed. This could have also been fixed only by adding the deep = true, but since the code was not requiring the whole object then I decided to watch the only property required.

PR #1849, opened 2024-05-10
Scope remaining styles

PhilBastian:
The migration of ServicePulse to Vue carried across many of the styles from the legacy site, but attempted to locate styles with components. In the interests of time and priority, many of these styles were left unscoped, which led to occasional class clashes once the owning controls were loaded. This change scopes all the remaining unscoped component-based css and, in addition, fixes some styling issues, most significantly around how modals are displayed at lower resolutions


src/Frontend/public/js/app.constants.js
```diff
@@ -4,5 +4,5 @@ window.defaultConfig = {
   version: '1.2.0',
   service_control_url: 'http://localhost:33333/api/',
   monitoring_urls: ['http://localhost:33633/'],
-  showPendingRetry: false,
+  showPendingRetry: true,
```
johnsimons, line 7: is this meant to be changed as part of this PR?

johnsimons (approved):
The change LGTM

src/Frontend/public/js/app.constants.js
PhilBastian: whoops

PR #1746, opened 2024-02-26
Adding Throughput Report area

johnsimons:
Part of https://github.com/Particular/ServicePulse/issues/1888

PR #1863, opened 2024-05-26
Report build fix: Adding default mocked respones for /licensing/* API calls

cquirosj:
Since the acceptance tests setup run the whole application for every single test, every call that can potentially be made to Service Control requires to be mocked. This PR adds default handlers for the /licensing paths (/test, /endpoints, /available) 

Since this commit [1c1640f](https://github.com/Particular/ServicePulse/pull/1746/commits/1c1640f8a20e4c71ef38bc40915e65992742970e) is now initializing the report store from the menuItem, it triggers the Throughput client, which in turn calls /licenses/tests in a timer. This requires having the report paths at least with a default handler. These handlers can be overwritten when writing specific test cases.


johnsimons:
Thanks @cquirosj

PR #1873, opened 2024-06-01
Remove unnecessary monitoring-endpoint calls

TravisNickels:
This removes a call to get monitored-endpoints from `useServiceControlMonitoringStats()`, since the results are never used there is no reason to set an interval to call this function every 5 seconds on app start up.

Also, refactored the monitoring store, removing the `initializeStore()` function

In the master branch you can see that even when the history period is set to 10 minutes (so the refresh should be 10 seconds) the fetch is happening 3 times right after the page is reloaded. And there is also a fetch to `history=1`, which was being done on a 5 second interval by the `useServiceControlMonitoringStats()` on app startup.
image

After removing the fetch calls from the from `useServiceControlMonitoringStats()` the 3 initial fetch calls are removed and the fetch call to `history=1` is also not being called.
image

### DIFF
--- a/src/Frontend/src/composables/serviceServiceControl.ts
+++ b/src/Frontend/src/composables/serviceServiceControl.ts
@@ -8,6 +8,7 @@ import type RootUrls from "@/resources/RootUrls";
 import type FailedMessage from "@/resources/FailedMessage";
 import type MonitoredEndpoint from "@/resources/MonitoredEndpoint";
 import { FailedMessageStatus } from "@/resources/FailedMessage";
+import { useMonitoringStore } from "@/stores/MonitoringStore";
 
 export const stats = reactive({
   active_endpoints: 0,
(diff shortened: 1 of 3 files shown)


src/Frontend/src/composables/serviceServiceControl.ts
```diff
@@ -8,6 +8,7 @@ import type RootUrls from "@/resources/RootUrls";
 import type FailedMessage from "@/resources/FailedMessage";
 import type MonitoredEndpoint from "@/resources/MonitoredEndpoint";
 import { FailedMessageStatus } from "@/resources/FailedMessage";
+import { useMonitoringStore } from "@/stores/MonitoringStore";
```
cquirosj, line 11: This import can be removed
```suggestion
```

src/Frontend/src/stores/MonitoringStore.ts
```diff
@@ -39,12 +38,6 @@ export const useMonitoringStore = defineStore("MonitoringStore", () => {
   });
 
   //STORE ACTIONS
-  async function initializeStore() {
```
cquirosj, line 42: Nice!!

cquirosj (approved):

src/Frontend/src/composables/serviceServiceControl.ts
```diff
@@ -160,10 +161,9 @@ async function useServiceControlStats() {
 }
 
 async function useServiceControlMonitoringStats() {
-  const monitoredEndpointsResult = getMonitoredEndpoints();
   const disconnectedEndpointsCountResult = getDisconnectedEndpointsCount();
 
-  const [, disconnectedEndpoints] = await Promise.all([monitoredEndpointsResult, disconnectedEndpointsCountResult]);
```
cquirosj, line 166: Very, very good catch!!

src/Frontend/src/composables/serviceServiceControl.ts
TravisNickels: Yep, that was left over from another idea I was trying first.

PR #1880, opened 2024-06-19
New code editor

johnsimons:
This is part of https://github.com/Particular/ServicePulse/pull/1746, we are splitting the PR in smaller chunks.
 
The other one had issues when re-rendering different text with different code style in the same instance, the text would not update.

Before:
image

After:
image


PhilBastian (approved):
does the new solution allow for copying the contents with a single click?

johnsimons:
> does the new solution allow for copying the contents with a single click?

@PhilBastian yes
Jun-19-2024 20-03-58

Issue #1766, opened 2024-03-04
When the monitoring backend is unavailable the "View connection settings" button URL is incorrect


mauroservienti:
#### Description

When the Monitoring backend is unavailable, users are presented with the following toast notification:

![image](https://github.com/Particular/ServicePulse/assets/1325611/b36474e7-b99e-44d7-99fe-58754f32b8bf)

The `View connection settings` button link is incorrect (see the above screenshot in the bottom left corner)

#### Versions

Please list the version of the relevant packages or applications in which the bug exists.        

1.38.1

### Steps to reproduce

- Make the Monitoring backend unavailable
- Observe the link in the toast notification

DavidBoike:
The fix will be released in 1.39.0.

Issue #1584, opened 2023-10-10
Failed messages button link is broken when ServicePulse is hosted in a subfolder


kentdr:
#### Description
     
When on the Monitoring page, Endpoint overview.  The "Failed messages" button does not take into account the subfolder where ServicePulse was served from

![image](https://github.com/Particular/ServicePulse/assets/83468000/4a79ca51-cd2e-4d00-8e18-b2fdd6f4d874)

#### Expected behavior

  The link should use the base URL when selected.  

#### Actual behavior
 
 The link does not use the base URL.  

#### Versions

 1.37.0

### Steps to reproduce

- Configure ServicePulse to be served from a subfolder.  
- Setup a monitoring instance and configure ServicePulse to reference it.
- Setup an NServiceBus endpoint with monitoring and cause a message to be written to the error queue.
- Navigate to the Monitoring page in ServicePulse
- Click the envelope icon next to an endpoint with errors.

Aszuul:
The events tab seems to have the same issue. the Service Pulse version we were testing is 1.37. Had to downgrade to 1.30 for now to avoid the issue.

DavidBoike:
The fix will be released in 1.39.0.

Issue #1375, opened 2023-02-14
Subfolder deployments do not work


tmasternak:
There is no way to deploy ServicePulse in a subfolder e.g. `https:\\somedomain.com\subfolder`. In order to host multiple ServicePulse under one domain separate subdomains need to be used e.g. `https:\\prod.domain.com`, `https:\\stagging.domain.com` etc. 

It seems that with VueJs a subfolder deployment option [needs to be specified during build time](https://stackoverflow.com/questions/43879418/deploy-vuejs-app-in-a-subdirectory). As a result, users requiring such a deployment might be required to build from the sources.

yogeshbissessur:
Hi,

we are also experiencing the same problem with our setup, is there a workaround or fixed for this?

tmasternak:
Hi @yogeshbissessur,

they only approach we can recommend for now is to use a subdomain instead of a subfolder.

WilliamBZA:
This will be fixed in 1.34.0. Hopefully it will be released soon.

WilliamBZA:
Changes merged to master, this will make it into the next release.

PR #1890, opened 2024-06-23
Remove base_url and update internal navigation with RouterLink

TravisNickels:
This PR removes the `base_url` property from `public/js/app.constants.js` and updates all anchor links to use the `RouterLink` component from vue-router to make all internal navigation route aware.

The `base_url` property was introduced when AngularJS and VueJs were being used simultaneously to migrate to VueJs.  This allowed ServicePulse versions that used AngluarJS and Vuejs to be deployed in a subfolder.

- https://github.com/Particular/ServicePulse/issues/1375

However, since AngularJS has been completely removed from ServicePulse, the base_url property is no longer needed as long `RouterLink` is used instead of anchor links or any route added to an anchor `href` property is derived from the vue-router.

This PR also resolves these two issues:

- https://github.com/Particular/ServicePulse/issues/1584 
- https://github.com/Particular/ServicePulse/issues/1766


src/Frontend/src/composables/serviceLicense.ts
```diff
@@ -66,11 +59,14 @@ const licenseStatus = reactive({
...
-  watch<UnwrapNestedRefs<License>>(license, async (newValue, oldValue) => {
+let router: ReturnType<typeof useRouter>;
+
+async function useLicense(vueRouter: ReturnType<typeof useRouter>) {
+  router = vueRouter;
+  watch<UnwrapNestedRefs<License>>(license, (newValue, oldValue) => {
     const checkForWarnings = oldValue !== null ? newValue && newValue.license_status != oldValue.license_status : newValue !== null;
     if (checkForWarnings) {
-      displayWarningMessage(newValue.license_status);
+      displayWarningMessage(LicenseStatus.ValidWithExpiringUpgradeProtection);
```
TravisNickels, line 69:
```suggestion
      displayWarningMessage(newValue.license_status);
```
Removing testing value

src/Frontend/src/composables/serviceLicense.ts
```diff
@@ -181,6 +182,7 @@ function getUpgradeDaysLeft(license: UnwrapNestedRefs<License>) {
 async function getLicense() {
   try {
     let [, data] = await useTypedFetchFromServiceControl<LicenseInfo>("license?refresh=true");
+    data.license_status = LicenseStatus.ValidWithExpiredUpgradeProtection;
```
TravisNickels, line 185:
```suggestion

```
Remove another testing value

cquirosj (approved):

PR #1909, opened 2024-07-03
Edit and Retry Content type check allows more text based payloads

ramonsmits:
Edit and Retry Content type check allows more text based payloads by allowing all text based content types which start with `text/` but also all `application/*+json` and `application/*+xml` payloads

Issue #1905, opened 2024-06-28
Refactor remaining composables used for global state management to use Pinia stores to resolve potential flaky acceptance tests caused by module-level state


TravisNickels:
#### Is your improvement related to a problem? Please describe.

Our current Vue.js application has some remaining composables that maintain state at the module level. While this approach simplifies state sharing across components, it introduces significant issues in our acceptance tests conducted with Vitest and JSDOM. Specifically, the shared module-level state could lead to flaky tests due to state persistence between test runs.

> [!NOTE] 
> The application already uses Pinia stores. See https://github.com/Particular/ServicePulse/tree/master/src/Frontend/src/stores. This issue is about completing the migration to Pinia but emphasizes the importance of this work to help with automated testing efforts.

#### Testing challenges when using module-level state

- **Shared State Across Tests:**
  - JavaScript modules are singletons, meaning the module-level state in our composables persists across different test cases.
  - Tests inadvertently affect each other's outcomes because they operate on the same shared state, leading to unpredictable results.

- **Flaky Tests:**
  - Tests may pass or fail inconsistently depending on the order of execution and the residual state left by previous tests.
  - This flakiness undermines the reliability of our test suite and complicates continuous integration efforts.

- **Testing Challenges with Vitest and JSDOM:**
  - Vitest and JSDOM do not automatically isolate module-level state between tests.
  - Workarounds to reset the state between tests are non-trivial and can add unnecessary complexity to the testing setup.

#### Impact

- **Reduced Confidence in Test Results:**
  - Flaky tests make it difficult to discern genuine issues from false negatives or positives.
  - Developers may spend excessive time debugging tests that fail due to shared state rather than actual code defects.

- **Slower Development Cycle:**
  - Time spent troubleshooting flaky tests slows down the development process.
  - Potential for real bugs to be overlooked if tests are assumed to be unreliable.

### Describe the suggested solution

Refactor our composables to use **Pinia** stores instead of maintaining state at the module level. Pinia is the [recommended state management library for Vue.js](https://vuejs.org/guide/scaling-up/state-management.html#pinia) that offers reactive and typesafe stores. 

Many, if not all, of the "composables" in the `Frontend\src\composables` folder can be either moved to Pinia stores or refactored into general functions, with their names adjusted to not start with "use".

With testability in mind, composables should primarily only be used for organizing shared functionality and logic, rather than for global state management.

> [!NOTE] 
[...]

PR #1915, opened 2024-07-10
Adding tests for usage area

johnsimons:


src/Frontend/test/drivers/vitest/driver.ts
```diff
@@ -31,10 +31,9 @@ function makeDriver() {
       return factory({ driver: this });
     },
     disposeApp() {
-      if (!app) {
-        throw new Error("App is not mounted, make your sure to await driver.goTo().");
```
johnsimons, line 35: @cquirosj [this](https://github.com/Particular/ServicePulse/pull/1909/files?diff=unified#diff-edc4597bf9a577eaf7c535af22acd31e0a8ed0b597acced41339e61289179423R34-R36) change, conflicts, with my change here.
In my tests I found it simpler to not having to deal with navigation and ensuring the whole app is running, hence if you look at some of the tests here, they have less setup.

cquirosj: @johnsimons, if you wish to re-use the preconditions functions from the acceptance tests setup, there is no need to use the custom "Test" function to have an instance of the driver. The component tests can create their own driver instance, as demonstrated in the following example where:
- A component test makes use of mocked responses by both:
  - Re-using the pre-defined mocked responses from the acceptance tests (optional)
  - And by setting up the mocked responses directly

https://github.com/Particular/ServicePulse/pull/1916/files#diff-d065d469d8fd3e77aaf7b6e797e3b3e08aa883643c2783a3119048559e0fa3db


By the way, this [PR](https://github.com/Particular/ServicePulse/pull/1916) demonstrates how to test the same scenarios at the view component level vs at the acceptance test level

src/Frontend/src/views/ThroughputReportView.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+        await driver.setUp(precondition.hasLicensingReportAvailable({ report_can_be_generated: false, reason: reason }));
+      });
+
+      expect(screen.getByRole("button", { name: /Download Report/i })).toBeDisabled();
+      expect(screen.getByText(reason)).toBeInTheDocument();
+    });
+  });
+
+  describe("when download report is clicked", () => {
+    test("and no warnings, download happens", async (driver) => {
```
cquirosj, line 93: Could we also include tests that describe the expected content of the file for a given set of usage conditions?

cquirosj: I suppose that is already tested on the backend.

johnsimons: yes the content is a backend responsibility

johnsimons:
@cquirosj anything else?

src/Frontend/src/views/throughputreport/endpoints/DetectedListView.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+            { name: `Set and needs updating`, is_known_endpoint: false, user_indicator: UserIndicator.PlannedToDecommission, max_daily_throughput: 50 },
+          ])
+        );
+        await driver.setUp(updateLicensingEndpoints());
+      });
+
+      const user = userEvent.setup();
+
+      const th = screen.getByText(columnTitle);
+      const table = th.parentElement?.parentElement?.parentElement as HTMLTableElement;
```
cquirosj, line 287: Consider avoiding navigating the current HTML structure and sticking to aria attributes. The current HTML structure is an implementation detail. Here is an example of where we have accessed tabular data in other tests by using aria attributes:

 This is how it looks while navigating the accessibility tree only (this screenshot was taken using Firefox "inspect accessibility properties" option):

image

This [link](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/) shows all the different roles available for the "Table" role (table, rowgroup, row, columheader, cell)

The benefit of sticking to aria only is that the implementation has the freedom to change its hierarchy. As long as the aria roles continue to be used, the specific levels of nesting of the hierarchy won't matter, and the test code has a better chance of surviving implementation changes.

src/Frontend/src/views/throughputreport/endpoints/DetectedListView.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+      });
+
+      const user = userEvent.setup();
+      await user.click(screen.getByRole("button", { name: /Sort by/i }));
+      await user.click(screen.getByRole("link", { name: "name" }));
+
+      const th = screen.getByText(columnTitle);
+      const table = th.parentElement?.parentElement?.parentElement as HTMLTableElement;
+
+      let orderedNames = names.sort((a, b) => a.localeCompare(b));
```
cquirosj, line 235: To avoid having logic leak into the test code, prefer asserting against a fixed expected result for the given set of preconditions (arrangement) the test case is validating. This also makes the test code more straightforward for the reader trying to understand the specific requirement the test is checking.

src/Frontend/src/views/throughputreport/setup/MasksView.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    await driver.setUp(precondition.errorsDefaultHandler);
+
+    return driver;
+  }
+
+  async function renderComponent(body: string[] = []) {
+    const driver = await setup();
+    driver.mockEndpoint(`${serviceControlInstanceUrl}licensing/settings/masks`, { body });
+    useServiceControlUrls();
+    await useServiceControl();
```
cquirosj, line 32: Directly calling implementation details from the system under test (SUT) indicates that the SUT is not written with testability in mind. 

One idea to fix this is to convert these two highlighted composables into Pinia stores. That way, all the data the SUT expects can be set in the arrangement section of the test via the Pinia testing API. Can the scope of the work you are doing be increased to also fix this? I suppose not.


Given that the SUT needs to be refactored with testability in mind, the current state of the code makes Acceptance Tests the best option to go with and leave component testing only for [presentational components](https://medium.com/prod-io/presentational-container-components-when-to-use-what-99be8fa28953) (components that receive the data through props). Testing components that are responsible for fetching the data (views) require us (today) to have these duplicate production calls in the testing setup (implementation details leakage), again, because the SUT is not test-friendly enough yet.

There is this [issue](https://github.com/Particular/ServicePulse/issues/1905) with the suggestion to refactor more of these chucky composables into Pinia store when applicable. useServiceControlUrls(), useServiceControl() and useLicense() are great candidates for that.

Consider adding a comment to that issue so that when the refactor happens, these tests can get updated to no longer require these calls.

johnsimons: It is not ideal to deal with the environmental state on all the tests, so hiding this somewhere can be advantageous. Whether that is a store, a function or something else is debatable.
I don't think that if a component does data load, we need to use acceptance tests only.  (behind the scenes, something else is doing this heavy lifting to make it work anyway)
In my opinion, we should always follow the testing pyramid, where the heavier testing practices are the smallest set.

I am glad I persisted with component testing and got it working.

src/Frontend/src/views/throughputreport/endpoints/DetectedListView.spec.ts
johnsimons: I fixed this

src/Frontend/src/views/throughputreport/endpoints/DetectedListView.spec.ts
johnsimons: I fixed this

johnsimons: Btw @cquirosj , it does look much nicer the syntax 😄

cquirosj (approved):

PR #1919, opened 2024-07-15
Fixing issue with System.Text.Json high severity vulnerability in our tests

johnsimons:
Also updated some dependencies

This fixes error NU1903: Warning As Error: Package 'System.Text.Json' 8.0.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-hh2w-p6rv-4g7w [D:\a\ServicePulse\ServicePulse\src\ServicePulse.sln]

This is just a dependency of the tests, hence no need to release.


jpalac (approved):
LGTM - it's a small change so if any issues we can always undo/change it

PR #1920, opened 2024-07-15
Update dependabot to include npm and docker

johnsimons:
Also updated docker base image to have a specific version, otherwise dependabot won't know when a new version is available


DavidBoike:
You will have to exclude the Dependabot config in the reposync file otherwise a GitHubSync PR will automatically try to clobber your changes within 5 hours to what’s in the RepoStandards repo.

PhilBastian (approved):

PR #1944, opened 2024-07-16
Adding frontend test summary

johnsimons:
Added a summary for the front-end tests.
You can see the result in https://github.com/Particular/ServicePulse/actions/runs/9954116874.
At the moment, I have [configured it](https://github.com/test-summary/action?tab=readme-ov-file#options) to show `all`, but I think `fail` is enough; otherwise, it's too long.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,13 @@ jobs:
       - name: Run application tests
         run: npm run test:application
         working-directory: src/Frontend
+      - name: Frontend Test Summary
+        uses: test-summary/action@v2
+        with:
+          paths: |
+            ./test-component.tap
+            ./test-application.tap
+        if: always()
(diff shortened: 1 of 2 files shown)


johnsimons:
@cquirosj do we need this?

cquirosj:
> @cquirosj do we need this?

@johnsimons 

Is there a format for the report where it can show the results nested? 

Without repeating the parent FEATURE and parent RULE per each test case? 

Something similar to how the test runner looks like. See the sample attached image of how the test runner looks. 

If there is a report like that, I would like to keep `all` as it serves as a quick way of seeing the requirements/features of the system under test otherwise `fail` is enough 

image

cquirosj (approved):

johnsimons:
Not really needed, the default works pretty well

PR #1963, opened 2024-07-23
Fix incorrect logic to display error icon for Usage setup 

johnsimons:
The logic should be:

1. if it is a broker, we return true if the broker connection is unsuccessful
2. if the Audit test connection is false, we return true symbolizing that we have errors;
3. if monitoring is enabled, we return whatever the value of the test connection is for monitoring;
4. by default we return no errors

### DIFF
--- a/src/Frontend/src/stores/ThroughputStore.ts
+++ b/src/Frontend/src/stores/ThroughputStore.ts
@@ -19,19 +19,24 @@ export const useThroughputStore = defineStore("ThroughputStore", () => {
   );
   const refresh = dataRetriever.executeAndResetTimer;
   const hasErrors = computed(() => {
-    if (isBrokerTransport.value) {
-      return !testResults.value?.broker_connection_result.connection_successful;
+    // if it is a broker transport, we return true if connection test is unsuccessful
+    if (isBrokerTransport.value && !testResults.value?.broker_connection_result.connection_successful) {
+      return true;
     }
 
+    // if Audit connection test fails, we will return true.
+    // the connection test will return true if there are no Audit instances configured.
     if (!testResults.value?.audit_connection_result.connection_successful) {
-      return false;
+      return true;
     }
 
+    // if Monitoring is enabled, we return whatever the value of the connection test
     if (useIsMonitoringEnabled()) {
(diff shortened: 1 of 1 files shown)


abparticular:
I'm not sure the logic is correct.  I think we only require either audit or monitoring.  

The current logic seems to always require audit for non-broker transports, which is not correct.

johnsimons:
We don’t require Audit, but I agree the logic is not the best. 
the problem is that we don’t have an api to figure out if audit is configured at all or not and hence we have to “repurpose” the test connection logic to tell us. 
to fix this properly we would need to figure out if the backend has audit configured or not.

abparticular:
GitHub isn't letting me do a suggestion over all of the relevant code lines, but how about:

```
    if (useIsMonitoringEnabled()) {
      return !testResults.value?.monitoring_connection_result.connection_successful;
    }

    return !testResults.value?.audit_connection_result.connection_successful;
 ```

**Comment by jpalac** [comment]
It's not one or the other. If both are connected, then if either one of them has an incorrect configuration then the error should be shown.

If both are connected, both result messages should be shown and both diagnostics.

If only one ia connected then only that connected services status should be used.

File: src/Frontend/src/stores/ThroughputStore.ts [hunk 1]
```diff
@@ -24,14 +24,14 @@ export const useThroughputStore = defineStore("ThroughputStore", () => {
     }
 
     if (!testResults.value?.audit_connection_result.connection_successful) {
```
jpalac, line 26: If we have both audit and monitoring enabled, and one of them has a connection error, then I think we should display the error icon.
So this would check for both:

```suggestion
    if (!testResults.value?.audit_connection_result.connection_successful || (useIsMonitoringEnabled() && !testResults.value?.monitoring_connection_result.connection_successful)) {
```

Then you can delete the last check from line 30 I think?

jpalac: You can ignore - it's just a refactor of what you had.

jpalac (approved):

jpalac (approved):

abparticular (approved):

PR #1972, opened 2024-07-24
grouping patch updates for npm

johnsimons:
We adding grouping for npm packages, so that it reduces the amount of PRs to deal with.
In this PR we are bundling all patch fixes in a single PR for all packages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,3 +39,10 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 1000
+  groups:
+    patch-updates:
+          applies-to: version-updates
+          update-types:
+          - "minor"
+          - "patch"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -39,3 +39,10 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 1000
+  groups:
+    patch-updates:
+          applies-to: version-updates
+          update-types:
+          - "minor"
+          - "patch"
+
```
jpalac, line 48: If we are only wanting to group patches, wouldn't we just have "patch" here?

johnsimons: oops, fixed @jpalac

jpalac (approved):

PR #1996, opened 2024-08-06
Read Custom.Build.props values before running minver

DavidBoike:


bording:
Since I've already been making changes to the release workflow in #1997, I've gone ahead and added this there already.

DavidBoike:
Well good since I already screwed it up here. :wink:

PR #1997, opened 2024-08-06
Add new host to be used for container

bording:
This PR makes the following changes:

- A new container host, based on Kestrel, has been introduced
  - For now, the new host is used only for the container and the PlatformSample.
  - The new host reads in the environment variables directly and uses the values to serve a dynamically constructed `app.contants.js` result, removing the need to write those out to disk before serving the website.
  - The new host also includes a reverse proxy so that all requests can go directly to the host's web application. API requests to ServiceControl and Monitoring APIs are forwarded by the revery proxy.
 - A new `MONITORING_URL` setting has been added to the container. It accepts a single URI instead of an array of URIs. If this new setting and the old `MONITORING_URLS` are both set, the new setting takes precedence.
- The container now runs processes internally as non-root, so the default port has been changed to `9090`.
- Some miscellaneous cleanup and improvements to other projects in the repo have also been included.


Container-README.md
```diff
@@ -68,9 +68,9 @@ The major version tag is never added to images pushed to [the GitHub Container R
 
 The latest release within a minor version will be tagged with `{major}.{minor}` on images pushed to Docker Hub. This allows users to target the latest patch within a specific minor version.
 
-## Built With
+## Image architecture
 
-This image is built from the stable Alpine version of the [nginx official Docker image](https://hub.docker.com/_/nginx/).
+This image is a multi-arch image based on the `mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled-composite` base image supporting `linux/arm64`, `linux/arm`, and `linux/amd64`.
```
DavidBoike, line 73: Right?
```suggestion
This image is a multi-arch image based on the `mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled-composite` base image supporting `linux/arm64` and `linux/amd64`.
```

src/Particular.PlatformSample.ServicePulse/Particular.PlatformSample.ServicePulse.csproj
```diff
@@ -1,28 +1,32 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net48</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
     <Description>Particular ServicePulse binaries for use by Particular.PlatformSample. Not intended for use outside of Particular.PlatformSample.</Description>
```
DavidBoike, line 7: I know it's outside the changeset, but does it still make sense to refer to the "binaries" here?

src/ServicePulse.Host.Tests/Api/APIApprovals.cs
```diff
@@ -1,13 +0,0 @@
-﻿using NUnit.Framework;
-using Particular.Approvals;
-using System.IO;
-
-class APIApprovals
-{
-    [Test]
-    public void PlatformSampleApprovals()
```
DavidBoike, line 8: Why isn't this necessary anymore?

DavidBoike (changes requested):

src/Particular.PlatformSample.ServicePulse/Particular.PlatformSample.ServicePulse.csproj
bording: If anything, I'd say that's more correct than it was before. Previously we were just packaging all the JS files, but now we're actually shipping the host binaries.

src/ServicePulse.Host.Tests/Api/APIApprovals.cs
bording: Because: 
`//HINT: If this test fails the Particular.PlatformSample project's app.constants.js probably needs to be updated`

Is no longer relevant. There will no longer need to be a separate `app.constants.js` in Particular.PlatformSample, so it can't get out of sync.

DavidBoike (approved):

boblangley (approved):

ivanl-out:
Thanks @bording.

Just note that the change from port 90 to 9090 was a breaking change for us since our ingress in Azure Container Apps previously exposed port 90. We also had a startup probe hitting port 90. It might be good to include this in the release notes for others (I may have missed this).

bording:
> Thanks @bording.
> 
> Just note that the change from port 90 to 9090 was a breaking change for us since our ingress in Azure Container Apps previously exposed port 90. We also had a startup probe hitting port 90. It might be good to include this in the release notes for others (I may have missed this).
> 
> Edit: I see the new port was noted in Container-README.md, so please disregard my comment, I just didn't read it since I was upgrading from a previous version.

We have added a note to the release notes and the announcement regarding the change.

However, the only thing that really changed was the port inside the container. Externally, you can continue to map that to whatever port you desire.

For example using docker:
```
docker run -p 90:9090 particular/servicepulse:latest
```

Would let you keep using 90 as the public-facing port.

Issue #1570, opened 2023-09-27
Edit and retry functionality can fail when multiple messages are correlated


cquirosj:
#### Description

While editing a message using the [edit and retry functionality](https://docs.particular.net/servicepulse/intro-editing-messages), if there are multiple correlated failed messages in the error queue, the editor dialog in ServicePulse could potentially load the incorrect message, making the editing procedure fail when the user submits the changes.

#### Expected behavior

When the user proceeds to edit the message: 
- An alert should get displayed in ServicePulse UI informing that is being retried: 'Retrying the message X'
- No warning errors being logged in ServiceControl related to the message just edited

#### Actual behavior

When the user proceeds to edit the message: 
- An error is displayed in ServicePulse UI with the following description: An error occurred while retrying the message, please check the ServiceControl logs for more details on the failure. 
- ServiceControl logs the following warning message: Warn|ServiceControl.MessageFailures.Api.EditFailedMessagesController|Locked headers have been modified on the edit-retry for MessageID

#### Versions
This is confirmed in version v1.32.4 and the latest at the moment of recording this bug( v1.37.0) This could be happening in all versions of ServicePulse, but more versions need to be checked to confirm this statement.

### Steps to reproduce

- Create an endpoint A that publishes an event.
- Create two different endpoints, Endpoint B and Endpoint C, and have them subscribe to the event published by Endpoint A. 
- Make the event handlers for each endpoint throw an exception.
- Configure the endpoints to use the platform.
- Using ServicePulse, inspect each failed message and locate the message with an incorrect value being displayed for the `NServiceBus.OriginatingEndpoint` header. For example, if the failed message was handled by endpoint A but `NServiceBus.OriginatingEndpoint` header shows endpoint B as the header's value.
- Use the edit and retry button to edit the body and click the retry button.

As an alternative, use the Saga Timeouts [sample](https://docs.particular.net/nservicebus/sagas/timeouts) and update OrderPlacedHandler.cs in the billing endpoint and ShippingPolicy.cs in the shipping endpoint to throw an exception where the OrderPlaced event is being handled

### Relevant log output

```shell
Warn|ServiceControl.MessageFailures.Api.EditFailedMessagesController|Locked headers have been modified on the edit-retry for MessageID
```

nodranael:
Hello,

Is there progress on this issue ? This was created through customer support, and my company relies on this feature.

lailabougria:
[...]

PR #2001, opened 2024-08-12
Filter message on endpoint address

johnsimons:
Fixes https://github.com/Particular/ServicePulse/issues/1570

### DIFF
--- a/src/Frontend/src/components/failedmessages/MessageView.vue
+++ b/src/Frontend/src/components/failedmessages/MessageView.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { computed, onMounted, onUnmounted, ref, watch } from "vue";
 import { RouterLink, useRoute } from "vue-router";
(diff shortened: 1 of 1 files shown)


jpalac (approved):

src/Frontend/src/components/failedmessages/MessageView.vue
```diff
@@ -123,14 +122,15 @@ async function downloadHeadersAndBody(message: ExtendedFailedMessage) {
   try {
     const [, data] = await useTypedFetchFromServiceControl<Message[]>(`messages/search/${message.message_id}`);
 
-    if (data[0] == null) {
+    const messageDetails = data.find((value) => value.receiving_endpoint.name === message.receiving_endpoint?.name);
```
johnsimons, line 125: This is the actual fix, instead of just taking the first message, we pick the correct one based on `receiving_endpoint` address

PhilBastian (approved):

PR #2038, opened 2024-09-03
create new control based on on/off switch for toggling tracking value

PhilBastian:


johnsimons:
So, based on the conversation yesterday, we may not need this.
Is this your take too @PhilBastian ?

PhilBastian:
> So, based on the conversation yesterday, we may not need this. Is this your take too @PhilBastian ?

I will cherry-pick the changes not directly related to ValueToggle control to a new PR

PR #2056, opened 2024-09-12
Elastic scale patch

PhilBastian:
fix results count not showing and enable deletion of non-tracked endpoints

### DIFF
--- a/src/Frontend/src/components/heartbeats/EndpointInstances.vue
+++ b/src/Frontend/src/components/heartbeats/EndpointInstances.vue
@@ -89,7 +89,7 @@ async function deleteInstance(instance: EndpointsView) {
 
 async function deleteAllInstances() {
   try {
-    await Promise.all(sortedInstances.value.filter((instance) => instance.name === endpointName).map((instance) => await store.deleteEndpointInstance(instance)));
+    await Promise.all(sortedInstances.value.filter((instance) => instance.name === endpointName).map((instance) => store.deleteEndpointInstance(instance)));
     useShowToast(TYPE.SUCCESS, "Endpoint deleted", "", false, { timeout: 1000 });
     await router.replace(backLink.value);
   } catch {
--- a/src/Frontend/src/components/heartbeats/EndpointInstances.vue
+++ b/src/Frontend/src/components/heartbeats/EndpointInstances.vue
@@ -26,12 +26,13 @@ enum Operation {
 }
 
 const route = useRoute();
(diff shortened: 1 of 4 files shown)


src/Frontend/src/components/heartbeats/HealthyEndpoints.vue
```diff
@@ -3,6 +3,7 @@ import NoData from "../NoData.vue";
 import { ColumnNames, useHeartbeatsStore } from "@/stores/HeartbeatsStore";
 import { storeToRefs } from "pinia";
 import HeartbeatsList from "./HeartbeatsList.vue";
+import ResultsCount from "../ResultsCount.vue";
```
johnsimons, line 6: This is really weird, how come this did not break the build?

src/Frontend/src/components/heartbeats/EndpointInstances.vue
```diff
@@ -84,6 +87,16 @@ async function deleteInstance(instance: EndpointsView) {
   }
 }
 
+async function deleteAllInstances() {
+  try {
+    await Promise.all(sortedInstances.value.filter((instance) => instance.name === endpointName).map((instance) => deleteInstance(instance)));
+    useShowToast(TYPE.SUCCESS, "Endpoint instance deleted", "", false, { timeout: 1000 });
```
johnsimons, line 93: Should this say
```suggestion
    useShowToast(TYPE.SUCCESS, "Endpoint deleted", "", false, { timeout: 1000 });
```

johnsimons (changes requested):

johnsimons (changes requested):
I have tested this and a few things that don't look right.
- I think the "delete button" should have a warning dialog given it is a dangerous action, thoughts?
- When the deletion occurs, we get multiple toasters, I assume one per instance, to me it would make more sense to be a single toaster

PhilBastian:
> I have tested this and a few things that don't look right.
> 
> * I think the "delete button" should have a warning dialog given it is a dangerous action, thoughts?
> * When the deletion occurs, we get multiple toasters, I assume one per instance, to me it would make more sense to be a single toaster

the other delete buttons don't give a warning
there should only be one toast message. I wonder if it's something to do with the route change?

johnsimons:
> the other delete buttons don't give a warning

the others are single this one is bulk delete

johnsimons (approved):
I don't think we need the warning dialog for delete endpoint.

PR #2060, opened 2024-09-16
Upgrading to latest version of eslint

johnsimons:
The latest version of `eslint` has changed the way it is configured. It now uses [flat config](https://eslint.org/blog/2023/10/flat-config-rollout-plans/)


src/Frontend/.eslintrc.cjs
```diff
@@ -1,33 +0,0 @@
...
-  extends: ["plugin:vue/vue3-essential", "eslint:recommended", "plugin:@typescript-eslint/recommended", "@vue/eslint-config-prettier/skip-formatting"],
-  parser: "vue-eslint-parser",
-  plugins: ["@typescript-eslint", "github", "import", "unused-imports", "prefer-arrow"],
-  parserOptions: {
-    ecmaVersion: "latest",
-    parser: "@typescript-eslint/parser",
-    extraFileExtensions: [".vue"],
-  },
-  rules: {
-    "prettier/prettier": "error",
```
johnsimons, line 20: replaced by eslint-config-prettier

src/Frontend/.eslintrc.cjs
```diff
@@ -1,33 +0,0 @@
...
-    "require-await": "error",
-    "no-await-in-loop": "error",
-    "github/no-then": "error",
```
johnsimons, line 23: couldn't find repacement

src/Frontend/.eslintrc.cjs
```diff
@@ -1,33 +0,0 @@
...
-    "prefer-rest-params": "error",
-    "prefer-spread": "error",
-    "no-var": "error",
-    "prefer-const": "error",
-    eqeqeq: ["error", "smart"],
-    "unused-imports/no-unused-imports": "error",
-    "import/no-duplicates": "error",
```
johnsimons, line 30: couldn't find replacement

src/Frontend/package.json
```diff
@@ -9,62 +9,59 @@
     "dev:mocks": "cross-env NODE_ENV=dev-mocks vite",
     "build": "vite build",
     "type-check": "vue-tsc --build --force",
-    "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --ignore-path .gitignore",
```
johnsimons, line 12: New version now respecs ignored files, and extension no longer needed

src/Frontend/package.json
```diff
@@ -9,62 +9,59 @@
...
+    "vue-router": "^4.4.5",
     "vue-tippy": "^6.4.4",
     "vue-toastification": "^2.0.0-rc.5",
     "vue3-cookies": "^1.0.6",
     "vue3-simple-typeahead": "^1.0.11"
   },
   "devDependencies": {
+    "@eslint/js": "^9.10.0",
     "@pinia/testing": "^0.1.5",
-    "@rushstack/eslint-patch": "^1.10.4",
```
johnsimons, line 38: not required

src/Frontend/.eslintrc.cjs
PhilBastian: i would suggest using the deprecated eslintrc until there is a replacement for these

johnsimons: We can't, we need to move out of it.
image

And those rules are not needed, we have moved to async await now.

PhilBastian: those rules are for ensuring future work follows the async/await paradigm. The blog says that eslintrc will not be maintained, but also won't be removed. I'm not suggesting staying on eslintrc long term, just until all the rules we want to enforce are supported

johnsimons: the problem is that those are not going to be updated, so we do need to look for replacements

johnsimons: replaced with https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md

src/Frontend/.eslintrc.cjs
johnsimons: replaced with https://eslint.org/docs/latest/rules/no-duplicate-imports

johnsimons:
Is this ok now @PhilBastian ?

PhilBastian:
i guess so? 🤷

PhilBastian (approved):

PR #2066, opened 2024-09-18
Add PostgreSQL to Usage

jpalac:
Implementation of https://github.com/Particular/ServiceControl/issues/4254

### DIFF
--- a/src/Frontend/src/views/throughputreport/transport.ts
+++ b/src/Frontend/src/views/throughputreport/transport.ts
@@ -7,4 +7,5 @@ export enum Transport {
   RabbitMQ = "RabbitMQ",
   SQLServer = "SqlTransport",
   AmazonSQS = "AmazonSQS",
+  PostgreSQL = "PostgreSql"
 }
--- a/src/Frontend/src/views/throughputreport/transport.ts
+++ b/src/Frontend/src/views/throughputreport/transport.ts
@@ -7,5 +7,5 @@ export enum Transport {
   RabbitMQ = "RabbitMQ",
   SQLServer = "SqlTransport",
   AmazonSQS = "AmazonSQS",
-  PostgreSQL = "PostgreSql"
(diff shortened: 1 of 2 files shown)


src/Frontend/src/views/throughputreport/transport.ts
```diff
@@ -7,4 +7,5 @@ export enum Transport {
   RabbitMQ = "RabbitMQ",
   SQLServer = "SqlTransport",
   AmazonSQS = "AmazonSQS",
+  PostgreSQL = "PostgreSql",
```
johnsimons, line 10: I assume this matches with the key in the report in SC?

johnsimons (approved):

src/Frontend/src/views/throughputreport/transport.ts
andreasohlund, line 10: Nitpick: The spelling seems to be "PostgreSQL" (https://www.postgresql.org/) should we align with that?

jpalac: We do - the left hand side is the enum value and the right hand side is whats coming from SC so we are using that just to match.

andreasohlund: Ah, should we adjust the sc side in that case to be consistent?

jpalac: It's internal only - in line with the project name.
I don't mind. Sql transport does something similar where the name of project etc uses Sql, then it shows externally as SQL So in the postgres naming we followed that pattern.
I'm fine either way.

andreasohlund: Ok, fair enough 👍

PhilBastian (approved):

SzymonPobiega (approved):

andreasohlund (approved):

PR #2103, opened 2024-10-10
A few small fixes

johnsimons:
I have decided to skip a random failing test for now, see https://github.com/Particular/ServicePulse/actions/runs/11296984618/job/31422942849?pr=2104#step:7:780

I have also updated the location of the docker file that dependbot looks for.


johnsimons (commented):
kjk

cquirosj (approved):

PR #2108, opened 2024-10-15
Using license extension URL from the ServiceControl API

tmasternak:


src/Frontend/src/resources/LicenseInfo.ts
```diff
@@ -7,6 +7,7 @@ export default interface LicenseInfo {
   instance_name: string;
   trial_license: boolean;
   license_status: LicenseStatus;
+  license_extension_url: string;
```
johnsimons, line 10: it seems this new prop should be nullable?
```suggestion
  license_extension_url?: string;
```
which I assume will have repercussions in other areas of the code

tmasternak: I might be missing something but the property doesn't need to be nullable. It has a well-defined default value which can be overridden by a value coming from the backend.

src/Frontend/test/preconditions/hasLicense.ts
```diff
@@ -37,6 +40,9 @@ const createLicenseMockedResponse =
         break;
     }
 
+    //We need to reset the global state to ensure the waring toast is always triggered by the value changing between multiple test runs. See documented issue and proposed solution https://github.com/Particular/ServicePulse/issues/1905
```
WilliamBZA, line 43:
```suggestion
    //We need to reset the global state to ensure the warning toast is always triggered by the value changing between multiple test runs. See documented issue and proposed solution https://github.com/Particular/ServicePulse/issues/1905
```

src/Frontend/src/resources/LicenseInfo.ts
WilliamBZA: The `LicenseInfo` type is retrieved by the call to `getLicense()` in the [serviceLicense](https://github.com/Particular/ServicePulse/pull/2108/files#diff-9aabc6274887bc0a98cc4eac00ef971b62366e71b4723aad50a4f14abcd391bbR147) file. This object is the evaluated in the `getOrUpdateLicenseStatus()` method, which is where the well-defined default value is provided.

So I think John is correct, the value here could be null.

WilliamBZA (approved):

src/Frontend/src/components/LicenseNotifications.vue
```diff
@@ -18,7 +19,7 @@ function displayWarningMessage(licenseStatus: LicenseStatus) {
       break;
     }
     case "ValidWithExpiringTrial": {
-      const trialExpiring = `<div><strong>Non-production development license expiring</strong><div>Your non-production development license will expire soon. To continue using the Particular Service Platform you'll need to extend your license.</div><a href="http://particular.net/extend-your-trial?p=servicepulse" class="btn btn-warning"><i class="fa fa-external-link-alt"></i> Extend your license</a><a href="${configurationRootLink}" class="btn btn-light">View license details</a></div>`;
+      const trialExpiring = `<div><strong>Non-production development license expiring</strong><div>Your non-production development license will expire soon. To continue using the Particular Service Platform you'll need to extend your license.</div><a href="${license.license_extension_url}" class="btn btn-warning"><i class="fa fa-external-link-alt"></i> Extend your license</a><a href="${configurationRootLink}" class="btn btn-light">View license details</a></div>`;
```
johnsimons, line 22: Something feels off here, we shouldn’t be writing html as strings, can this be extracted to a component?

src/Frontend/src/components/configuration/PlatformLicense.vue
```diff
@@ -62,7 +62,7 @@ const loading = computed(() => {
                     </span>
                     <div class="license-expired-text" v-if="licenseStatus.isPlatformTrialExpired">Your license expired. To continue using the Particular Service Platform you'll need to extend your license.</div>
                     <div class="license-page-extend-trial" v-if="licenseStatus.isPlatformTrialExpiring && licenseStatus.isPlatformTrialExpired">
-                      <a class="btn btn-default btn-primary" href="https://particular.net/extend-your-trial?p=servicepulse" target="_blank">Extend your license&nbsp;&nbsp;<i class="fa fa-external-link"></i></a>
+                      <a class="btn btn-default btn-primary" :href="license.license_extension_url" target="_blank">Extend your license&nbsp;&nbsp;<i class="fa fa-external-link"></i></a>
```
johnsimons, line 65: It may be the right time to create an ExternalLink component

src/Frontend/src/components/LicenseNotifications.vue
tmasternak: I've crated a dedicated refactoring issue to capture that https://github.com/Particular/ServicePulse/issues/2116

src/Frontend/src/components/configuration/PlatformLicense.vue
tmasternak: I'm missing a lot of context when in comes to Vue.js. Could you explain what would be the benefit of introducing an ExternalLink component?

johnsimons: At the moment a dev needs to know all the styles, the external icon, the spaces between the icon and text ….
Those are all internal concerns. 
With something like
`<ExternalLink href=x />` they don’t need to know about those concerns. Also that makes it easier to update the external link look and feel in a single component. 
In webapps now a days with ui frameworks most html is abstracted away.

tmasternak: Makes sense. I've captured that in an issue https://github.com/Particular/ServicePulse/issues/2117

PR #2209, opened 2025-01-08
Check for the license type on Configuration screen

poornimanayar:


soujay:
Closing this PR as we have [another](https://github.com/Particular/ServicePulse/pull/2227) one that covers both expired and expiring licenses

PR #2216, opened 2025-01-20
Add custom checks acceptance tests

soujay:
Adding some mock data and tests to Custom checks


src/Frontend/test/preconditions/customChecks.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (status === "fail") {
+      return {
+        body: failedCustomCheckItems,
+        headers: { "Total-Count": failedCustomCheckItems.length.toString() },
+      };
+    }
+
+    return {
+      body: passedCustomCheckItems,
+      headers: { "Total-Count": passedCustomCheckItems.length.toString() },
```
cquirosj, line 31: I believe ServiceControl returns all checks, including passing and not passing checks, when the /customchecks path is accessed without the `status=fail` in the query string.

src/Frontend/test/specs/customchecks/viewing-no-data.spec.ts
```diff
@@ -1,21 +1,47 @@
...
 
 describe("FEATURE: No data", () => {
   describe("RULE: When there is no data to show a message should be displayed ", () => {
-    test.todo("EXAMPLE: 'No failed custom checks' should be displayed when there are no custom checks");
-
     /* SCENARIO
           Given there are no custom checks
           When navigating to the custom checks tab
           Then a message is shown "No failed custom checks"
         */
```
cquirosj, line 13: Once the test code is implemented, I recommend removing these comments to avoid potential discrepancies, as the test code itself serves as the ultimate source of truth. If comments seem necessary, it might indicate that the test code is not effectively functioning as a DSL (Domain-Specific Language). Ideally, the test code should be designed to read almost like prose, making its intent clear without requiring additional explanation.

src/Frontend/test/specs/customchecks/viewing-no-data.spec.ts
```diff
@@ -1,21 +1,47 @@
...
+      });
+    });
     /* SCENARIO
           Given there are custom checks
           And all custom checks are in a success state
           When navigating to the custom checks tab
-          The a message is shown "No failed custom checks"
+          Then a message is shown "No failed custom checks"
         */
+    test("EXAMPLE: 'No failed custom checks' should be displayed when all custom checks are in a success state", async ({ driver }) => {
```
cquirosj, line 33:
```suggestion
    test("EXAMPLE: There are custom checks but none of them are failing", async ({ driver }) => {
```

src/Frontend/test/specs/customchecks/viewing-no-data.spec.ts
```diff
@@ -1,21 +1,47 @@
...
 describe("FEATURE: No data", () => {
   describe("RULE: When there is no data to show a message should be displayed ", () => {
-    test.todo("EXAMPLE: 'No failed custom checks' should be displayed when there are no custom checks");
-
     /* SCENARIO
           Given there are no custom checks
           When navigating to the custom checks tab
           Then a message is shown "No failed custom checks"
         */
+    test("EXAMPLE: 'No failed custom checks' should be displayed when there are no custom checks", async ({ driver }) => {
```
cquirosj, line 14:
```suggestion
    test("EXAMPLE: There are no failed or passing custom checks", async ({ driver }) => {
```

src/Frontend/test/specs/customchecks/viewing-no-data.spec.ts
```diff
@@ -1,21 +1,47 @@
...
           And all custom checks are in a success state
           When navigating to the custom checks tab
-          The a message is shown "No failed custom checks"
+          Then a message is shown "No failed custom checks"
         */
+    test("EXAMPLE: 'No failed custom checks' should be displayed when all custom checks are in a success state", async ({ driver }) => {
+      //Arrange
+      await driver.setUp(precondition.serviceControlWithMonitoring);
+      // given that the custom check list is not empty and status of all checks are passing
+      await driver.setUp(precondition.hasCustomChecksPassing);
```
cquirosj, line 37: To help detect potential regressions in the future, the backend should return a combination of both passing and failing custom checks when the URL being hit does not include status=fail. This approach ensures that the test will fail if:

- A bug is introduced causing the implementation to stop using status=fail.
- The filtering of custom checks is shifted from the server to the browser but the new implementation fails to exclude the passing checks.

cquirosj (approved):
This is a solid first step! While I see several areas that could be refined further, this moves us in the right direction, and we can build on it in future iterations. Approving this now — great work!

PR #2220, opened 2025-01-23
MassTransit retry functionality

PhilBastian:


irinascurtu (approved):

src/Frontend/src/components/failedmessages/EditRetryDialog.vue
```diff
@@ -106,7 +106,7 @@ function isContentTypeSupported(contentType: string) {
 
 function getMessageIntent() {
   const intent = findHeadersByKey("NServiceBus.MessageIntent");
-  return intent.value;
+  return intent?.value ?? "";
```
johnsimons, line 109: This looks weird, @PhilBastian; why this change?
I looked more closely at this component, and it is totally untyped, which is scary!
I also looked at `getContentType()`, which is quite similar to this function.
So, let's add types to this component to be on the safe side.

src/Frontend/src/composables/connectedApplications.ts
johnsimons, line 1: This file should be renamed

src/Frontend/src/composables/connectedApplications.ts
```diff
@@ -0,0 +1,7 @@
+import { computed } from "vue";
+import { useConfiguration } from "./configuration";
+
+export function useIsMassTransitConnected() {
+  const configuration = useConfiguration();
+  return computed(() => configuration.value?.mass_transit_connector != null);
```
johnsimons, line 6:
```suggestion
  return computed(() => configuration.value?.mass_transit_connector !== undefined);
```

src/Frontend/src/resources/LicenseInfo.ts
```diff
@@ -11,6 +13,12 @@ export default interface LicenseInfo {
   status: string;
 }
 
+export function typeText(license: LicenseInfo, configuration: Configuration | null) {
+  if (license.license_type === "Non-Production Development" && configuration?.mass_transit_connector) {
```
johnsimons, line 17:
```suggestion
  if (license.trial_license && configuration?.mass_transit_connector) {
```

src/Frontend/src/views/DashboardView.vue
johnsimons, line 1: What are these changes related to?

johnsimons (changes requested):
A few comments, and the proxy code is failing

src/Frontend/src/components/failedmessages/EditRetryDialog.vue
johnsimons: @PhilBastian I refactor this and added types, can you review those changes

src/Frontend/src/views/DashboardView.vue
johnsimons, line 1: What are these changes related to?

johnsimons (approved):
From my side these changes look good.
Pending a review from @PhilBastian on the refactor to add types.

src/Frontend/src/composables/connectedApplications.ts
PhilBastian: standard practice is to use != null to check for `null` _or_ `undefined`

src/Frontend/src/views/DashboardView.vue
PhilBastian: getting it to display correct should we choose to hide the heartbeats icon

PR #2259, opened 2025-02-24
Improving provenance of image

johnsimons:
Using native action to login to ghcr.io.
Replaced multiple tagging commands with a single one Using build-push-action instead of script.
This adds the provenance by default, see https://docs.docker.com/build/ci/github-actions/attestations/#default-provenance .

This gives the image a better score in DockerHub Scout health score.
image

This is the same changes as https://github.com/Particular/ServiceControl.Connector.MassTransit/pull/182.


johnsimons:
This is what looks like in the connector
image

PhilBastian (approved):

PR #2273, opened 2025-03-13
Reword the heading of the popup message in retry dialog

soujay:
Reword the heading  of the retry popup dialog to a descriptive text

### DIFF
--- a/src/Frontend/src/components/messages/MessageView.vue
+++ b/src/Frontend/src/components/messages/MessageView.vue
@@ -388,8 +388,8 @@ onUnmounted(() => {
                 showRetryConfirm = false;
                 retryMessage();
               "
-              :heading="'Retry Message'"
-              :body="'Are you sure you want to retry this message?'"
+              heading="Retry Message"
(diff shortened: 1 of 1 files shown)


src/Frontend/src/components/failedmessages/MessageView.vue
```diff
@@ -496,7 +496,7 @@ onUnmounted(() => {
                 retryMessage();
               "
               :heading="'Are you sure you want to retry this message?'"
-              :body="'Are you sure you want to retry this message?'"
+              :body="'Ensure that the selected message was not processed previously as this will create a duplicate message.'"
```
johnsimons, line 499: I am a bit hesitant to change it to this.
The reason is that we want to help customers fall into the pit of success, so asking them to do this feels a bit wrong.
We should be doing all this to ensure they do not run into issues.

johnsimons: @PhilBastian proposed to remove the body section of the dialog is body is not set

johnsimons (approved):

PR #2281, opened 2025-03-16
Copy feature in message headers in message view

soujay:
This PR adds a copy icon next to each message header, enabling the copy of the header value to the clipboard.

### DIFF
--- a/src/Frontend/src/components/CopyToClipboard.vue
+++ b/src/Frontend/src/components/CopyToClipboard.vue
@@ -2,9 +2,16 @@
 import { Tippy, TippyComponent } from "vue-tippy";
 import { useTemplateRef } from "vue";
 
-const props = defineProps<{
-  value: string;
-}>();
+const props = defineProps({
+  value: {
+    type: String,
+    required: true,
+  },
+  isIconOnly: {
+    type: Boolean,
+    default: false, // Default value
+  },
+});
 
 const tippyRef = useTemplateRef<TippyComponent | null>("tippyRef");
 let timeoutId: number;
@@ -19,6 +26,6 @@ async function copyToClipboard() {
 
 <template>
   <Tippy content="Copied" ref="tippyRef" trigger="manual">
-    <button type="button" class="btn btn-secondary btn-sm" @click="copyToClipboard"><i class="fa fa-copy"></i> Copy to clipboard</button>
+    <button type="button" class="btn btn-secondary btn-sm" @click="copyToClipboard"><i class="fa fa-copy"></i> <span v-if="!props.isIconOnly">Copy to clipboard</span></button>
   </Tippy>
 </template>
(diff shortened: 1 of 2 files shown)


src/Frontend/src/components/CopyToClipboard.vue
```diff
@@ -18,7 +25,7 @@ async function copyToClipboard() {
 </script>
 
 <template>
-  <Tippy content="Copied" ref="tippyRef" trigger="manual">
-    <button type="button" class="btn btn-secondary btn-sm" @click="copyToClipboard"><i class="fa fa-copy"></i> Copy to clipboard</button>
+  <Tippy content="Copied" ref="tippyRef" trigger="manual" :title="props.isIconOnly ? 'Copy to Clipboard' : ''">
```
PhilBastian, line 28: does this show an empty tooltip, rather than no tooltip? Perhaps it should be null rather than ''?

src/Frontend/src/components/messages/HeadersView.vue
```diff
@@ -12,12 +19,21 @@ const props = defineProps<{
...
   </table>
   <div v-if="props.message.headersNotFound" class="alert alert-info">Could not find message headers. This could be because the message URL is invalid or the corresponding message was processed and is no longer tracked by ServiceControl.</div>
 </template>
 
-<style scoped></style>
+<style scoped>
+.headercopy {
+  display: flex;
+  align-items: top;
+  gap: 10px;
```
PhilBastian, line 37: we want this to scale with fontsize/zoom so should be in rem

src/Frontend/src/components/CopyToClipboard.vue
```diff
@@ -2,9 +2,16 @@
...
+const props = defineProps({
+  value: {
+    type: String,
+    required: true,
+  },
+  isIconOnly: {
+    type: Boolean,
+    default: false, // Default value
+  },
+});
```
PhilBastian, line 14: default values are handled differently elsewhere in SP

```suggestion
const props = defineProps<{
    value: string,
    isIconOnly?: boolean
}().withDefaults({ isIconOnly: false })
```

src/Frontend/src/components/messages/HeadersView.vue
```diff
@@ -12,12 +19,21 @@ const props = defineProps<{
       <tr class="interactiveList" v-for="(header, index) in props.message.headers" :key="index">
         <td nowrap="nowrap">{{ header.key }}</td>
         <td>
-          <pre>{{ header.value }}</pre>
+          <div class="headercopy" @mouseover="toggleHover(index, true)" @mouseleave="toggleHover(index, false)">
+            <pre>{{ header.value }}</pre>
+            <CopyToClipboard v-if="hoverStates[index]" :value="header.value || ''" :isIconOnly="true" />
```
PhilBastian, line 24: this doesn't work with keyboard navigation

PhilBastian (changes requested):

johnsimons (changes requested):

johnsimons (approved):

PhilBastian (approved):

PR #2284, opened 2025-03-17
Fixed flickering on flow diagram initial render

johnsimons:
Also added controls to zoom and fit.


johnsimons:
@PhilBastian this seems to fix the flicker, can you check it out

PhilBastian (approved):

PR #2290, opened 2025-03-18
Change to useTemplateRef for refs

johnsimons:
As per https://vuejs.org/guide/essentials/template-refs

I noticed this as part of the current work we are doing.

### DIFF
--- a/src/Frontend/src/components/failedmessages/DeletedMessageGroups.vue
+++ b/src/Frontend/src/components/failedmessages/DeletedMessageGroups.vue
@@ -40,7 +40,6 @@ interface ExtendedFailureGroupView extends FailureGroupView {
 }
 
 let pollingFaster = false;
-const messageGroupList = ref();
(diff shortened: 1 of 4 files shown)


src/Frontend/src/components/failedmessages/DeletedMessageGroups.vue
```diff
@@ -68,12 +67,12 @@ function saveDefaultGroupingClassifier(classifier: string) {
   cookies.set("archived_groups_classification", classifier);
 }
 
-function classifierChanged(classifier: string) {
+async function classifierChanged(classifier: string) {
   saveDefaultGroupingClassifier(classifier);
 
   selectedClassifier.value = classifier;
   archiveGroups.value = [];
-  messageGroupList.value = loadArchivedMessageGroups(classifier);
```
johnsimons, line 76: `messageGroupList` was not being used as a `ref` nor for anything given that `loadArchivedMessageGroups(classifier)` returns void!

PhilBastian (approved):

PR #2291, opened 2025-03-18
Turn on wordwrap on code editor

johnsimons:
This fixes issue with overflow in dialogs

### DIFF
--- a/src/Frontend/src/components/CodeEditor.vue
+++ b/src/Frontend/src/components/CodeEditor.vue
@@ -53,7 +53,7 @@ const extensions = computed(() => {
     <div v-if="props.showCopyToClipboard" class="toolbar">
       <CopyToClipboard :value="code" />
     </div>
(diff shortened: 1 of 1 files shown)


src/Frontend/src/components/CodeEditor.vue
```diff
@@ -53,7 +53,7 @@ const extensions = computed(() => {
     <div v-if="props.showCopyToClipboard" class="toolbar">
       <CopyToClipboard :value="code" />
     </div>
-    <CodeMirror v-model="code" :extensions="extensions" :basic="props.showGutter" :minimal="!props.showGutter" :readonly="props.readOnly" :gutter="!props.readOnly"></CodeMirror>
+    <CodeMirror v-model="code" :extensions="extensions" :basic="props.showGutter" :minimal="!props.showGutter" :readonly="props.readOnly" :gutter="!props.readOnly" :wrap="true"></CodeMirror>
```
PhilBastian, line 56: should this be put on all usages of CodeMirror, so that in future if people copy it from a page other than this one they get this same behaviour?

PR #2300, opened 2025-04-03
Replace the message header search with the filter component 

soujay:
Replace the message header search with the filter component


src/Frontend/src/components/messages/HeadersView.vue
```diff
@@ -26,7 +27,9 @@ const filteredHeaders = computed(() => {
       <div class="col">
         <div class="text-search-container">
           <div class="text-search">
-            <input type="search" aria-label="Filter by name" v-model="searchTerm" class="form-control format-text" placeholder="Search for a header key or value..." />
+            <div class="filter-group">
+              <FilterInput v-model="searchTerm" />
```
johnsimons, line 31: @soujay I think we need to change the `FilterInput` component to be able to set `placeholder` and `aria-label`
So the code here should be something like
```
<FilterInput v-model="searchTerm" placeholder="Search for a header key or value..." aria-label="....." />
```

PhilBastian (approved):

PR #2302, opened 2025-04-03
Sequence diagram first part

PhilBastian:
right flow only, and no message selection


src/Frontend/src/components/heartbeats/EndpointInstances.vue
```diff
@@ -89,7 +89,7 @@ async function deleteInstance(instance: EndpointsView) {
 
 async function deleteAllInstances() {
   try {
-    await Promise.all(sortedInstances.value.filter((instance) => instance.name === endpointName).map((instance) => store.deleteEndpointInstance(instance)));
+    await Promise.all(sortedInstances.value.filter((instance: EndpointsView) => instance.name === endpointName).map((instance: EndpointsView) => store.deleteEndpointInstance(instance)));
```
johnsimons, line 92: Is this really needed?
Isn't the array typed already?

src/Frontend/src/components/heartbeats/HeartbeatConfiguration.vue
```diff
@@ -35,7 +36,7 @@ async function proceedWarningDialog() {
 
   try {
     await store.updateEndpointSettings(
-      filteredEndpoints.value.filter((endpoint) => (dialogWarningOperation.value === Operation.Track && !endpoint.track_instances) || (dialogWarningOperation.value === Operation.DoNotTrack && endpoint.track_instances))
+      filteredEndpoints.value.filter((endpoint: LogicalEndpoint) => (dialogWarningOperation.value === Operation.Track && !endpoint.track_instances) || (dialogWarningOperation.value === Operation.DoNotTrack && endpoint.track_instances))
```
johnsimons, line 39: Is this really needed?
Isn't the array typed already?

src/Frontend/src/components/messages/SequenceDiagram.vue
```diff
@@ -0,0 +1,46 @@
...
+
+const store = useSequenceDiagramStore();
+store.setConversationId("39907d51-12e5-4202-82c3-b2b30077ebd4");
+
+const { maxWidth, maxHeight } = storeToRefs(store);
+</script>
+
+<template>
+  <div class="outer">
+    <svg class="sequence-diagram" :width="`max(100%, ${isNaN(maxWidth) ? 0 : maxWidth}px)`" :height="maxHeight + 20">
```
johnsimons, line 17: Should we be rendering at all if `isNaN(maxWidth)` is `true`?
I have created a [spinner](https://github.com/Particular/ServicePulse/pull/2311/files#diff-fd610f09e5530881446332f2c2ba6694b7fdbe268db643e96baed8f251fd3415) for exactly this purpose?

src/Frontend/src/components/messages/SequenceDiagram/EndpointsComponent.vue
```diff
@@ -0,0 +1,103 @@
...
+      :width="endpoint.surround.width"
+      :height="endpoint.surround.height"
+      :fill="endpoint.surround.fill"
+      :rx="endpoint.surround.rx"
+      :stroke-width="endpoint.surround.strokeWidth"
+      :stroke="endpoint.surround.stroke"
+    ></rect>
+    <g :transform="`translate(${(endpoint.x ?? Endpoint_Width / 2) - ((endpoint.textWidth ?? 0) + Endpoint_Image_Width) / 2},0)`">
+      <path fill="var(--gray40)" d="M 0,0 M 18,18 M 0,2 v 14 h 14 v -4 h -6 v -6 h 6 v -4 h -14 M 9,7 v 4 h 9 v -4"></path>
+      <text :x="Endpoint_Image_Width" y="10" alignment-baseline="middle" text-anchor="start" :ref="(el) => setEndpointRef(el as SVGTextElement, i)">{{ endpoint.name }}</text>
```
johnsimons, line 92: The ref function [on unmount](https://vuejs.org/guide/essentials/template-refs.html#function-refs) passes null as the `el`, is that an issue?

src/Frontend/src/components/messages/SequenceDiagram/EndpointsComponent.vue
```diff
@@ -0,0 +1,103 @@
...
+      :x="endpoint.surround.x"
+      :y="endpoint.surround.y"
+      :width="endpoint.surround.width"
+      :height="endpoint.surround.height"
+      :fill="endpoint.surround.fill"
+      :rx="endpoint.surround.rx"
+      :stroke-width="endpoint.surround.strokeWidth"
+      :stroke="endpoint.surround.stroke"
+    ></rect>
+    <g :transform="`translate(${(endpoint.x ?? Endpoint_Width / 2) - ((endpoint.textWidth ?? 0) + Endpoint_Image_Width) / 2},0)`">
```
johnsimons, line 90:
```suggestion
    <g :transform="`translate(${(endpoint.x ?? Endpoint_Width / 2) - ((endpoint.textWidth ?? 0) + Endpoint_Image_Width) / 2}, 0)`">
```

src/Frontend/src/components/messages/SequenceDiagram/HandlersComponent.vue
```diff
@@ -0,0 +1,97 @@
...
+  return result;
+});
+
+function setMessageTypeRef(el: SVGTextElement, index: number) {
+  if (el) messageTypeRefs.value[index] = el;
+}
+</script>
+
+<template>
+  <g>
```
johnsimons, line 72: Why d we need a wrapper `g`?

src/Frontend/src/components/messages/SequenceDiagram/HandlersComponent.vue
```diff
@@ -0,0 +1,97 @@
...
+      icon,
+      iconSize,
+      messageType: handler.name,
+      messageTypeOffset: handler.direction === Direction.Right ? ((messageTypeElement?.getBBox().width ?? 0) + 24) * -1 : 20,
+      messageTypeHighlight: handler.route?.name === highlightId.value,
+    };
+  });
+
+  store.setMaxHeight(nextY);
+  store.setHandlerLocations(result.map((handler) => ({ id: handler.id, left: handler.left, right: handler.right, y: handler.y, height: handler.height })));
```
johnsimons, line 61: This feels like a side effect.
Wondering if there is a better way to do this?

src/Frontend/src/components/messages/SequenceDiagram/RoutesComponent.vue
```diff
@@ -0,0 +1,108 @@
...
+  })
+);
+
+function setMessageTypeRef(el: SVGTextElement, index: number) {
+  if (el) messageTypeRefs.value[index] = el;
+}
+</script>
+
+<template>
+  <template v-for="(arrow, i) in arrows" :key="arrow?.id">
```
johnsimons, line 64: Why is arrow null? Should we filter it out as part of the computed?

src/Frontend/src/components/messages/SequenceDiagram/RoutesComponent.vue
```diff
@@ -0,0 +1,108 @@
...
+    <g v-if="arrow != null">
```
johnsimons, line 65: Soes this need to be a `g`, could it be just a `template`?

src/Frontend/src/components/messages/SequenceDiagram/RoutesComponent.vue
```diff
@@ -0,0 +1,108 @@
...
+      </g>
+      <!--Message Type and Icon-->
+      <g
+        class="clickable"
+        :transform="`translate(${arrow.messageTypeOffset}, ${arrow.y - 7.5 - Message_Type_Margin})`"
+        :fill="arrow.highlight ? 'var(--highlight)' : 'black'"
+        @mouseover="() => store.setHighlightId(arrow.id)"
+        @mouseleave="() => store.setHighlightId()"
+      >
+        <rect v-if="arrow.highlight && arrow.messageTypeOffset" :width="arrow.highlightTextWidth + 19 + 4 + 4" :height="arrow.highlightTextHeight + 4 + 4" fill="var(--highlight-background)" />
```
johnsimons, line 85: What are the magic numbers?
` + 19 + 4 + 4`
` + 4 + 4`

src/Frontend/src/resources/SequenceDiagram/Endpoint.ts
johnsimons, line 1: Why are we using underscores for private variables?
You can use private visibility - https://www.typescriptlang.org/docs/handbook/2/classes.html#private
Or you can set it at the cstructor - https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

src/Frontend/src/resources/SequenceDiagram/SequenceModel.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  const typeClass = messageType.split(",")[0];
+  const typeName = typeClass.split(".").reverse()[0];
+  return typeName.replace(/\+/g, ".");
+}
+
+export class ModelCreator implements ConversationModel {
+  #endpoints: Endpoint[];
+  #handlers: Handler[];
+  #processingRoutes: MessageProcessingRoute[];
```
johnsimons, line 23: Declare them as private instead? Easier to understand for c# devs

src/Frontend/src/stores/HeartbeatInstancesStore.spec.ts
```diff
@@ -81,7 +81,7 @@ describe("HeartbeatInstancesStore tests", () => {
       [{ name: "", track_instances: true }]
     );
 
-    const names = () => filteredInstances.value.map((value) => value.host_display_name);
+    const names = () => filteredInstances.value.map((value: EndpointsView) => value.host_display_name);
```
johnsimons, line 84: is this needed?

src/Frontend/src/stores/HeartbeatsStore.spec.ts
```diff
@@ -302,7 +302,7 @@ describe("HeartbeatsStore tests", () => {
         ]
       );
 
-      const names = () => filteredHealthyEndpoints.value.map((value) => value.name);
+      const names = () => filteredHealthyEndpoints.value.map((value: LogicalEndpoint) => value.name);
```
johnsimons, line 305: is this needed?

johnsimons (approved):
A few comments added but overall LGTM

src/Frontend/src/components/heartbeats/EndpointInstances.vue
PhilBastian: it started failing my build. Will try removing it again

src/Frontend/src/resources/SequenceDiagram/SequenceModel.ts
PhilBastian: not truly private though. I was only reverting to the typescript private where necessary due to use of proxies

src/Frontend/src/components/messages/SequenceDiagram/EndpointsComponent.vue
PhilBastian: no, since the setEndpointRef function checks for null

src/Frontend/src/components/messages/SequenceDiagram/HandlersComponent.vue
PhilBastian: i could add it to a watch, but since I need to setMaxHeight from here anyway, and therefore have the value that I'd be watching, it doesn't make sense to

src/Frontend/src/components/messages/SequenceDiagram/RoutesComponent.vue
PhilBastian: if anything isn't set yet in the store from other components the other components (e.g. handler locations) then the route will temporarily be null. We don't want to filter it, because we need the index to match with the element ref. An unfortunate result of needing to have the actual DOM elements to get their size/location

src/Frontend/src/components/messages/SequenceDiagram/RoutesComponent.vue
PhilBastian: g's in svg, like divs in html, make for nicer grouping when examining the DOM (e.g. can hover over it and see the full extent of the route, rather than just an individual route)

src/Frontend/src/resources/SequenceDiagram/Endpoint.ts
PhilBastian: why not use underscore?

johnsimons: The code base does not use _ to me private anywhere, are we planning to make this a rule now?

PhilBastian: no, it's `private _variableName`, not just `_variableName`

PR #2310, opened 2025-04-04
Sequence Diagram reverse and local message flow and events

PhilBastian:
timeouts will be included separately.
There is currently a bug with the `startX` value when attempting to set from `HandlersComponent` where multiple items override the value and it ends up being pushed too far to the right


src/Frontend/src/components/messages/SequenceDiagram/HandlersComponent.vue
```diff
@@ -41,8 +41,14 @@ const handlerItems = computed(() => {
       return 0;
     })();
 
+    const messageTypeOffset = handler.direction === Direction.Right ? ((messageTypeElement?.getBBox().width ?? 0) + 24) * -1 : 20;
```
johnsimons, line 44: `* -1` we may need more comments here, it is hard to grok!

src/Frontend/src/resources/SequenceDiagram/RoutedMessage.ts
```diff
@@ -66,7 +66,7 @@ class MessageProcessingRouteItem implements MessageProcessingRoute {
     this.processingHandler = processingHandler;
 
     if (routedMessage && this.processingHandler) {
-      this.name = `${processingHandler?.name}(${routedMessage.messageId})`;
+      this.name = `${processingHandler?.name}(${routedMessage.messageId}###${routedMessage.receiving.name}})`;
```
johnsimons, line 69: What is this for?

johnsimons (approved):
LGTM

src/Frontend/src/resources/SequenceDiagram/RoutedMessage.ts
PhilBastian: the messageId is not unique for the sequence diagram, since it's shared by all routes for an event. This ensures that only a single route is highlighted when mouseovering a handler

johnsimons: could u use the `message.id`?

PhilBastian: that's what was being used originally

johnsimons: originally looks like `messageId`, I am suggesting to use `id` only

PR #2312, opened 2025-04-04
Clears previous timeout when copying

johnsimons:
Small improvement, it seems using a watcher is a better way to clearing previous timeouts


PhilBastian:
where did you source this best practice from? Connect the dots for future reviewers

PhilBastian (approved):

johnsimons:
This cam mostly from reading the docs about side effects - https://vuejs.org/guide/essentials/watchers.html#side-effect-cleanup

PR #2315, opened 2025-04-07
Fixing typescript typings

johnsimons:
Added debounce for filter component


src/Frontend/src/components/FilterInput.vue
```diff
@@ -1,16 +1,23 @@
...
 const model = defineModel<string>({ required: true });
-defineProps({
-  placeholder: {
-    type: String,
-    default: "Filter by name...", // Default value
-  },
+const props = withDefaults(defineProps<{ placeholder?: string; ariaLabel?: string }>(), { placeholder: "Filter by name...", ariaLabel: "filter by name" });
+const localInput = ref<string>(model.value);
+
+const debounceUpdateModel = debounce((value: string) => {
```
PhilBastian, line 9: I think including a library dependency for this functionality is overkill

PhilBastian (approved):

src/Frontend/src/components/FilterInput.vue
PhilBastian: yes, for something as simple as a setTimeout, it is safer to reinvent the wheel rather than pull in a library

PR #2323, opened 2025-04-09
Adding Tippy in TimeSince component

johnsimons:


johnsimons:
To many issues with existing tests, i would have to rewrite them!

PR #2324, opened 2025-04-09
Adding TextEllipses component

johnsimons:
This is to deal with large text display.

The component will automatically add a tooltip based on whether the text overflows.
The ellipses can be added to either side of the text.

### DIFF
--- a/src/Frontend/src/components/TextEllipses.vue
+++ b/src/Frontend/src/components/TextEllipses.vue
@@ -1,0 +1,53 @@
+<script setup lang="ts">
+import { ref, watch, onMounted } from "vue";
+
+const props = withDefaults(
+  defineProps<{
+    text: string;
+    ellipsesStyle?: "RightSide" | "LeftSide";
+  }>(),
+  { ellipsesStyle: "RightSide" }
+);
+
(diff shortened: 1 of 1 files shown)


src/Frontend/src/components/TextEllipses.vue
```diff
@@ -0,0 +1,53 @@
...
+  updateTooltip();
+});
+
+watch([() => props.text], () => {
+  updateTooltip();
+});
+</script>
+
+<template>
+  <div ref="textContainer" title="" class="text-container hackToPreventSafariFromShowingTooltip" :class="{ 'left-side-ellipsis': ellipsesStyle === 'LeftSide' }" v-tippy="{ content: tooltipText, maxWidth: 'none' }">
```
PhilBastian, line 31: given the likely usecases of this control, would it make more sense to be a span rather than a div?

PhilBastian (approved):

src/Frontend/src/components/TextEllipses.vue
johnsimons: It can't be a span, it need to be a block element, I tried it

PR #2330, opened 2025-04-10
Re-introduce back link

johnsimons:


src/Frontend/src/components/messages2/FlowDiagram.vue
```diff
@@ -227,8 +229,8 @@ const greenColor = hexToCSSFilter("#00c468").filter;
             <i class="fa" :class="typeIcon(data.type)" v-tippy="data.type" />
             <div class="lead">
               <strong>
-                <RouterLink v-if="data.isError" :to="{ path: routeLinks.messages.failedMessage.link(data.id) }"><TextEllipses style="width: 204px" :text="data.nodeName" ellipses-style="LeftSide" /></RouterLink>
-                <RouterLink v-else :to="{ path: routeLinks.messages.successMessage.link(data.messageId, data.id) }"><TextEllipses style="width: 204px" :text="data.nodeName" ellipses-style="LeftSide" /></RouterLink>
+                <RouterLink v-if="data.isError" :to="{ path: routeLinks.messages.failedMessage.link(data.id), query: { back: route.path } }"><TextEllipses style="width: 204px" :text="data.nodeName" ellipses-style="LeftSide" /></RouterLink>
```
PhilBastian, line 232: won't this make the back route be this message, rather than the list/whatever that originally brought us to this list? We want back to be `route.query.back` or something similar instead?

PhilBastian (changes requested):

src/Frontend/src/components/messages2/FlowDiagram.vue
johnsimons: Fixed @PhilBastian

PhilBastian (approved):

PR #2331, opened 2025-04-10
Fixed issue with content type not being set

johnsimons:


src/Frontend/src/components/failedmessages/EditRetryDialog2.vue
```diff
@@ -122,6 +108,11 @@ async function retryEditedMessage() {
 }
 
 function initializeMessageBodyAndHeaders() {
+  function getHeaderValue(key: string) {
+    const header = local.headers.find((header: HeaderWithEditing) => header.key === key);
```
PhilBastian, line 112: i realise this works because of hoisting, but this function is declared before `local` is. 
```suggestion
    const header = headers.value.find((header: HeaderWithEditing) => header.key === key);
```

PhilBastian (approved):

PR #2334, opened 2025-04-11
Sequence Diagram message selection

PhilBastian:
Implement message selection/navigation as per ServiceInsight functionality, highlighting current message
also keep endpoint names visible during y-scrolling

### DIFF
--- a/src/Frontend/src/stores/SequenceDiagramStore.ts
+++ b/src/Frontend/src/stores/SequenceDiagramStore.ts
@@ -1,6 +1,6 @@
 import { acceptHMRUpdate, defineStore, storeToRefs } from "pinia";
 import { computed, ref, watch } from "vue";
-import { friendlyTypeName, ModelCreator } from "@/resources/SequenceDiagram/SequenceModel";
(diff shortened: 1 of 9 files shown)


src/Frontend/src/stores/SequenceDiagramStore.ts
```diff
@@ -36,6 +40,8 @@ export const useSequenceDiagramStore = defineStore("SequenceDiagramStore", () =>
   const handlerLocations = ref<HandlerLocation[]>([]);
   const highlightId = ref<string>();
 
+  const selectedId = computed(() => `${friendlyTypeName(state.value.data.message_type ?? "")}(${state.value.data.id})`);
```
johnsimons, line 43: Why is `selectedId` using `friendlyTypeName`, feels a bit strange?

PhilBastian: changed (throughout) to use full message_type for internal ids and only use friendlyName for display purposes

johnsimons (approved):

PR #2338, opened 2025-04-14
cleanup TODO items

PhilBastian:
- move interfaces/classes that aren't directly read from ServiceControl APIs from resources to the component folder
- extract util functions for DRY
- add link to (currently ServiceInsight) help page for Sequence Diagram
- fix Sequence Diagram startX offset that was broken by moving the endpoint render after all other elements


johnsimons:
Could we separate these into individual PRs

soujay (approved):

PhilBastian:
will break this up as suggested by John

PR #2350, opened 2025-04-24
Handle receiving endpoint not being  present for failures

andreasohlund:
As discovered in https://github.com/Particular/ServiceControl/pull/4943/files#diff-b318f9c690fb6d2daac8156a8c023e2cb1fcb7d10caf0d4535c27aa25357f11dR62 there are scenarios where the receiving endpoint can't be detected. This change handles that edge case


cquirosj:
@andreasohlund, there is a new MessageView.Vue component under a folder named Messages2. This will eventually replace the original MessageView. In the new version, these values are already conditionally rendered:

https://github.com/Particular/ServicePulse/blob/d3e8b60c4aa1c0d277c60e82f13a01a2f478c4cf/src/Frontend/src/components/messages2/MessageView.vue#L134

johnsimons:
@andreasohlund this is interesting, becasue there are a few places in the UI where we use this value, and I am wondering if we need to handle this scenario more explicitly.

So we have things like the flow diagram that displays the addresses
image

We also have other diagrams that I am not 100% sure what happens there!

It may be worth doing a code search everywhere to see how this data is used and what are the ramifications if there is no data.

andreasohlund:
I'll take a look.

Side note:

If we merge https://github.com/Particular/ServiceControl/pull/4945 the result will be that we'll always have an endpoint name (since its deducted from the FailedQ headers which is mandatory) and it will only be the host that can be missing

PR #2357, opened 2025-04-28
fix startX determination now that endpoints render last

PhilBastian:


johnsimons:
@PhilBastian I am not too sure what I ma reviewing here, can you provide more context on what this is fixing?
Maybe a before and after screenshot, or how do we test it?

PhilBastian:
image
vs
image

this was working previously, but was broken by rendering the endpoints last (resulting in the centrepoint not being known for the initial render of the handlers)

johnsimons (approved):

PR #2360, opened 2025-04-29
size header list according to contents and eliminate size jump on hover

PhilBastian:
alternative to https://github.com/Particular/ServicePulse/pull/2314 
image

### DIFF
--- a/src/Frontend/src/components/messages2/HeadersView.vue
+++ b/src/Frontend/src/components/messages2/HeadersView.vue
@@ -35,19 +35,15 @@ const filteredHeaders = computed(() => {
       </div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


johnsimons:
@PhilBastian I pushed a change to this PR to make the first column fixed size and the second column is content based but with extra allocated size for the clipboard.
This makes it so the clipboard button is still close to the content and not all the way to the right.

johnsimons (approved):

PhilBastian:
> @PhilBastian I pushed a change to this PR to make the first column fixed size and the second column is content based but with extra allocated size for the clipboard. This makes it so the clipboard button is still close to the content and not all the way to the right.

IMO this looks far worse. Lets take it to the taskforce

PR #2361, opened 2025-04-29
Tooltip to use wildcards

johnsimons:
I have to use a font awesome image for the wildcard character becasue by default it prints quite small `*`. So not really visible!

image

### DIFF
--- a/src/Frontend/src/components/audit/FiltersPanel.vue
+++ b/src/Frontend/src/components/audit/FiltersPanel.vue
@@ -3,14 +3,16 @@ import FilterInput from "@/components/FilterInput.vue";
 import { storeToRefs } from "pinia";
 import { FieldNames, useAuditStore } from "@/stores/AuditStore.ts";
 import ListFilterSelector from "@/components/audit/ListFilterSelector.vue";
-import { computed } from "vue";
+import { computed, useTemplateRef } from "vue";
 import DatePickerRange from "@/components/audit/DatePickerRange.vue";
+import { Tippy, TippyComponent } from "vue-tippy";
 
 const store = useAuditStore();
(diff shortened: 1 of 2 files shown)


src/Frontend/src/components/audit/FiltersPanel.vue
```diff
@@ -51,13 +53,31 @@ function findKeyByValue(searchValue: string) {
...
-      <div class="filter-component text-search-container"><FilterInput v-model="messageFilterString" placeholder="Search messages..." aria-label="Search messages" /></div>
+      <div class="filter-component text-search-container">
+        <Tippy ref="wildcardTooltipRef" trigger="click" :hideOnClick="false">
+          <template #content>
+            <h4>Use <i class="fa fa-asterisk asterisk" /> to do wildcard searches <i class="fa fa-lightbulb-o" style="color: #e6c201" /></h4>
+            <p>
+              Example: <i><i class="fa fa-asterisk asterisk" />World!</i> or <i>Hello<i class="fa fa-asterisk asterisk" /></i>, to look for <i>Hello World!</i>
+            </p>
+          </template>
+          <FilterInput v-model="messageFilterString" placeholder="Search messages..." aria-label="Search messages" @focus="() => toggleWildcardToolTip(true)" @blur="() => toggleWildcardToolTip(false)" />
```
cquirosj, line 78: @johnsimons, what do you think about hiding the tooltip when the user starts typing? It isn't very pleasant to always have it visible the entire time the user is focused. 
```suggestion
          <FilterInput v-model="messageFilterString" placeholder="Search messages..." aria-label="Search messages" @focus="() => toggleWildcardToolTip(true)" @blur="() => toggleWildcardToolTip(false)" @input="() => toggleWildcardToolTip(false)" />
```

johnsimons: that is a good idea, merged

PhilBastian (approved):

PR #2365, opened 2025-04-30
Fix reset original body

johnsimons:
The edit and retry dialog was suppose to show a "reset body" back to the original button.
But unfortunately the logic had a few bugs which cuased the button to not always be shown.

This PR fixes that issue as well as brings the button to the top of the dialog to be more inline with current design in other areas.

### DIFF
--- a/src/Frontend/src/components/failedmessages/EditRetryDialog2.vue
+++ b/src/Frontend/src/components/failedmessages/EditRetryDialog2.vue
@@ -9,6 +9,7 @@ import CodeEditor from "@/components/CodeEditor.vue";
 import { useMessageStore } from "@/stores/MessageStore";
 import { storeToRefs } from "pinia";
 import LoadingSpinner from "@/components/LoadingSpinner.vue";
+import debounce from "lodash/debounce";
 
 interface HeaderWithEditing extends Header {
   isLocked: boolean;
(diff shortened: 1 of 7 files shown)


src/Frontend/src/components/failedmessages/EditRetryDialog2.vue
```diff
@@ -41,28 +42,32 @@ const localMessage = ref<LocalMessageState>({
...
 const { state, headers, body, edit_and_retry_config } = storeToRefs(store);
 const id = computed(() => state.value.data.id ?? "");
-const messageBody = computed(() => body.value.data.value);
+const uneditedMessageBody = computed(() => body.value.data.value ?? "");
+const regExToPruneLineEndings = new RegExp(/[\n\r]*/, "g");
+const debounceBodyUpdate = debounce((value: string) => {
+  const newValue = value.replaceAll(regExToPruneLineEndings, "");
+  localMessage.value.isBodyChanged = newValue !== uneditedMessageBody.value.replaceAll(regExToPruneLineEndings, "");
+  localMessage.value.isBodyEmpty = newValue === "";
+}, 1000);
```
PhilBastian, line 63: these values affect the save? What happens during the second that you're waiting for the update to happen? Why is the debounce required?

PhilBastian (commented):
can you provide a more detailed description of what the issue was that this is fixing? Difficult to get a handle on it from just the code...

src/Frontend/src/components/failedmessages/EditRetryDialog2.vue
johnsimons: Becasue we don't want to fire this callback on every keystroke. The callback has a few things happening that can be expensive.
The worse it happens is that we won't see the "reset" button for 1 second.

johnsimons:
@PhilBastian I have updated the description

src/Frontend/src/components/failedmessages/EditRetryDialog2.vue
PhilBastian: as mentioned in our call, this should be removed for the isBodyEmpty check, retained for the isBodyChanged

PhilBastian (approved):

PR #2375, opened 2025-05-01
Add autofocus for filter dropdown

johnsimons:


PhilBastian:
question about the design: does the bootstrap dropdown get created every time its shown and destroyed every time it's not show, or is it just hidden when it's not shown? If the former, can a prop be provided to filterinput to allow it to focus itself onMounted?

johnsimons:
I think it only gets created when it is toggled

PhilBastian (approved):

PR #2376, opened 2025-05-01
change endpoint name to html element so that long names can wrap

PhilBastian:
image


PhilBastian:
mask-image is broken in safari

johnsimons (approved):
LGTM

PR #2383, opened 2025-05-02
replace wildcard tooltip with static text, replace the instructions text

soujay:
replace wildcard tooltip with static text and  replace the instructions text to get a cleaner and tighter look.

Old way of using a tooltip
image

New way of using a static text
image


PhilBastian:
the extra wasted space this adds is not nice, especially with the large margin at the bottom. I definitely prefer the tooltip

soujay:
@PhilBastian 
I made an improvement from the following 

image

to 
image

cquirosj (approved):

PR #2400, opened 2025-05-06
add loading spinner in all messages list page

soujay:
add loading spinner in all messages list page


johnsimons (commented):
I have pushed a change to it @soujay, have a look

PhilBastian (approved):

PR #2402, opened 2025-05-07
Display saga messages body using CodeEditor

cquirosj:
TODO:
- [x] improve styles for the maximize button in maximizable control
- [x] Test with XML

Potential improvements:

- [ ] Extraction of the logic to load and parse a message body into a composable to be reused by Messagestore and SagaDiagramStore
- [ ] Potential reusable component for displaying the message body, to be used in BodyView.vue and MessageDataBox.vue - Both can also be maximizable

### DIFF
--- a/src/Frontend/src/components/MaximizableCodeEditor.vue
+++ b/src/Frontend/src/components/MaximizableCodeEditor.vue
@@ -1,0 +1,168 @@
+<script setup lang="ts">
+import { ref, onMounted, onBeforeUnmount } from "vue";
+import CodeEditor from "@/components/CodeEditor.vue";
+import DiffMaximizeIcon from "@/assets/diff-maximize.svg";
+import DiffCloseIcon from "@/assets/diff-close.svg";
+import { Extension } from "@codemirror/state";
+import { CodeLanguage } from "@/components/codeEditorTypes";
+
+// Define the model value to be passed to CodeEditor
+const modelValue = defineModel<string>({ required: true });
+
+// Define props by extending the CodeEditor props and adding maximize-specific ones
+withDefaults(
+  defineProps<{
+    language?: CodeLanguage;
+    readOnly?: boolean;
+    showGutter?: boolean;
+    showCopyToClipboard?: boolean;
+    ariaLabel?: string;
+    extensions?: Extension[];
+    modalTitle?: string;
+  }>(),
+  {
+    readOnly: false,
+    showGutter: false,
+    showCopyToClipboard: false,
+    extensions: () => [],
+    modalTitle: "Code View",
+  }
+);
+
+// Component state for maximize functionality
+const showMaximizeModal = ref(false);
+
+// Handle maximize functionality
+const toggleMaximizeModal = () => {
+  showMaximizeModal.value = !showMaximizeModal.value;
+};
+
+// Handle ESC key to close modal
+const handleKeyDown = (event: KeyboardEvent) => {
+  if (event.key === "Escape" && showMaximizeModal.value) {
+    showMaximizeModal.value = false;
+  }
+};
+
+// Setup keyboard events for maximize modal
+onMounted(() => {
+  window.addEventListener("keydown", handleKeyDown);
+});
+
+// Clean up event listeners when component is destroyed
+onBeforeUnmount(() => {
+  window.removeEventListener("keydown", handleKeyDown);
+});
+</script>
+
+<template>
+  <div class="code-editor-wrapper">
+    <!-- Regular CodeEditor with maximize button in toolbarRight -->
+    <CodeEditor v-model="modelValue" :language="language" :read-only="readOnly" :show-gutter="showGutter" :show-copy-to-clipboard="showCopyToClipboard" :aria-label="ariaLabel" :extensions="extensions">
+      <template #toolbarLeft>
+        <slot name="toolbarLeft"></slot>
+      </template>
+      <template #toolbarRight>
+        <slot name="toolbarRight"></slot>
+        <img class="maximize-icon-inline" :src="DiffMaximizeIcon" alt="Maximize" @click="toggleMaximizeModal" title="Maximize view" />
+      </template>
+    </CodeEditor>
+
+    <!-- Maximize modal for CodeEditor -->
+    <div v-if="showMaximizeModal" class="maximize-modal">
+      <div class="maximize-modal-content">
+        <div class="maximize-modal-toolbar">
+          <span class="maximize-modal-title">{{ modalTitle }}</span>
+          <img class="maximize-modal-close" :src="DiffCloseIcon" alt="Close" @click="toggleMaximizeModal" title="Close" />
+        </div>
+        <div class="maximize-modal-body">
+          <CodeEditor v-model="modelValue" :language="language" :read-only="readOnly" :show-copy-to-clipboard="true" :show-gutter="true" :aria-label="ariaLabel" :extensions="[]" />
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+.code-editor-wrapper {
+  position: relative;
+  width: 100%;
+}
+
+/* Maximize icon styles */
+.maximize-icon-inline {
+  width: 14px;
+  height: 14px;
+  cursor: pointer;
+  opacity: 0.7;
+  transition: opacity 0.2s ease;
+  margin-left: 8px;
+}
+
+.maximize-icon-inline:hover {
+  opacity: 1;
+}
+
+/* Modal styles copied from DiffViewer */
+.maximize-modal {
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: rgba(0, 0, 0, 0.5);
+  z-index: 1000;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.maximize-modal-content {
+  background-color: white;
+  width: calc(100% - 40px);
+  height: calc(100% - 40px);
+  border-radius: 4px;
+  overflow: hidden;
+  display: flex;
+  flex-direction: column;
+  box-shadow: 0 4px 8px rgba(0, 0, 0, 0.1);
+}
+
+.maximize-modal-toolbar {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 10px 15px;
+  background-color: #f8f8f8;
+  border-bottom: 1px solid #ddd;
+}
+
+.maximize-modal-title {
+  font-weight: bold;
+  font-size: 16px;
+}
+
+.maximize-modal-close {
+  background: none;
+  border: none;
+  cursor: pointer;
+  width: 16px;
+  height: 16px;
+}
+
+.maximize-modal-body {
+  flex: 1;
+  overflow: auto;
+  padding: 15px;
+}
+
+/* Ensure the CodeEditor wrapper fills the modal body */
+.maximize-modal-body :deep(.wrapper) {
+  height: calc(100% - 20px);
+}
+
+.maximize-modal-body :deep(.cm-editor),
+.maximize-modal-body :deep(.cm-scroller) {
+  height: 100%;
+}
+</style>
--- a/src/Frontend/src/components/MaximizableCodeEditor.vue
+++ b/src/Frontend/src/components/MaximizableCodeEditor.vue
@@ -6,10 +6,8 @@ import DiffCloseIcon from "@/assets/diff-close.svg";
 import { Extension } from "@codemirror/state";
 import { CodeLanguage } from "@/components/codeEditorTypes";
 
-// Define the model value to be passed to CodeEditor
 const modelValue = defineModel<string>({ required: true });
 
-// Define props by extending the CodeEditor props and adding maximize-specific ones
 withDefaults(
   defineProps<{
     language?: CodeLanguage;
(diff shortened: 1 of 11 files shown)


johnsimons:
the font size on the initial message seems larger
image

src/Frontend/src/components/MaximizableCodeEditor.vue
```diff
@@ -0,0 +1,207 @@
...
+});
+</script>
+
+<template>
+  <div class="code-editor-wrapper" @mouseenter="onEditorMouseEnter" @mouseleave="onEditorMouseLeave">
+    <!-- Regular CodeEditor -->
+    <div class="editor-container">
+      <!-- Maximize Button (shown on hover) -->
+      <button v-if="showMaximizeButton" @click="toggleMaximizeModal" class="maximize-button" title="Maximize view">
+        <img :src="DiffMaximizeIcon" alt="Maximize" width="14" height="14" />
```
johnsimons, line 70: it would be better to use the svg directly not via a img tag

johnsimons (approved):
Overall looks pretty good.
I am a bit unsure about us creating yet another CodeEditor though!

src/Frontend/src/components/MaximizableCodeEditor.vue
```diff
@@ -0,0 +1,207 @@
...
+  window.removeEventListener("keydown", handleKeyDown);
+});
+</script>
+
+<template>
+  <div class="code-editor-wrapper" @mouseenter="onEditorMouseEnter" @mouseleave="onEditorMouseLeave">
+    <!-- Regular CodeEditor -->
+    <div class="editor-container">
+      <!-- Maximize Button (shown on hover) -->
+      <button v-if="showMaximizeButton" @click="toggleMaximizeModal" class="maximize-button" title="Maximize view">
```
johnsimons, line 69:
```suggestion
      <button v-if="showMaximizeButton" @click="toggleMaximizeModal" class="maximize-button" v-tippy="`Maximize view`">
```

johnsimons (approved):

src/Frontend/src/components/MaximizableCodeEditor.vue
```diff
@@ -0,0 +1,207 @@
+<script setup lang="ts">
+import { ref, onMounted, onBeforeUnmount } from "vue";
+import CodeEditor from "@/components/CodeEditor.vue";
+import DiffMaximizeIcon from "@/assets/diff-maximize.svg";
+import DiffCloseIcon from "@/assets/diff-close.svg";
```
johnsimons, line 5: We should try to use fontawesome icons if available

src/Frontend/src/components/MaximizableCodeEditor.vue
```diff
@@ -0,0 +1,207 @@
...
+        <img :src="DiffMaximizeIcon" alt="Maximize" width="14" height="14" />
+      </button>
+
+      <CodeEditor class="maximazable-code-editor--inline-instance" v-model="modelValue" :language="language" :read-only="readOnly" :show-gutter="showGutter" :show-copy-to-clipboard="false" :aria-label="ariaLabel" :extensions="extensions">
+        <template #toolbarLeft>
+          <slot name="toolbarLeft"></slot>
+        </template>
+        <template #toolbarRight>
+          <slot name="toolbarRight"></slot>
+        </template>
```
johnsimons, line 79:
```suggestion
```

src/Frontend/src/components/MaximizableCodeEditor.vue
```diff
@@ -0,0 +1,207 @@
...
+  border-radius: 0;
+  background-color: transparent;
+  padding: 0;
+  margin-bottom: 0;
+}
+
+:deep(.wrapper.maximazable-code-editor--inline-instance .cm-editor) {
+  /* Override any borders from the default theme */
+  border: none;
+}
```
johnsimons, line 141: Why do we still need this here?
Shouldn't this be extensions?

src/Frontend/src/components/messages2/SagaDiagram/MessageDataBox.vue
```diff
@@ -1,27 +1,58 @@
...
+    <span class="message-data-box-text--error">Message body not found</span>
+  </div>
+  <div v-else-if="!messageDataLoading && !messageData.body.data.value" class="message-data-box">
     <span class="message-data-box-text--empty">Empty</span>
   </div>
-  <div v-else-if="messageData.length > 0" v-for="(item, index) in messageData" :key="index" class="message-data-box">
-    <b class="message-data-box-text">{{ item.key }}</b>
-    <span class="message-data-box-text">=</span>
-    <span class="message-data-box-text--ellipsis" :title="item.value">{{ item.value }}</span>
+  <div v-else-if="contentType.isSupported" class="message-data-box message-data-box-content">
```
PhilBastian, line 51: I would be inclined to have this as the v-else element, since it's the default, rather than content not supported

src/Frontend/src/components/messages2/SagaDiagram/SagaUpdateNode.vue
```diff
@@ -202,7 +190,7 @@ const hasStateChanges = computed(() => {
 
             <!-- Initial state display -->
             <div v-else-if="update.IsFirstNode" class="json-container json-container--first-node">
-              <CodeEditor :model-value="sagaUpdateStateChanges.formattedState || ''" language="json" :showCopyToClipboard="false" :showGutter="false" :extensions="[monospaceTheme]" />
+              <MaximizableCodeEditor :model-value="sagaUpdateStateChanges.formattedState || ''" language="json" :showGutter="false" modalTitle="Saga Initialized" :extensions="[monospaceTheme]" />
```
PhilBastian, line 193:
```suggestion
              <MaximizableCodeEditor :model-value="sagaUpdateStateChanges.formattedState || ''" language="json" :showGutter="false" modalTitle="Initial Saga State" :extensions="[monospaceTheme]" />
```

src/Frontend/src/components/messages2/SagaDiagram/MessageDataBox.vue
PhilBastian: also, in the store, contentType defaults to "text/plain", so is the contentType check even needed?

src/Frontend/src/stores/SagaDiagramStore.ts
```diff
@@ -3,16 +3,13 @@ import { ref, watch } from "vue";
...
 
-const StandardKeys = ["$type", "Id", "Originator", "OriginalMessageId"];
-export interface SagaMessageDataItem {
-  key: string;
-  value: string;
-}
 export interface SagaMessageData {
   message_id: string;
-  data: SagaMessageDataItem[];
+  body: DataContainer<{ value?: string; content_type?: string }>;
```
PhilBastian, line 12: why content_type instead of contentType? It seems like this is not an API response object

src/Frontend/src/stores/SagaDiagramStore.ts
```diff
@@ -73,45 +70,40 @@ export const useSagaDiagramStore = defineStore("SagaDiagramStore", () => {
...
-
   async function fetchSagaMessageData(message: SagaMessage): Promise<SagaMessageData> {
     const bodyUrl = (message.body_url ?? formatUrl(MessageBodyEndpoint, message.message_id)).replace(/^\//, "");
+    const result: SagaMessageData = {
+      message_id: message.message_id,
+      body: { data: {} },
+    };
+
+    result.body.loading = true;
+    result.body.failed_to_load = false;
```
PhilBastian, line 81: why failed_to_load and not_found rather than failedToLoad and notFound?

src/Frontend/src/stores/SagaDiagramStore.ts
```diff
@@ -73,45 +70,40 @@ export const useSagaDiagramStore = defineStore("SagaDiagramStore", () => {
...
+      result.body.data.content_type = contentType ?? "text/plain";
+      result.body.data.value = await response.text();
 
-      let data: SagaMessageDataItem[];
-      if (typeof body === "string" && body.trim().startsWith("<?xml")) {
-        data = getXmlData(body);
-      } else {
-        data = processJsonValues(body);
+      if (contentType === "application/json") {
+        result.body.data.value = stringify(parse(result.body.data.value), null, 2) ?? result.body.data.value;
```
PhilBastian, line 95: shouldn't need to parse and stringify if we're using the codeeditor for display, since that formats for us?

src/Frontend/src/stores/SagaDiagramStore.ts
```diff
@@ -73,45 +70,40 @@ export const useSagaDiagramStore = defineStore("SagaDiagramStore", () => {
...
       }
-      // Check if parsed data is empty
-      if (!data || data.length === 0) {
-        return createEmptyMessageData(message.message_id);
+      if (contentType === "text/xml") {
+        result.body.data.value = xmlFormat(result.body.data.value, { indentation: "  ", collapseContent: true });
```
PhilBastian, line 98: do we need to format? Can the codeeditor do it for us?

PhilBastian (changes requested):

src/Frontend/src/stores/SagaDiagramStore.ts
johnsimons: Nope, unfortunately

src/Frontend/src/components/MaximizableCodeEditor.vue
```diff
@@ -0,0 +1,207 @@
...
+        </template>
+      </CodeEditor>
+    </div>
+
+    <!-- Maximize modal for CodeEditor -->
+    <div v-if="showMaximizeModal" class="maximize-modal">
+      <div class="maximize-modal-content">
+        <div class="maximize-modal-toolbar">
+          <span class="maximize-modal-title">{{ modalTitle }}</span>
+          <button @click="toggleMaximizeModal" class="maximize-modal-close" title="Close">
```
johnsimons, line 88: I thought we already have a dialog component, shouldn't we use that for this dialog?

src/Frontend/src/components/MaximizableCodeEditor.vue
soujay: @cquirosj : It will good to decrease the size of the modal window
- components\messages2\DiffViewer.vue
The following change in .maximize-modal-content CSS decreases the window slightly
.maximize-modal-content {
  background-color: white;
  width: calc(100% - 200px);
  height: calc(100% - 200px);
  border-radius: 4px;
  overflow: hidden;
  display: flex;
  flex-direction: column;
  box-shadow: 0 4px 8px rgba(0, 0, 0, 0.1);
}

src/Frontend/src/components/MaximizableCodeEditor.vue
cquirosj: I tried using extensions for themes, but that resulted in a bug where other instances of CodeEditor get affected. Not sure if it is a bug in CodeMirror or in the third-party library we are using that wraps CodeMirror. I am going to keep this implementation using: deep for now.

src/Frontend/src/components/MaximizableCodeEditor.vue
```diff
@@ -0,0 +1,207 @@
...
+            <img :src="DiffCloseIcon" alt="Close" width="16" height="16" />
+          </button>
+        </div>
+        <div class="maximize-modal-body">
+          <CodeEditor class="maximazable-code-editor--pop-up-instance" v-model="modelValue" :language="language" :read-only="readOnly" :show-copy-to-clipboard="true" :show-gutter="true" :aria-label="ariaLabel" :extensions="[]" />
```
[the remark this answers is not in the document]
cquirosj: The control has two CodeEditors; the extensions pro was meant to affect only the inline editor, not the one in the pop-up. If customizing the one in the pop-up is needed, the feature can be implemented when the need arises

src/Frontend/src/components/messages2/SagaDiagram/MessageDataBox.vue
```diff
@@ -1,27 +1,58 @@
...
+const modalTitle = computed(() => {
+  return props.maximizedTitle ? `Message Data - ${props.maximizedTitle}` : "Message Data";
+});
+
 const sagaDiagramStore = useSagaDiagramStore();
 const { messageDataLoading } = storeToRefs(sagaDiagramStore);
+
+const contentType = computed(() => parseContentType(props.messageData.body.data.content_type));
+
+const body = computed(() => props.messageData.body.data.value || "");
```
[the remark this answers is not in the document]
cquirosj: It is used; otherwise, the linting would have complained

src/Frontend/src/components/messages2/SagaDiagram/MessageDataBox.vue
cquirosj: The value of contentType comes from invoking the function parseContentType in contentTypeParser.ts, which can result in an unsupported type. The default from the request code is defaulting to "text/plain" only if the content-type attribute is missing in the HTTP request result.

src/Frontend/src/stores/SagaDiagramStore.ts
cquirosj: https://github.com/Particular/ServicePulse/pull/2402#discussion_r2077022039

src/Frontend/src/components/MaximizableCodeEditor.vue
cquirosj: Gonna leave this one out for now. But that is a good catch.

PhilBastian (approved):

PR #2404, opened 2025-05-07
Allow users to see more diagram

johnsimons:
This PR removed the default setting of zooming in and out based on scrolling instead users use Ctrl+scroll or the + - controls. This allows much more fine control.

And also allows the user to use as much of their screen size as possible  to see the whole diagram by not confining it to a small window.

Here is a before UX:
image

After UX:
image


PhilBastian:
scroll wheel is more intuitive than using +- controls. How does the user get out of this view now?

johnsimons:
> How does the user get out of this view now?

Scroll up!

PhilBastian:
oh... I see, so there's a bigger view for the flow diagram, but still not the full height necessarily, and the user needs to scroll on the page and then still move the diagram around using mouse click+drag. I do not think this is a UI improvement

PR #2411, opened 2025-05-10
fix the tippy text on the message node of flow diagram

soujay:
fix the tippy text on the message node of flow diagram


cquirosj (approved):

src/Frontend/src/components/messages2/FlowDiagram/FlowDiagram.vue
```diff
@@ -225,7 +225,7 @@ const selectedErrorColor = hexToCSSFilter("#e8e6e8").filter;
       <VueFlow :nodes="nodes" :edges="edges" :min-zoom="0.1" :max-zoom="1.2" :only-render-visible-elements="true" @nodes-initialized="layoutGraph">
         <Controls :show-interactive="false" position="top-left" class="controls" />
         <template #node-message="{ id, data }: { id: string; data: NodeData }">
-          <TextEllipses class="address" :text="`${data.sendingEndpoint.name}@${data.sendingEndpoint.host}`" />
+          <TextEllipses class="address" :text="`${data.sendingEndpoint.name}@${data.sendingEndpoint.host}`" v-tippy="'Sending Endpoint'" />
```
johnsimons, line 228: We can't do this, becasue the whole point of the `TextEllipses` component is it renders a shorter endpoint name with ellipses and then uses the tooltip to show the full name!

johnsimons (changes requested):
The tooltips on the addresses need to be removed.
We can't have 2 tooltips!

johnsimons (approved):

Issue #2424, opened 2025-05-21
Edit & Retry sends both edited and unedited copies of the errored message


jpalac:
#### Description

When doing a `Edit & Retry` of a message in ServicePulse, both the edited and unedited versions of the message are sent/retried.

#### Expected behavior

Only the edited version of the message is sent/retried.

#### Versions

ServicePulse 2.0+

### Steps to reproduce

- Run ServiceControl and ServicePulse
- Setup and endpoint with a handler that accepts a message with a boolean property, eg Error
- In the handler, if the property is set to true, throw an error, if it's false process the message successfully
- Run the endpoint with the property being set to `true` so that the message ends up in ServicePulse
- In ServicePulse, find the errored message, go to `Edit & Retry`, click on `Message Body` and change the property to `false`
- Press `Retry` and then confirm it
- Notice that the message will be processed by the endpoint twice - once with the edited body and once with the original one, hence it will have one successful processing and one failure.

### Relevant log output

```shell
NA
```

jpalac:
ServicePulse [2.0.1](https://discuss.particular.net/t/servicepulse-2-0-1-patch-release-available/4336) has been released.

PR #2425, opened 2025-05-21
Fixes double retry when using edit and retry

johnsimons:
Fixes https://github.com/Particular/ServicePulse/issues/2424

### DIFF
--- a/src/Frontend/src/components/messages2/EditAndRetryButton.vue
+++ b/src/Frontend/src/components/messages2/EditAndRetryButton.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { useMessageStore } from "@/stores/MessageStore";
 import { computed, ref } from "vue";
-import { showToastAfterOperation } from "@/composables/toast";
+import { useShowToast } from "@/composables/toast";
 import { TYPE } from "vue-toastification";
 import EditRetryDialog2 from "@/components/failedmessages/EditRetryDialog2.vue";
 import { MessageStatus } from "@/resources/Message";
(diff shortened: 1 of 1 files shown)


src/Frontend/src/components/failedmessages/EditRetryDialog.vue
johnsimons, line 1: This is just a clean-up.
We used to have two edit and retry dialogs, but in reality only `EditRetryDialog2.vue` was being used, so all i have done is to rename `EditRetryDialog2.vue` -> `EditRetryDialog.vue`.

The reason we had 2 dialogs was when we had a feature flag, and we forogt to do this clean-up as part of removing the feature flag.

src/Frontend/src/components/messages2/EditAndRetryButton.vue
```diff
@@ -19,8 +19,7 @@ const handleConfirm = async () => {
   isConfirmDialogVisible.value = false;
 
   const message = `Retrying the edited message ${state.value.data.id} ...`;
-  await showToastAfterOperation(store.retryMessage, TYPE.INFO, "Info", message);
```
johnsimons, line 22: This is the actual fix, we no longer call `store.retryMessage` from here given that is already called in the dialog itself.

jpalac:
@johnsimons I have verified that the specific issue that was reported is not longer happening on this branch.
However since there is all this other code cleanup/changes in there for which I have no context, I'd feel more comfortable if someone else in your task force reviewed that bit.

johnsimons:
@jpalac I rolled back those cleanups

jpalac (approved):

PR #2476, opened 2025-06-19
Saga localization tests fix: Retrieve UI date strings and convert to date values for performing assertions

cquirosj:
Removing the need to test too close to the implementation

import { toLocalDateTimeString } from "@/composables/formatUtils";

### DIFF
--- a/src/Frontend/src/components/messages/SagaDiagram.spec.ts
+++ b/src/Frontend/src/components/messages/SagaDiagram.spec.ts
@@ -5,7 +5,6 @@ import makeRouter from "@/router";
 import { createTestingPinia } from "@pinia/testing";
 import { MessageStore } from "@/stores/MessageStore";
 import { MessageStatus } from "@/resources/Message";
-import { toLocalDateTimeString } from "@/composables/formatUtils";
 
 //Defines a domain-specific language (DSL) for interacting with the system under test (sut)
 interface componentDSL {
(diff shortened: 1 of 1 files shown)


src/Frontend/src/components/messages/SagaDiagram.spec.ts
```diff
@@ -350,8 +349,16 @@ function rendercomponent({ initialState = {} }: { initialState?: { MessageStore?
 
         const sagaUpdatesTimestampsValues = sagaUpdatesTimestamps.map((item) => item.innerHTML);
 
-        // //check if the values are in the same order as the sagaUpdates array passed to this function
-        expect(sagaUpdatesTimestampsValues).toEqual(sagaUpdates.map((item) => item.expectedRenderedLocalTime));
+        // Parse the rendered timestamp strings back to Date objects for comparison
+        const parsedDatesFromUI = sagaUpdatesTimestampsValues.map((timestampString) => {
+          // Parse the retrieved timestamp string  back to a Date
+          return new Date(timestampString);
```
jasontaylordev, line 355: How will `new Date(string)` reliably parse localized strings like "04/03/2025 03:04 PM" or "03/04/2025 03:04 PM"?
Either could be 4 Mar or 3 Apr depending on parsing.

cquirosj: We can use a set of expectations that are not susceptible to that. Could you update the constants?

jasontaylordev (approved):

PR #2479, opened 2025-06-20
Revert test fix

jasontaylordev:
Reverting the previous fix to these tests, as it still fails across different locale formats (e.g. en-AU).
The tests pass in GitHub Actions (which uses en-US), but they can fail in local development environments with different locale settings.
I’ve created a [new bug](https://github.com/Particular/ServicePulse/issues/2480) to track and address this issue separately.


cquirosj (changes requested):
I think this can be closed, this [PR](https://github.com/Particular/ServicePulse/pull/2483) is another attempt to fix the issue (pretty much the original solution you had, but with a small difference). This has been tested on a computer using en-AU to confirm. @jasontaylordev could you please test and confirm? 🙏 thanks!

PR #2506, opened 2025-07-08
Upgrade to latest vite

johnsimons:
I read the [upgrade guide](https://vite.dev/guide/migration.html) and there does not seem to be anything that we need to change on our end.

### DIFF
--- a/src/Frontend/src/components/audit/DatePickerRange.vue
+++ b/src/Frontend/src/components/audit/DatePickerRange.vue
@@ -43,7 +43,15 @@ function formatDate([fromDate, toDate]: DateRange) {
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


src/Frontend/src/components/audit/DatePickerRange.vue
```diff
@@ -43,7 +43,15 @@ function formatDate([fromDate, toDate]: DateRange) {
 </script>
 
 <template>
-  <VueDatePicker ref="datePicker" class="dropdown" v-model="internalModel" :format="formatDate" :range="{ partialRange: false }" :enable-seconds="true" :action-row="{ showNow: false, showCancel: false, showSelect: true }">
```
johnsimons, line 46: The `:format` type, had issues, it seems the new version of the `vue-tsc` is more type safe and need an explicit cast.

PR #2509, opened 2025-07-10
All messages anchor links POC

cquirosj:


src/Frontend/src/components/messages/MessageView.vue
```diff
@@ -98,6 +98,16 @@ onMounted(() => {
     backLink.value = { path: back.toString(), query: otherArgs };
   }
 });
+
+// Update document title to include message status and type for easier tab identification
+watchEffect(() => {
+  if (state.value.data.message_type) {
+    const statusEmoji = isError.value ? "❌" : "✅";
```
johnsimons, line 105: I wouldn't add emojis to document.title, I think it would look funny given that by default it displays the site icon, so we would endup with double icons

Issue #2511, opened 2025-07-11
ServicePulse doesn't indicate when a too-large audit body was not stored


PhilBastian:
### Symptoms

Messages with no body do not show any indicator that the body wasn't saved.

### Who's affected

Users examining audited messages.

### Root cause

If a message body is larger than `ServiceControl/MaxBodySizeToStore` then it is not persisted as part of auditing. This information is not presented in the UI.

### Confirmed workarounds

None.

<details>
<summary>Original bug report</summary>

#### Description

If a message body is larger than `ServiceControl/MaxBodySizeToStore` then it is not persisted as part of auditing. ServiceInsight displays this information to the user

<img width="1229" height="786" alt="Image" src="https://github.com/user-attachments/assets/77d5ded7-b3eb-4b50-8c38-1cac05549368" />

ServicePulse does not display the same information, instead just showing a blank body

#### Expected behavior

An indication to the user that the message body was too large to be stored

#### Actual behavior

a blank body is displayed

<img width="912" height="375" alt="Image" src="https://github.com/user-attachments/assets/92b02e65-fda2-4d38-ba94-19ac9a0a380b" />

#### Versions

Please list the version of the relevant packages or applications in which the bug exists.        

### Steps to reproduce

- send a message with a body exceeding `ServiceControl/MaxBodySizeToStore` to an endpoint that has auditing enabled.
- view the message in "All Messages" in ServicePulse. The body shows as blank with no indication why

### Relevant log output

```shell

```

#### Additional information

</details>

ramonsmits:
The body can be empty/null, if that is the case I would expect a nice message instead of the current state as it feels like the UI is broken.

afprtclr:
@ramonsmits This has been discussed internally and the conclusion was that showing curly braces is good enough.
The fix will be shipped with release 2.2.0.

PR #2585, opened 2025-08-13
Indicating when the body wasn't stored for the audited message

afprtclr:
For:

- https://github.com/Particular/ServicePulse/issues/2511

image

### DIFF
--- a/src/Frontend/src/components/messages/SagaDiagram/MessageDataBox.vue
+++ b/src/Frontend/src/components/messages/SagaDiagram/MessageDataBox.vue
@@ -30,7 +30,7 @@ const body = computed(() => props.messageData.body.data.value || "");
   <div v-else-if="messageData.body.failed_to_load" class="message-data-box message-data-box-error">
     <span class="message-data-box-text--error">Failed to load message data, there might be a connection issue or the message may no longer be available.</span>
   </div>
-  <div v-else-if="!messageDataLoading && (!messageData.body.data.value || messageData.body.not_found)" class="message-data-box">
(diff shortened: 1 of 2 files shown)


src/Frontend/src/components/messages/BodyView.vue
```diff
@@ -25,6 +25,7 @@ const body = computed(() => bodyState.value.data.value);
     <div v-if="bodyState.not_found" class="alert alert-info">Could not find the message body. This could be because the message URL is invalid or the corresponding message was processed and is no longer tracked by ServiceControl.</div>
     <div v-else-if="bodyState.failed_to_load" class="alert alert-info">Message body unavailable.</div>
     <LoadingSpinner v-else-if="bodyState.loading" />
+    <div v-else-if="body === ''" class="alert alert-info">Body was too large and not stored. Edit ServiceControl/MaxBodySizeToStore to be larger in the ServiceControl configuration.</div>
```
soujay, line 28: Suggestion - it will be good to link to the doc as well 
https://docs.particular.net/servicecontrol/audit-instances/configuration#performance-tuning-servicecontrol-auditmaxbodysizetostore

afprtclr: Fixed both. Please let me know if that looks better.

danielmarbach:
@afprtclr that's interesting. I pushed a commit showing the path that reaches to download the message body from the same endpoint URL, if I'm not mistaken, and therefore should also deal with the 204 returned. I haven't had a chance to wrap my head around the specs but maybe we can also add one or two specs to both code paths

PR #2596, opened 2025-08-21
Update dependencies

johnsimons:
This resolves the issue we identified, where the build hangs.
It also updates all dependencies.


src/Frontend/tsconfig.app.json
```diff
@@ -13,6 +13,7 @@
       "@/*": ["./src/*"],
       "@component-test-utils": ["./test/utils.ts"]
     },
+    "noUncheckedIndexedAccess": false,
```
johnsimons, line 16: This is required because of [this](https://github.com/vuejs/tsconfig/releases/tag/v0.8.0) change to the latest version of @vue/tsconfig

PR #2597, opened 2025-08-22
Improving the connection setup instructions

johnsimons:
This PR improves the rendering of the instructions so that it supports an empty list of instructions, which is added in https://github.com/Particular/ServiceControl/pull/5104.

So for ASQ, we were displaying:
image

and now we display:
image

For MSMQ we were displaying:

and now we display:

For broker transports such as ASB we were displaying:
image

and now we display:
image

This PR is not 100% backwards compatible, so if the user upgrades SC to latest and does not upgrade SP, they will see:
image
which is not too bad.


src/Frontend/src/views/throughputreport/setup/ConnectionSetupView.vue
```diff
@@ -20,7 +20,7 @@ onMounted(async () => {
 
 <template>
   <div class="row">
-    <p v-if="(settingsInfo?.broker_settings.length ?? 0 > 0) || (!isBrokerTransport && useIsMonitoringEnabled())">
+    <p v-if="(settingsInfo?.broker_settings.length ?? 0 > 0) || ((settingsInfo?.monitoring_settings.length ?? 0 > 0) && useIsMonitoringEnabled()) || (settingsInfo?.service_control_settings.length ?? 0 > 0)">
```
WilliamBZA, line 23: This logic is a bit hard for me to parse (especially with the `.monitoring_settings` and `.service_control_settings` differences which looks very similar to the rest to the code).

Is it possible to put this behind a named function to make it more readable?

`v-if="shouldDisplayCollectionInstructions()` or something like that?

johnsimons: I have refactor this, thanks for pointing it out

Issue #2417, opened 2025-05-13
Share utility function between diagrams and other places


johnsimons:
### Describe the suggested improvement

As part of the release of version 2 we identified that therre are quite a bit of comonality between all the diagrams.

Examples:
- Rendering of dates and times
- Rendering of type name, and ways to shorten it using ellipses
- Render of saga name
- Render of icons and tootips for them

PR #2659, opened 2025-10-14
Refactoring code to be more component base

johnsimons:
This PR addresses:

- https://github.com/Particular/ServicePulse/issues/2417.


PhilBastian (approved):
I haven't smoke tested all scenarios, but it looks like a solid refactor

PR #2675, opened 2025-10-22
refactor the remaining service modules which contained internal state into stores/views

PhilBastian:
part of https://github.com/Particular/ServicePulse/issues/1905

### DIFF
--- a/src/Frontend/src/components/customchecks/CustomCheckView.vue
+++ b/src/Frontend/src/components/customchecks/CustomCheckView.vue
@@ -67,6 +67,7 @@ const endpointColor = hexToCSSFilter("#929E9E").filter;
 
 .pa-endpoint {
   position: relative;
+  display: inline-block;
   top: 3px;
   background-image: url("@/assets/endpoint.svg");
   background-position: center;
--- a/src/Frontend/src/components/failedmessages/MessageList.vue
+++ b/src/Frontend/src/components/failedmessages/MessageList.vue
@@ -196,6 +196,7 @@ div.failed-message-data {
 
 .pa-endpoint {
   position: relative;
+  display: inline-block;
   top: 3px;
   background-image: url("@/assets/endpoint.svg");
   background-position: center;


src/Frontend/src/components/configuration/HealthCheckNotifications.vue
```diff
@@ -104,6 +125,22 @@ async function getEmailNotifications() {
...
+async function getResponseOrError(action: () => Promise<Response>, responseStatusTextOverride?: (response: Response) => string) {
+  const responseStatusTextDefault = (response: Response) => response.statusText;
+  const responseStatusText = responseStatusTextOverride ?? responseStatusTextDefault;
+  try {
+    const response = await action();
+    return {
+      message: response.ok ? "success" : `error:${responseStatusText(response)}`,
+    };
+  } catch (err) {
+    console.log(err);
```
johnsimons: Should this be`console.log`?

src/Frontend/src/components/configuration/HealthCheckNotifications.vue
```diff
@@ -83,14 +91,27 @@ async function testEmailNotifications() {
...
 
 async function getEmailNotifications() {
   showEmailConfiguration.value = false;
-  const result = await useEmailNotifications();
+  let result: EmailNotifications | null = null;
+  try {
+    const [, data] = await serviceControlStore.fetchTypedFromServiceControl<EmailNotifications>("notifications/email");
+    result = data;
+  } catch (err) {
+    console.log(err);
```
johnsimons: Should this be`console.log`?

src/Frontend/src/components/PageHeader.vue
```diff
@@ -13,13 +12,17 @@ import DashboardMenuItem from "@/components/dashboard/DashboardMenuItem.vue";
...
 
+const serviceControlStore = useServiceControlStore();
+const { isMonitoringEnabled } = storeToRefs(serviceControlStore);
 // prettier-ignore
 const menuItems = computed(
   () => [
   DashboardMenuItem,
   HeartbeatsMenuItem,
-  ...(isMonitoringEnabled() ? [MonitoringMenuItem] : []),
+  ...(isMonitoringEnabled ? [MonitoringMenuItem] : []),
```
johnsimons: Does this need `.value`?
```suggestion
  ...(isMonitoringEnabled.value ? [MonitoringMenuItem] : []),
```

src/Frontend/src/components/ServiceControlNotAvailable.vue
```diff
@@ -1,10 +1,11 @@
 ﻿<script setup lang="ts">
```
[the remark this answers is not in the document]
johnsimons: So are you saying that we should refactor both and build a single one with slots?

src/Frontend/src/components/configuration/RetryRedirects.vue
```diff
@@ -1,37 +1,37 @@
...
+const { licenseStatus } = licenseStore;
+
+const isExpired = licenseStatus.isExpired;
 
 const loadingData = ref(true);
-const redirects = reactive<{ total: number; data: Redirect[] }>({
-  total: 0,
-  data: [],
-});
+const redirects = redirectsStore.redirects;
```
johnsimons:
```suggestion
const redirects = storeToRefs(redirectsStore.redirects);
```

src/Frontend/src/components/failedmessages/LastTenOperations.vue
```diff
@@ -1,19 +1,21 @@
...
+import { useServiceControlStore } from "@/stores/ServiceControlStore";
 
 const historicOperations = ref<HistoricRetryOperation[]>([]);
 const showHistoricRetries = ref(false);
 
+const serviceControlStore = useServiceControlStore();
+
 async function getHistoricOperations() {
-  const [, data] = await useTypedFetchFromServiceControl<RecoverabilityHistoryResponse>("recoverability/history");
+  const [, data] = await serviceControlStore.fetchTypedFromServiceControl<RecoverabilityHistoryResponse>("recoverability/history");
```
johnsimons: I have been reviewing a few usages of these "fetch/post" methods, and I feel that having these come out of the `serviceControlStore` feel weird.
I was expecting in this specific scenario to have a store for "recoverability/history", where we would do something like `recoverabilityStore.getHistory()`.

src/Frontend/src/components/configuration/RetryRedirects.vue
```diff
@@ -134,7 +129,7 @@ function deleteRedirect(redirect: Redirect) {
 }
 
 async function saveDeletedRedirect() {
-  const result = handleResponse(await deleteFromServiceControl(`redirects/${selectedRedirect.value.message_redirect_id}`));
+  const result = handleResponse(await serviceControlStore.deleteFromServiceControl(`redirects/${selectedRedirect.value.message_redirect_id}`));
```
johnsimons: we have a `redirectsStore`, I think the "delete" should be part of it

src/Frontend/src/components/failedmessages/PendingRetries.vue
```diff
@@ -60,12 +67,11 @@ const sortOptions: SortOptions<GroupOperation>[] = [
   },
 ];
 const periodOptions = ["All Pending Retries", "Retried in the last 2 Hours", "Retried in the last 1 Day", "Retried in the last 7 Days"];
-const isMassTransitConnected = useIsMassTransitConnected();
 
 watch(pageNumber, () => loadPendingRetryMessages());
 
 async function loadEndpoints() {
-  const [, data] = await useTypedFetchFromServiceControl<QueueAddress[]>("errors/queues/addresses");
+  const [, data] = await serviceControlStore.fetchTypedFromServiceControl<QueueAddress[]>("errors/queues/addresses");
```
johnsimons: should this method calls be part of the errorsStore?

src/Frontend/src/components/failedmessages/messageGroupClient.ts
```diff
@@ -0,0 +1,68 @@
...
+  evaluateResponse(response: Response): SuccessResponse | ErrorResponse {
+    return response.ok ? ({} as SuccessResponse) : ({ message: response.statusText } as ErrorResponse);
+  }
+
+  public isError(obj: SuccessResponse | ErrorResponse): obj is ErrorResponse {
+    return (obj as ErrorResponse).message !== undefined;
+  }
+}
+
+export default () => new MessageGroupClient();
```
johnsimons: Is the intend for this to be a singleton, or multiple instances?

src/Frontend/src/views/throughputreport/setup/ConnectionSetupView.vue
```diff
@@ -21,7 +26,7 @@ const needsConfiguration = computed(() => {
   const broker = settingsInfo.value?.broker_settings?.length ?? 0;
   const monitoring = settingsInfo.value?.monitoring_settings?.length ?? 0;
   const serviceControl = settingsInfo.value?.service_control_settings?.length ?? 0;
-  return broker > 0 || (monitoring > 0 && isMonitoringEnabled()) || serviceControl > 0;
+  return broker > 0 || (monitoring > 0 && isMonitoringEnabled) || serviceControl > 0;
```
johnsimons:
```suggestion
  return broker > 0 || (monitoring > 0 && isMonitoringEnabled.value) || serviceControl > 0;
```

src/Frontend/src/components/failedmessages/messageGroupClient.ts
PhilBastian: multiple, so that the store can be accessed in the correct lifecycle step. There's no reason for this to be a singleton as opposed to a transient class instance

src/Frontend/src/components/configuration/HealthCheckNotifications.vue
johnsimons: it would be more correct, also we do that in other places

src/Frontend/src/components/TimeSince.vue
```diff
@@ -32,12 +30,7 @@ function updateText() {
   }
 }
 
-onMounted(() => {
-  interval = window.setInterval(updateText, 5000);
-  updateText();
-});
-
-onUnmounted(() => window.clearInterval(interval));
+watch(() => props.dateUtc, updateText, { immediate: true });
```
johnsimons: This is not going to work because the `props.dateUtc` does not change

johnsimons: We need the timer, because we want to update the relative time

src/Frontend/src/stores/ThroughputStore.ts
```diff
@@ -29,11 +29,12 @@ export const useThroughputStore = defineStore("ThroughputStore", () => {
     // if Audit connection test fails, we will return true.
     // the connection test will return true if there are no Audit instances configured.
     if (!testResults.value?.audit_connection_result.connection_successful) {
+      //TODO: should this be a warning rather than an error?
```
johnsimons: we have no concept of warning in this scenario

src/Frontend/src/components/messages/MessageView.vue
```diff
@@ -93,7 +93,7 @@ watch(
   },
   { immediate: true }
 );
-const endpointColor = hexToCSSFilter("#929E9E").filter;
+const endpointColor = hexToCSSFilter("#777F7F").filter;
```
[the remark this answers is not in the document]
PhilBastian: at some stage this colour got out of sync with the other icons

src/Frontend/src/components/configuration/HealthCheckNotifications.vue
```diff
@@ -55,6 +55,7 @@ async function toggleEmailNotifications() {
     emailToggleSuccessful.value = true;
   } else {
     emailToggleSuccessful.value = false;
+    console.log(result.message);
```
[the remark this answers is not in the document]
PhilBastian: not really; it just seemed strange to me that there's code to craft these meaningful failure messages and those messages are never used

src/Frontend/src/stores/ThroughputStore.ts
PhilBastian: we can create one though... we have the warning triangle used elsewhere in the app

src/Frontend/src/components/configuration/HealthCheckNotifications.vue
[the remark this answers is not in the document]
johnsimons: maybe raise a bug/improvement about it?

src/Frontend/src/components/TimeSince.vue
PhilBastian: re-put the timer in but left the watch in place for props changes

src/Frontend/src/components/messages/MessageView.vue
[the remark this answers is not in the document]
PhilBastian: i'll check and fix the remaining filter references. We can't use CSS variables here, since it's done in code rather than styles, so if other styles change then the author needs to remember to change the relevant filters.
It is possible to achieve what we're doing here with styles, but at the moment Safari doesn't render it correctly

src/Frontend/src/stores/ServiceControlStore.ts
```diff
@@ -0,0 +1,191 @@
...
+
+export const useServiceControlStore = defineStore("ServiceControlStore", () => {
+  const serviceControlUrl = ref<string | null>();
+  const monitoringUrl = ref<string | null>();
+
+  const isMonitoringDisabled = computed(() => monitoringUrl.value == null || monitoringUrl.value === "" || monitoringUrl.value === "!");
+  const isMonitoringEnabled = computed(() => !isMonitoringDisabled.value);
+
+  function getServiceControlUrl() {
+    if (!serviceControlUrl.value) refresh();
```
johnsimons: We should throw here to prevent weird behaviour later on.
```suggestion
    if (!serviceControlUrl.value) {
       refresh();
    }
    if (!serviceControlUrl.value) {
      throw new Error("Service Control URL is not configured");
    }
```

src/Frontend/src/stores/ServiceControlStore.ts
```diff
@@ -0,0 +1,191 @@
...
+    if (!response.ok) throw new Error(response.statusText ?? "No response");
+    const data = await response.json();
+
+    return [response, data];
+  }
+
+  async function fetchTypedFromMonitoring<T>(suffix: string): Promise<[Response?, T?]> {
+    if (isMonitoringDisabled.value) {
+      return [];
+    }
```
johnsimons: In other parts of the code the caller is enforcing whether it should call the function or not, I think we should do the same here.

```suggestion
```

PhilBastian:
the refactoring to remove the service modules is done. After discussion, the remaining work to extract more stores, rather than relying on servicecontrolstore in most places for calling the servicecontrol API, will be done as separate PRs

src/Frontend/src/components/messages/MessageView.vue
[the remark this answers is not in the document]
PhilBastian: https://github.com/Particular/ServicePulse/pull/2689

PR #2714, opened 2025-11-18
Patch npm packages

johnsimons:
Updated packages and added override to glob to fix high-severity vulnerabilities - https://github.com/advisories/GHSA-5j98-mcp5-4vw2 

Have also updated VueDatePicker to v12 following  https://vue3datepicker.com/migration/from-v11#migrating-from-v11

### DIFF
--- a/src/Frontend/package.json
+++ b/src/Frontend/package.json
@@ -29,7 +29,7 @@
     "@tinyhttp/content-disposition": "2.2.2",
     "@vue-flow/controls": "1.1.3",
     "@vue-flow/core": "1.47.0",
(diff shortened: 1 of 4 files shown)


src/Frontend/package.json
```diff
@@ -49,39 +49,42 @@
...
+    "vite-plugin-vue-devtools": "8.0.5",
+    "vitest": "4.0.10",
+    "vue-tsc": "3.1.4"
   },
   "msw": {
     "workerDirectory": [
       "public"
     ]
+  },
+  "overrides": {
```
PhilBastian: how will we know when to remove this override?

PR #2725, opened 2025-11-21
Introduced a monitoringClient and serviceControlClient

johnsimons:
This PR separates the client APIs from the store.
So, in this PR we make the separation to use the store as just a state management store, and the calling of the REST api is delegated to the client api.

### DIFF
--- a/src/Frontend/src/components/configuration/EndpointConnection.vue
+++ b/src/Frontend/src/components/configuration/EndpointConnection.vue
@@ -3,14 +3,9 @@ import { onMounted, ref } from "vue";
 import LicenseNotExpired from "../LicenseNotExpired.vue";
 import ServiceControlAvailable from "../ServiceControlAvailable.vue";
 import CodeEditor from "@/components/CodeEditor.vue";
-import serviceControlClient from "@/components/serviceControlClient";
+import serviceControlClient, { ServiceControlInstanceConnection } from "@/components/serviceControlClient";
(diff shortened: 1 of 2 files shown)


src/Frontend/src/components/configuration/EndpointConnection.vue
```diff
@@ -96,19 +84,10 @@ async function serviceControlConnections() {
 
 async function getServiceControlConnection() {
```
PhilBastian: should move to the SC client?

src/Frontend/src/stores/MessageStore.ts
```diff
@@ -73,20 +73,23 @@ export const useMessageStore = defineStore("MessageStore", () => {
...
   const { configuration } = storeToRefs(configStore);
   const error_retention_period = computed(() => dayjs.duration(configuration.value?.data_retention?.error_retention_period ?? "PT0S").asHours());
 
-  watch(serviceControlUrl, loadConfig, { immediate: true });
-  async function loadConfig() {
-    if (!serviceControlUrl.value) return;
-    const [, data] = await serviceControlStore.fetchTypedFromServiceControl<EditAndRetryConfig>("edit/config");
-    edit_and_retry_config.value = data;
-  }
+  serviceControlClient
```
PhilBastian: this can still be done as async await, just by creating an encapsulating async function and then calling that function

PhilBastian (approved):

PR #2786, opened 2026-01-15
Enable Integrated Mode

mikeminutillo:
Alternative to:

- #2782

Closes:

- #2848 

Integrated mode is intended for bundling ServicePulse with ServiceControl. The core ServicePulse application is packaged separately and [hosted out of a ServiceControl Error instance](https://github.com/Particular/ServiceControl/pull/5260) (when enabled by configuration).

Documentation PR:

- https://github.com/Particular/docs.particular.net/pull/8073

- Separates the ServicePulse application from it's host to enable integrating the application with other hosts
- Adds an application setting to show if it is running in an integrated mode
  - Do not allow changing the ServiceControl URL
  - Do not show the ServicePulse version

## Reviewer Checklist

### DIFF
--- a/src/Frontend/test/specs/monitoring/sorting-endpoints.spec.ts
+++ b/src/Frontend/test/specs/monitoring/sorting-endpoints.spec.ts
@@ -1,5 +1,6 @@
 import { expect } from "vitest";
 import { test, describe } from "../../drivers/vitest/driver";
+import { waitFor } from "@testing-library/vue";
 import { groupEndpointsBy } from "./actions/groupEndpointsBy";
 import { endpointGroupNames } from "./questions/endpointGroupNames";
 import { endpointGroup } from "./questions/endpointGroup";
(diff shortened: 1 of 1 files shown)


src/Frontend/src/components/serviceControlClient.ts
```diff
@@ -96,6 +96,10 @@ class ServiceControlClient {
   }
 
   private getUrl() {
+    if (window.defaultConfig && window.defaultConfig.isEmbedded && window.defaultConfig.service_control_url && window.defaultConfig.service_control_url.length) {
```
PhilBastian:
```suggestion
    if (window.defaultConfig?.isEmbedded && window.defaultConfig.service_control_url?.length) {
```

src/ServicePulse.Core/ConstantsFile.cs
```diff
@@ -0,0 +1,41 @@
...
+        var version = GetVersionInformation();
+
+        var constantsFile = $$"""
+window.defaultConfig = {
+  default_route: '{{settings.DefaultRoute}}',
+  version: '{{version}}',
+  service_control_url: '{{settings.ServiceControlUrl}}',
+  monitoring_urls: ['{{settings.MonitoringUrl ?? "!"}}'],
+  showPendingRetry: {{(settings.ShowPendingRetry ? "true" : "false")}},
+  isEmbedded: {{(settings.IsEmbedded ? "true" : "false")}}
```
PhilBastian: neater?
```suggestion
  isEmbedded: {{(settings.IsEmbedded.ToString().ToLower())}}
```

src/ServicePulse.Core/ServicePulseSettings.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// Flag to indicate if ServicePulse is running in embedded mode.
+    /// </summary>
+    public required bool IsEmbedded { get; init; }
+
+    /// <summary>
+    /// Loads the settings from environment variables.
+    /// </summary>
+    public static ServicePulseSettings GetFromEnvironmentVariables()
+    {
+        var serviceControlUrl = Environment.GetEnvironmentVariable("SERVICECONTROL_URL") ?? "http://localhost:33333/api/";
```
PhilBastian: suggest use of StringBuilder

mikeminutillo: I looked at this and do not believe it is worth it in this case. We keep needing the current value of the string, and for a stringbuilder we'd need to do that comparison on a character by character basis. We only need to add three strings in the case where a customer provides a url without `/` and without `api/`. e.g. Something like `localhost:80`.

PR #2881, opened 2026-03-12
Replace collapse button with accordion-style chevron toggle

dvdstelt:
Fixes:

- https://github.com/Particular/ServicePulse/issues/2888

The separate pill button is replaced with a rotating chevron icon inline with the section title, following the standard accordion pattern. The expand state uses the same visual affordance, keeping the interaction consistent and familiar.

collapse

## Reviewer Checklist


dvdstelt:
I just selected a few random people that I know somewhat regularly work on ServicePulse

PhilBastian (approved):

warwickschroeder (approved):

PR #2885, opened 2026-03-13
Fix all messages / audit view not loading data when selected

ramonsmits:
ALWAYS when I open all messages view nothing happens. Switching do a different tab and back still no data until I do a full page reload.

> [!NOTE]
> This is a suggestion by Claude so this really needs to be reviewed by someone that understands this. I don't really have an understanding how the refresh is designed.

## Background info

The auto-refresh composable was mutating the store's refresh method via Object.assign, replacing it with checkForSuccessfulMessages. This broke AuditList's full fetch since it also called store.refresh().

Add a customRefresh parameter to useStoreAutoRefresh so the lightweight checkForSuccessfulMessages can be used for polling without mutating the store.

## Reviewer Checklist

### DIFF
--- a/src/Frontend/src/composables/useAuditStoreAutoRefresh.ts
+++ b/src/Frontend/src/composables/useAuditStoreAutoRefresh.ts
@@ -1,10 +1,6 @@
 import { useAuditStore } from "@/stores/AuditStore";
 import { useStoreAutoRefresh } from "./useAutoRefresh";
 
-// Override the refresh method to use checkForSuccessfulMessages, which is more lightweight
-const useAuditStoreWithRefresh = () => {
-  const store = useAuditStore();
-  return Object.assign(store, { refresh: store.checkForSuccessfulMessages });
-};
-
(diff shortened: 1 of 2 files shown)


johnsimons:
@ramonsmits thanks for pinging us, ClaudeCode is right, `Object.assign` is the culprit in this bug.
We looked around, and there is [another usage](https://github.com/Particular/ServicePulse/blob/ba031f2845837b92d9530757802cbc1eabef2a27/src/Frontend/src/composables/useMonitoringStoreAutoRefresh.ts#L7) of it that causes a similar bug!

Even though this PR fixes the bug, it would be better to consolidate all this functionality into the existing pinia store and use existing patterns.
So we have created [another PR](https://github.com/Particular/ServicePulse/pull/2892) that addresses both of the bugs.

dvdstelt (approved):

johnsimons:
This has been superseded by:

- https://github.com/Particular/ServicePulse/pull/2892

PR #2893, opened 2026-03-17
Add activity filter for endpoint message types

### LINKED ISSUE #2890: Add filter 'Active only' to only show active Message Types in Monitoring
#### Is your improvement related to a problem? Please describe.
When I click thru in the Monitoring dashboard to a specific endpoint, the active processing Message Types or not in the top of the list. I need to go thru the pages to see counters for the message types that are actively being processed at the moment.

#### Describe the suggested solution
Be able to filter out the Message Types by "Active only" so you can intentionally narrow the list.

Preferable this shows the active messages for the selected time range (1m 5m 10m 15m 3m 1h), but a fixed period like the last hour is also ok for me.

Nice to have: When new message types are being processed in the time range, add them in the filtered Active Message Types list automatically.

Alternative a secondary section above the table called "Active message types" could be added (a snapshot for the selected time range).
        
### Additional Context

When opening the endpoint page:
Image

I need to click to page 4 to see active message types.
Image

### Additional Context

_No response_

dvdstelt:
_Closes #2890_

Adds a client-side filter to the Message Types tab on the Endpoint Details monitoring view, allowing users to show only message types with recent throughput activity.

  - Filter options: All (default) | 1 min | 5 min | 15 min
  - Filters by checking only the trailing portion of the throughput data points matching the selected time window
  - Filter options adapt to the selected history period (e.g. "15 min" is hidden when viewing 1 or 5 minute history)
  - Pagination resets to page 1 and reflects the filtered count when a filter is active

servicepulse

## Reviewer Checklist

### DIFF
--- a/src/Frontend/src/components/monitoring/EndpointMessageTypes.vue
+++ b/src/Frontend/src/components/monitoring/EndpointMessageTypes.vue
@@ -13,9 +13,11 @@ import ColumnHeader from "@/components/ColumnHeader.vue";
 import { CriticalTime, MessageType, ProcessingTime, ScheduledRetries, Throughput } from "@/resources/MonitoringResources";
 import FAIcon from "@/components/FAIcon.vue";
 import { faWarning } from "@fortawesome/free-solid-svg-icons";
+import { useMonitoringHistoryPeriodStore } from "@/stores/MonitoringHistoryPeriodStore";
 
 const monitoringStore = useMonitoringEndpointDetailsStore();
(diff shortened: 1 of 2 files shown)


jasontaylordev (approved):

warwickschroeder (approved):

johnsimons (changes requested):
https://github.com/Particular/ServicePulse/pull/2893#issuecomment-4086291023

PR #2907, opened 2026-03-20
Add hex view for message body

ramonsmits:
Screencast From 2026-03-20 15-54-40

## Summary

- Adds a hex view tab alongside the formatted view in the message body panel
- Virtual-scrolled hex dump with offset, hex bytes grouped in 8-byte lanes, and ASCII representation
- Responsive layout that adjusts bytes-per-row based on available width
- Click-drag selection across hex and ASCII columns with copy to clipboard
- Auto-switches to hex view when message body parsing fails (invalid JSON/XML), showing a warning banner
- Wraps JSON/XML formatting in try/catch so parse failures surface a `parse_failed` flag instead of crashing
- Mock scenario `recoverability-available` updated with a test message for hex view

## How to test

Run `VITE_MOCK_SCENARIO=recoverability-available npm run dev:mocks` to test with mock data

### DIFF
--- a/src/Frontend/src/components/messages/HexView.vue
+++ b/src/Frontend/src/components/messages/HexView.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
-import { ref, computed, onMounted, onBeforeUnmount, watch } from "vue";
-import { useResizeObserver } from "@vueuse/core";
+import { ref, computed, onMounted, watch } from "vue";
+import { useMousePressed, useResizeObserver } from "@vueuse/core";
 import CopyToClipboard from "@/components/CopyToClipboard.vue";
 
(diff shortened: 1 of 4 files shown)


src/Frontend/src/components/messages/HexView.vue
```diff
@@ -0,0 +1,411 @@
...
+}
+
+function onDocumentMouseUp() {
+  endSelection();
+}
+
+let resizeObserver: ResizeObserver | null = null;
+
+onMounted(() => {
+  document.addEventListener("mouseup", onDocumentMouseUp);
```
johnsimons: Consider using https://vueuse.org/core/useMousePressed/#usemousepressed

PR #3026, opened 2026-06-09
Url encoding fix

### LINKED ISSUE #3005: An error occurred while trying to load the message.
## Symptoms

When navigating to a failed message, the following error may be displayed:

> An error occurred while trying to load the message. Please check the ServiceControl logs to learn what the issue is.

## Who's affected

Users of ServicePulse with MSMQ-based systems.

## Root cause

Incorrect handling of `\` characters in native MSMQ message ids caused invalid requests for message details.

## Confirmed workarounds

None

#### Description
After updating ServicePulse from version 1.39 (hosted on IIS) to 2.6.0 (hosted from a Docker image), an error appears for some messages.

#### Expected behavior

Message details show up

#### Actual behavior
 After entering the details, the following error occurs:

**An error occurred while trying to load the message. Please check the ServiceControl logs to learn what the issue is.**

#### Versions
2.6.0 (hosted from a Docker image)

### Steps to reproduce

Clicking on a message in the error list. For some messages, details are displayed, for others an error occurs.

### Relevant log output

```shell
There is no log for this event in ServicePulse and on the ServiceControl side
```

#### Additional information
The same message appears correctly on an older version of ServicePulse. I compared the network traffic to ServiceControl on both versions, and it looks identical.

There are two calls:

ServicePulse/api/errors/last/7670683f-e09d-d789-4cc0-572280aa34d0
with a 200 response

and

ServicePulse/api/messages/search/e900c065-9aa8-4e2b-be37-d05b87708091/113074851
with a 404 response.

The response from the first call returns something like this: 
{
  "id": "7670683f-e09d-d789-4cc0-572280aa34d0",
  "message_type": "***************************************",
  "time_sent": "2026-03-06T09:56:32.30921+01:00",
  "is_system_message": false,
  "exception": {
    "exception_type": "S****************************",
    "message": "**********: **********************************************",
    "source": "*****************",
    "stack_trace": "***************"
  },
  "message_id": "e900c065-9aa8-4e2b-be37-d05b87708091\\113074851",
  "number_of_processing_attempts": 1,
  "status": "unresolved",
  "receiving_endpoint": {
    "name": "********************",
    "host_id": "5dc0ecfc-045a-bb92-8e64-61586f7ab6be",
    "host": "*********"
  },
  "queue_address": "******************************",
  "time_of_failure": "2026-03-06T08:56:33.215548Z",
  "last_modified": "2026-03-06T08:56:33.2719085Z",
  "edited": false,
  "edit_of": ""
}

The error was noticed within one endpoint for one message type.

hazel-bohon:
* Fixes #3005

In MSMQ, Message ID (native message id) can contain `\` and other characters which must be url encoded in order to be correctly passed to the ServiceControl. This PR adds url encoding when required to allow.

## Reviewer Checklist


DavidBoike (approved):

src/Frontend/test/preconditions/recoverability.spec.ts
cquirosj, line 1: The recoverability.ts file under test/preconditions is testing infrastructure; it doesn't need a spec.ts file to test the testing infrastructure. So this file can be deleted

src/Frontend/test/preconditions/recoverability.ts
cquirosj, line 1: The test/recoverability.ts is testing the infrastructure. Adding the call to encodeURIComponent is technically a testing anti-pattern because it duplicates implementation details of the production code. If at any point the scenario in which the messageId needs to be encoded is exercised, the test cases using this testing infrastructure class will add hard-coded routes with values already encoded. In conclusion, the changes to this file can be discarded from this PR.

src/Frontend/src/stores/MessageStore.spec.ts
cquirosj, line 1: I wonder if this test is exercising more of the store than necessary for the behavior being introduced.

The production change in MessageStore is specifically about how the `messages/search/:messageId` API path is built. To test that through `loadMessage`, the test has to create the real store and mock several unrelated dependencies.

Would it be simpler to extract the endpoint construction into a small helper, e.g. `messageSearchPath(messageId)`, and test that directly? Then the store can call that helper, and the test can focus only on the fact that transport-defined message IDs like MSMQ `{GUID}/{Counter}` are encoded before being used as a path segment.

A cleaner option could be to extract the API path construction into a small helper and test that directly:

```ts
const messageSearchPath = (messageId: string) =>
  `messages/search/${encodeURIComponent(messageId)}`;
```

Then `MessageStore.ts` becomes:

```ts
const [, data] =
  await serviceControlClient.fetchTypedFromServiceControl<Message[]>(
    messageSearchPath(messageId)
  );
```

And the test becomes:

```ts
test.each([
  ["id\\with\\backslash", "id%5Cwith%5Cbackslash"],
  ["id/with/slash", "id%2Fwith%2Fslash"],
])("message search path encodes message ID %s", (messageId, encodedMessageId) => {
  expect(messageSearchPath(messageId)).toBe(
    `messages/search/${encodedMessageId}`
  );
});
```

src/Frontend/src/router/routeLinks.ts
```diff
@@ -34,7 +34,7 @@ const messagesLinks = (root: string) => {
   return {
     root,
     failedMessage: { link: (id: string) => `${root}/${id}`, template: "/messages/:id" },
-    successMessage: { link: (messageId: string, id: string) => `${root}/${messageId}/${id}`, template: "/messages/:messageId/:id" },
+    successMessage: { link: (messageId: string, id: string) => `${root}/${encodeURIComponent(messageId)}/${id}`, template: "/messages/:messageId/:id" },
```
cquirosj, line 38: I would not necessarily expand this to every dynamic value unless we know those values can contain URL-reserved characters. But I think it would be useful to make the reason explicit, either with a small helper like `encodeMessageIdForRoute` or a short comment, so this does not look like arbitrary encoding in only some routes.

```
const encodeMessageId = (messageId: string) =>
  encodeURIComponent(messageId);
```

or a comment

```
// Native message IDs are transport-defined. For example, MSMQ IDs can contain `/`,
// so the value must be encoded before being used as a URL path segment.
```

src/Frontend/src/stores/MessageStore.ts
```diff
@@ -152,7 +152,7 @@ export const useMessageStore = defineStore("MessageStore", () => {
     state.not_found = headers.value.not_found = false;
 
     try {
-      const [, data] = await serviceControlClient.fetchTypedFromServiceControl<Message[]>(`messages/search/${messageId}`);
+      const [, data] = await serviceControlClient.fetchTypedFromServiceControl<Message[]>(`messages/search/${encodeURIComponent(messageId)}`);
```
cquirosj, line 156: I would not necessarily expand this to every dynamic value unless we know those values can contain URL-reserved characters. But I think it would be useful to make the reason explicit, either with a small helper like `encodeMessageIdForRoute` or a short comment, so this does not look like arbitrary encoding in only some routes.

```
const encodeMessageId = (messageId: string) =>
  encodeURIComponent(messageId);
```

or a comment

```
// Native message IDs are transport-defined. For example, MSMQ IDs can contain `/`,
// so the value must be encoded before being used as a URL path segment.
```

cquirosj (changes requested):

src/Frontend/src/stores/MessageStore.spec.ts
hazel-bohon: The extracted method seems like too small of a piece of code to be worth testing. I think that having the two mocks to validate that the url encoding is done as part of making the `LoadMessage` calls helps to keep the test more implementation agnostic.

cquirosj (approved):

PR #3030, opened 2026-06-12
Display usage monthly throughput graphs for usage screen when data is available

PhilBastian:
Adds a graph to each detected queue on the usage page to show throughput information over the past year. This is designed to assist with classifying the endpoint type by providing usage context

image

## Reviewer Checklist

### DIFF
--- a/src/Frontend/src/views/throughputreport/endpoints/ThroughputGraph.vue
+++ b/src/Frontend/src/views/throughputreport/endpoints/ThroughputGraph.vue
@@ -1,0 +1,104 @@
+<script setup lang="ts">
+import type { MonthlyThroughput } from "@/resources/EndpointThroughputSummary";
+import { computed } from "vue";
+
+const props = defineProps<{ data: MonthlyThroughput[] }>();
+
+const date = new Date();
+date.setMonth(date.getMonth() - 13);
+const reportPeriod = Array.from({ length: 13 }).map((_) => {
+  date.setMonth(date.getMonth() + 1);
+  return `${date.getFullYear()}-${(date.getMonth() + 1).toString().padStart(2, "0")}`;
+});
+
+const maxValue = computed(() => Math.max(...[...props.data.map(({ throughput }) => throughput), 1]));
+const maxValueText = computed(() => {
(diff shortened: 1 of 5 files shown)


src/Frontend/src/views/throughputreport/endpoints/ThroughputGraph.vue
```diff
@@ -0,0 +1,103 @@
...
+date.setMonth(date.getMonth() - 13);
+const reportPeriod = Array.from({ length: 13 }).map(() => {
+  date.setMonth(date.getMonth() + 1);
+  return `${date.getFullYear()}-${(date.getMonth() + 1).toString().padStart(2, "0")}`;
+});
+
+const maxValue = computed(() => Math.max(...[...props.data.map(({ throughput }) => throughput), 1]));
+const maxValueText = computed(() => {
+  if (maxValue.value >= 1_000_000) return `${(Math.round((maxValue.value * 10) / 1_000_000) / 10 + 0.1).toFixed(1)}m`;
+  if (maxValue.value >= 10_000) return `${(Math.round((maxValue.value * 100) / 1_000) / 100 + 0.01).toFixed(2)}k`;
```
mikeminutillo: Should `m` and `k` be upper-cased?

mikeminutillo (approved):

src/Frontend/src/views/throughputreport/endpoints/ThroughputGraph.vue
PhilBastian: k, no. m debatable but I found `useFormatLargeNumber` which uses M already so I'll change it to that

PhilBastian:
```suggestion
  if (maxValue.value >= 1_000_000) return `${(Math.round((maxValue.value * 10) / 1_000_000) / 10 + 0.1).toFixed(1)}M`;
  if (maxValue.value >= 10_000) return `${(Math.round((maxValue.value * 100) / 1_000) / 100 + 0.01).toFixed(2)}k`;
```

PR #3035, opened 2026-06-19
Hide tabs based on user authorization levels

WilliamBZA:
## Summary

Gates the ServicePulse UI on the set of ServiceControl API routes the signed-in user is actually allowed to call (the `my/routes` manifest). Navigation items and Configuration sub-tabs the user cannot reach are hidden; action buttons they cannot use are disabled with an explanatory tooltip (not hidden), so the UI stays discoverable.

Gating is fail-open: it only applies once authorization is enabled, the user is authenticated, and the manifest has loaded. In every other state the full UI is shown, so an unreachable or slow ServiceControl never locks a user out.

This PR consolidates the work previously split across #3041 and #3044. It supersedes the earlier permission-string gating model, which was introduced and then removed on this branch in favour of the route-based approach.

## How it works

- **Route registry** (`composables/apiRoutes.ts`): maps each gated UI capability to the ServiceControl HTTP route it represents. This is the single point of coupling to ServiceControl's route surface; the UI gates on routes it already calls, never on internal permission names.
- **Structural matching** (`composables/routeMatching.ts`): normalizes a (method, path) pair into a comparison key with route parameters collapsed to `{}`, so matching couples to method plus path structure and survives server-side parameter renames.
- **Manifest store** (`stores/AllowedRoutesStore.ts`): fetches `my/routes` from the Primary and Monitoring instances, merges them, and keeps a per-store in-flight guard so concurrent callers share one request. Per-instance fail-open: a failed or non-array response is ignored rather than fataled.
- **Gating primitive** (`composables/useAllowedRoutes.ts`): exposes `canCall` and `canAnyCall`. Resource-owning stores call these and surface capability getters; templates bind to those getters.
- **PermissionGate wrapper** (`components/PermissionGate.vue`): reusable component that disables a slotted action and shows a tooltip with the reason when the user lacks the route, instead of removing the control.

## What is gated

- Navigation menu (`PageHeader.vue`) and Configuration sub-tabs
  (`ConfigurationView.vue`): hidden when no backing route is allowed.
- Action buttons, disabled with tooltip: failed messages, deleted messages and message groups, custom-check dismiss, heartbeat endpoint instances, redirects, health-check notifications (manage and test), edit and retry, and monitored endpoint deletion.
- Resource-owning stores expose capability getters: `HealthChecksStore`, `HeartbeatInstancesStore`, `MessageStore`, `MonitoringEndpointDetailsStore`, `RedirectsStore`, `ThroughputStore`.
- A new User Permissions configuration page (`UserPermissions.vue`) renders the capabilities derived from the user's allowed routes.

## Design decisions

- Fail-open everywhere: uncertainty (auth disabled, not yet authenticated, manifest not loaded) shows the UI rather than hiding it.
- Gate on routes the UI already calls, decoupling the frontend from ServiceControl's internal permission model.
- Structural, parameter-insensitive route keys so the contract is method plus path shape, not exact parameter names.
- The manifest is stored as a Map (not a Set) so a future per-route resource scope can be carried per entry.
- The startup fetch surfaces a 403 denial reason so an explicitly forbidden user gets a clear message.

## Backend dependency

Requires ServiceControl to serve the `my/routes` allowed-route manifest on the
Primary instance, and the equivalent on the Monitoring instance.

## Reviewer Checklist

### DIFF
--- a/src/Frontend/src/components/configuration/UserPermissions.vue
+++ b/src/Frontend/src/components/configuration/UserPermissions.vue
@@ -1,0 +1,188 @@
+<script lang="ts">
+import { ApiRoutes } from "@/composables/apiRoutes";
+
+// Static capability groupings: each area lists the API routes the user may or may not be able to call.
+// Permissions are derived at render time from canCall — no permission strings are involved.
+// Defined at module scope (outside setup) so it is shared across component instances.
+export const groups = [
+  {
+    area: "Failed messages",
+    caps: [
+      { label: "View", ref: ApiRoutes.viewFailedMessages },
+      { label: "Retry", ref: ApiRoutes.retryMessage },
+      { label: "Edit", ref: ApiRoutes.editMessage },
+      { label: "Delete", ref: ApiRoutes.deleteMessage },
+      { label: "Restore", ref: ApiRoutes.restoreMessage },
+    ],
+  },
+  {
+    area: "Recoverability groups",
+    caps: [
+      { label: "Retry group", ref: ApiRoutes.retryGroup },
+      { label: "Delete group", ref: ApiRoutes.deleteGroup },
+      { label: "Restore group", ref: ApiRoutes.restoreGroup },
+    ],
+  },
+  {
+    area: "Audit",
+    caps: [{ label: "View", ref: ApiRoutes.viewAuditMessages }],
+  },
+  {
+    area: "Monitoring",
+    caps: [
+      { label: "View", ref: ApiRoutes.viewMonitoredEndpoints },
+      { label: "Remove endpoint", ref: ApiRoutes.deleteMonitoredEndpoint },
+    ],
+  },
+  {
+    area: "Custom checks",
+    caps: [
+      { label: "View", ref: ApiRoutes.viewCustomChecks },
+      { label: "Dismiss", ref: ApiRoutes.dismissCustomCheck },
+    ],
+  },
+  {
+    area: "Event log",
+    caps: [{ label: "View", ref: ApiRoutes.viewEventLog }],
+  },
+  {
+    area: "Configuration — License",
+    caps: [{ label: "View", ref: ApiRoutes.viewLicense }],
+  },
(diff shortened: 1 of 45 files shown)


ramonsmits (commented):
Added some improvement on:

- https://github.com/Particular/ServicePulse/pull/3036

src/Frontend/src/components/configuration/UserPermissions.vue
```diff
@@ -0,0 +1,188 @@
+<script lang="ts">
+import { ApiRoutes } from "@/composables/apiRoutes";
+
+// Static capability groupings: each area lists the API routes the user may or may not be able to call.
+// Permissions are derived at render time from canCall — no permission strings are involved.
+// Defined at module scope (outside setup) so it is shared across component instances.
+export const groups = [
+  {
+    area: "Failed messages",
+    caps: [
```
WilliamBZA: What does `caps` stand for/mean?

dvdstelt: `caps` is short for capabilities, the individual actions listed under each area on the User Permissions page.

Each area (e.g. "Recoverability groups") has a caps array where every entry pairs a human-readable label with the ref to the ServiceControl API route that action calls, taken from the `ApiRoutes` registry:
```
{ label: "Retry group", ref: ApiRoutes.retryGroup }
```
At render time the page maps each capability through canCall(ref), which checks whether that route is in the user's `my/routes` manifest, and shows a ✓ or ✗ accordingly. So caps is just the per-area list of "things you might be able to do here," and the table resolves each one to `allowed/not-allowed` for the current user.

dvdstelt: We could rename it or add a comment.

WilliamBZA: I created a typescript type for this and renamed

dvdstelt (approved):

PR #3042, opened 2026-06-24
Re-authenticate instead of going blank when the session is lost

dvdstelt:
## Problem
When OIDC auth is enabled and the access token expires (and silent renewal does
not succeed), `App.vue` renders nothing and the user has to manually refresh the
browser to recover.

`App.vue` gates the entire app behind `shouldShowApp` (`authEnabled`,
`isAuthenticated`, `isAnonymousRoute`). On token expiry, `oidc-client-ts` clears
the token, `isAuthenticated` flips to `false`, and that `v-if` renders nothing.
Nothing re-triggers authentication, so the page stays blank until a refresh
re-runs the auth flow on mount.

## Fix
Watch for the session being lost while the app is running and re-trigger
authentication via the existing `useAuth().authenticate()` flow:
- With a live identity-provider session this is a silent redirect round-trip and
  the user keeps working.
- Otherwise the user lands on the provider's login page.
- Skipped on anonymous routes (e.g. the logged-out page) and while a sign-in is
  already in progress, to avoid loops.

## Tests
- `App.spec.ts`: re-authenticates on token loss; does not while already
  authenticating; does not on an anonymous route.
- Existing authentication integration specs still pass; typecheck and lint clean.

## Notes
- this is a pre-existing auth/session bug, so it targets `master` directly.
- Renewal reliability itself (so the token rarely lapses in the first place) is an
  identity-provider configuration concern (the provider must grant `offline_access`
  so a refresh token is issued); ServicePulse already requests it.


src/Frontend/src/App.vue
```diff
@@ -8,16 +8,28 @@ import LicenseNotifications from "@/components/LicenseNotifications.vue";
...
 const isAnonymousRoute = computed(() => route.meta?.allowAnonymous === true);
 const shouldShowApp = computed(() => !authEnabled.value || isAuthenticated.value || isAnonymousRoute.value);
 // Show full app layout (header, footer, notifications) only when authenticated or auth is disabled
 const shouldShowFullLayout = computed(() => !authEnabled.value || isAuthenticated.value);
+
+// If the session is lost while the app is running (e.g. the access token expired and silent
+// renewal failed), re-trigger authentication instead of rendering a blank page. With a live
+// identity-provider session this is a silent redirect round-trip; otherwise the user lands on
+// the provider's login page. Without this the app renders nothing until a manual refresh.
+watch([authEnabled, isAuthenticated, isAnonymousRoute], ([enabled, authenticated, anonymous]) => {
```
PhilBastian: I'm thinking that all this logic, including the isAnonymousRoute, isn't actually the responsibility of the App page and should instead be in the authStore. Pulling in @warwickschroeder for comment.

warwickschroeder: Agreed, it probably shouldn't live here. I mean it'll work but its better handled in the auth domain. We are already handling [OIDC events in useAuth.ts](https://github.com/Particular/ServicePulse/blob/eeeae2a669554939190db554af7a065070bfb382/src/Frontend/src/composables/useAuth.ts#L36-L42). These could be used to pickup on the "token exired" or "silent renewal error" events and then reauthenticate. This way we arent "watching" a proxy state, but rather reacting to the actual OIDC event.

src/Frontend/src/App.spec.ts
johnsimons, line 1: Given we did not touch App.vue, there is no need for this test

src/Frontend/src/App.vue
johnsimons, line 1: No need to change this file

src/Frontend/src/composables/useAuth.ts
```diff
@@ -12,6 +13,24 @@ let userManager: UserManager | null = null;
...
+    }
+    if (window.location.hash === `#${routeLinks.loggedOut}`) {
+      return;
+    }
+    authStore.setAuthenticating(true);
+    userManager?.signinRedirect().catch((error) => {
+      authStore.setAuthenticating(false);
+      logger.error("Re-authentication after session loss failed:", error);
+    });
+  }
```
johnsimons:
```suggestion
  async function reauthenticate() {
    if (authStore.isAuthenticating) {
      return;
    }
    if (window.location.hash === `#${routeLinks.loggedOut}`) {
      return;
    }
    authStore.setAuthenticating(true);
    try {
      await userManager?.signinRedirect();
    } catch (error) {
      logger.error("Re-authentication after session loss failed:", error);
      authStore.setAuthError(error instanceof Error ? error.message : "Re-authentication after session loss failed");
    } finally {
      authStore.setAuthenticating(false);
    }
  }
```

src/Frontend/src/composables/useAuth.ts
```diff
@@ -33,12 +52,18 @@ export function useAuth() {
         }
       });
 
+      // Token fully expired, or silent renewal errored: clear the stale token and re-authenticate
+      // rather than rendering a blank app. Reacting to the OIDC events directly keeps recovery in
+      // the auth domain and distinguishes session loss from an intentional logout, which arrives
+      // as addUserUnloaded and must not re-trigger authentication.
       userManager.events.addAccessTokenExpired(() => {
         authStore.clearToken();
+        reauthenticate();
```
johnsimons:
```suggestion
userManager.events.addAccessTokenExpired(async () => {
	authStore.clearToken();
    await reauthenticate();
```

src/Frontend/src/composables/useAuth.ts
```diff
@@ -33,12 +52,18 @@ export function useAuth() {
...
       });
 
       userManager.events.addSilentRenewError((error) => {
         logger.error("Silent renew error:", error);
+        reauthenticate();
       });
```
johnsimons:
```suggestion
userManager.events.addSilentRenewError(async (error) => {
	logger.error("Silent renew error:", error);
    await reauthenticate();
});
```

PhilBastian (approved):
looks good with John's suggested changes

warwickschroeder (approved):

PR #3053, opened 2026-07-02
Allow for camelCasing or snake_casing

WilliamBZA:


src/Frontend/src/stores/AllowedRoutesStore.ts
```diff
@@ -8,9 +8,11 @@ import type RootUrls from "@/resources/RootUrls";
 
 export interface ManifestEntry {
   method: string;
-  // ServiceControl pins this field to snake_case on every instance (RouteManifestEntry uses
-  // [JsonPropertyName]). Optional at the type level because wire data is validated at runtime.
+  // Field name is inconsistent across instances: Primary serializes snake_case (url_template),
+  // Monitoring serializes camelCase (urlTemplate). Both must be accepted defensively — optional
```
ramonsmits, line 12: wow... should we fix this?

I understand the fallback, but I think all APIs on the same version should be returning the same casing style?

src/Frontend/src/stores/AllowedRoutesStore.ts
```diff
@@ -66,14 +68,15 @@ export const useAllowedRoutesStore = defineStore("AllowedRoutesStore", () => {
       const merged = new Map<string, ManifestEntry>();
       for (const list of [primary, monitoring]) {
         for (const entry of list ?? []) {
+          const urlTemplate = typeof entry.url_template === "string" ? entry.url_template : typeof entry.urlTemplate === "string" ? entry.urlTemplate : undefined;
```
ramonsmits:
```suggestion
          // Need to support both camel as snake casing
          const urlTemplate = typeof entry.url_template === "string" ? entry.url_template
                            : typeof entry.urlTemplate === "string" ? entry.urlTemplate : undefined;
```

src/Frontend/src/stores/AllowedRoutesStore.ts
WilliamBZA: As a follow up issue IMO. Every single monitoring api expects camel case so we'd have to change and test everything. I think for now, this is an appropriate fix as long as we follow through with creating a follow up.

dvdstelt (changes requested):
This was already fixed, but somehow not merged (correctly) in a branch from last week. I cherry picked that change and added it [in SC with this PR](https://github.com/Particular/ServiceControl/pull/5573). As a result we can close this.

WilliamBZA:
Withdrawing in favour of the SC change.

PR #3057, opened 2026-07-06
Disable tabs if no access

WilliamBZA:
An alternative to:

- #3056

### DIFF
--- a/src/Frontend/src/composables/apiRoutes.ts
+++ b/src/Frontend/src/composables/apiRoutes.ts
@@ -23,6 +23,7 @@ export const ApiRoutes = {
   viewEndpoints: { method: "GET", path: "/api/endpoints" }, // error:endpoints:view (EndpointsMonitoringController)
   manageThroughput: { method: "POST", path: "/api/licensing/settings/masks/update" }, // error:throughput:manage (LicensingController)
   // ---- actions ----
+  manageEndpointSettings: { method: "PATCH", path: "/api/endpointssettings/{}" },
   manageNotifications: { method: "POST", path: "/api/notifications/email" },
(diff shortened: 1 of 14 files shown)


src/Frontend/src/composables/apiRoutes.ts
```diff
@@ -23,6 +23,7 @@ export const ApiRoutes = {
   viewEndpoints: { method: "GET", path: "/api/endpoints" }, // error:endpoints:view (EndpointsMonitoringController)
   manageThroughput: { method: "POST", path: "/api/licensing/settings/masks/update" }, // error:throughput:manage (LicensingController)
   // ---- actions ----
+  manageEndpointSettings: { method: "PATCH", path: "/api/endpointssettings/{}" },
```
ramonsmits, line 26: There is also a GET API: https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Monitoring/Web/EndpointsSettingsController.cs

ramonsmits (approved):

src/Frontend/src/composables/apiRoutes.ts
WilliamBZA: I guess we can either check for GET or PATCH when doing the ability to manage endpoint settings. Is there a benefit to using the GET instead of PATCH?

WilliamBZA: If we had a separate view endpoint settings to the manage endpoint settings, then we'd need to use the GET and the PATCH. But since we only have the manage endpoint settings which includes viewing the settings, I think we're fine with using PATCH for this

dvdstelt (approved):

PR #3061, opened 2026-07-10
Add the stub for tippy to the Throughput report tests

tamararivera:
This should prevent some of the noise when running the component test for the Throughput report, as now we get a lot of warnings because this is missing.

## Reviewer Checklist


ngallegos (commented):
Should we add a global tippy stub so future tests don't have to add it?

This suggestion PR does that and removes test-specific stubs that generate duplicate warnings when the global stub is introduced:
- https://github.com/Particular/ServicePulse/pull/3064

ngallegos (approved):

tamararivera:
Now there is a lot less noise on the test output 🎉

PR #3074, opened 2026-07-16
Frontend test warnings cleanup

ngallegos:
The frontend tests have a lot of warning noise, either from untested/unmocked console output or misconfigured route mocks.  This cleans up those warnings to make the output cleaner and less confusing for identifying legitimate errors.

## Reviewer Checklist

### DIFF
--- a/src/Frontend/src/App.vue
+++ b/src/Frontend/src/App.vue
@@ -9,10 +9,12 @@ import BackendChecksNotifications from "@/components/BackendChecksNotifications.
 import { storeToRefs } from "pinia";
 import { useAuthStore } from "@/stores/AuthStore";
 import { useAllowedRoutes } from "@/composables/useAllowedRoutes";
+import { useConfigurationStore } from "@/stores/ConfigurationStore";
 
 const authStore = useAuthStore();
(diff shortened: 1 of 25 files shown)


src/Frontend/src/App.vue
```diff
@@ -33,6 +35,16 @@ const isAnonymousRoute = computed(() => route.meta?.allowAnonymous === true);
 const shouldShowApp = computed(() => !authEnabled.value || isAuthenticated.value || isAnonymousRoute.value);
 // Show full app layout (header, footer, notifications) only when authenticated or auth is disabled
 const shouldShowFullLayout = computed(() => !authEnabled.value || isAuthenticated.value);
+
+watch(
+  [loading, authEnabled, isAuthenticated],
+  ([isLoading, enabled, authenticated]) => {
+    if (!isLoading && (!enabled || authenticated)) {
+      configurationStore.ensureLoaded();
```
cquirosj: I wonder if the watch function can be async so this call can be awaited. There are other calls to this function elsewhere where I saw it not being awaited.

cquirosj (approved):

PR #3077, opened 2026-07-21
Display for license metadata

PhilBastian:
new functionality to show additional license details for customers with a supported license and system. This feature is being slowly rolled out and will not yet be available to all customers.
Documentation for this feature is pending, this issue will be updated once it's available

## Reviewer Checklist


src/Frontend/src/components/serviceControlClient.ts
```diff
@@ -37,7 +37,18 @@ class ServiceControlClient {
...
+      let error: Error = new HttpError(response.status, response.statusText ?? "No response");
+      if (response.status === 400) {
+        try {
+          const errorResponse = await response.json();
+          error = new Error(errorResponse.detail ?? errorResponse.title);
+          // eslint-disable-next-line no-empty
+        } catch {}
+      }
+      throw error;
+    }
```
mikeminutillo, line 49: This block won't be executed. If the response is not OK we already threw an HttpError on the line above.

src/Frontend/src/stores/LicenseDetailsStore.ts
```diff
@@ -0,0 +1,82 @@
...
+  const endpoints = ref<Endpoint[]>([]);
+  const infrastructureQueues = ref<Queue[]>([]);
+  const excludedQueues = ref<Queue[]>([]);
+  const serviceEndDate = ref<Date>();
+  const validId = ref<boolean>(true);
+  const hasLicenseDetails = ref(false);
+  const error = ref<string | null>();
+
+  async function refresh() {
+    if (!isLicenseDetailsSupported) return;
```
mikeminutillo: Is there any chance that license details could become unsupported while the app is running (i.e. If I update the ServiceControl instance I am pointing at to an older one while the app is running)

If so, should this clear the values?

src/Frontend/src/stores/LicenseDetailsStore.ts
```diff
@@ -0,0 +1,82 @@
...
+      const scQueues = await throughputClient.queues();
+
+      const toQueue = (metaQ: ApiQueue) =>
+        ({
+          nameHash: metaQ.name_hash,
+          scope: metaQ.scope,
+          details: scQueues.find((scQ) => scQ.name_hash === metaQ.name_hash),
+        }) as Queue;
+
+      const sortedProducts = data.products.toSorted((p1, p2) => (p1.monthly_throughput ?? Number.MAX_VALUE) - (p2.monthly_throughput ?? Number.MAX_VALUE));
```
mikeminutillo, line 53: This means that everything without throughput will be lumped together at the bottom of the list in a semi-random order (they will sort out to 0s). Do we need more definition than that or is this really an edge case?

src/Frontend/src/stores/LicenseDetailsStore.ts
```diff
@@ -0,0 +1,82 @@
...
+      const sortedEndpointSizes = sortedProducts.map((product, i) => new EndpointSize(product.product_code, i > 0 ? (sortedProducts[i - 1].monthly_throughput ?? 0) : 0, product.monthly_throughput));
```
mikeminutillo: Should there be a +1 on the min to prevent endpoints right on the line from matching multiple sizes?

src/Frontend/src/views/throughputreport/licenseDetails/EndpointDetails.vue
```diff
@@ -0,0 +1,70 @@
...
+    .map(([month, throughput]) => ({ month, throughput }) as MonthlyThroughput)
+);
+const sortedQueues = computed(() => props.endpoint.queues.toSorted((q1, q2) => q1.details?.name.toLowerCase().localeCompare(q2.details?.name.toLowerCase())));
+</script>
+
+<template>
+  <div class="card-body" :key="endpoint.clientId">
+    <div class="details">
+      <DetailsItem label="Average Throughput/Month">
+        {{ endpoint.totalMonthlyThroughput.toLocaleString() }}
```
mikeminutillo: If this is the average, should the field be called that? Total does not sound like Average.

src/Frontend/src/views/throughputreport/licenseDetails/QueueName.vue
```diff
@@ -0,0 +1,39 @@
...
+import { WarningLevel } from "@/components/WarningLevel";
+
+defineProps<{ queue: Queue }>();
+</script>
+
+<template>
+  <div class="queue-name">
+    <span>
+      <span class="scope" v-if="queue.scope != null">{{ queue.scope }}</span>
+      <span class="name" v-if="queue.details">{{ queue.details.name }}</span>
```
mikeminutillo, line 14: I wonder if it would makes sense to have a v-else with `Unknown Queue` or something similar. I know we have the exclamation mark, and maybe that's sufficient.

src/Frontend/src/views/throughputreport/licenseDetails/NoMetadata.vue
```diff
@@ -0,0 +1,34 @@
...
+</script>
+
+<template>
+  <div class="no-metadata mt-2">
+    <div v-if="metadataError" class="alert alert-danger">{{ metadataError }}</div>
+    <CapabilityCard
+      :title="Capability.LicensedEndpointTracking"
+      subtitle="Track throughput against licensed endpoints"
+      :status="CapabilityStatus.InstanceNotConfigured"
+      description="Depending on your system setup, it is possible to track current usage and maintain queues associated with licensed endpoints. This can assist when determining renewal costs, and also assists Particular with determining the correct processing of your usage report. This functionality requires an endpoint metadata file which is linked to your license and, if available for your system, can be downloaded from the customer portal to be uploaded into ServiceControl here."
```
mikeminutillo, line 16: The existing cards have a dynamic description based on the status, but in this case, are we just re-using the control and we are not adding to the list of capability cards on the dashboard?

mikeminutillo (approved):

src/Frontend/src/components/serviceControlClient.ts
PhilBastian: merge 🤦

src/Frontend/src/stores/LicenseDetailsStore.ts
PhilBastian: these are the product sizes, of which Unlimited is the only one which can be empty

src/Frontend/src/views/throughputreport/licenseDetails/QueueName.vue
PhilBastian: I think the exclaimationmark is sufficient, especially since it has a tooltip

mikeminutillo (approved):

Issue #2317, opened 2025-04-07
Add a "platform health" section


mauroservienti:
#### Describe the requested feature

ServicePulse lacks a platform health section showing details about the overall platform components' health status.  For example, it should display the following information:

- the ServiceControl instances' topology
- for each instance:
  - the status of circuit breakers
  - the status of async timers
  - disk space
  - database indexes status
  - database memory and CPU load
  - ability to download the diagnostic file
  - ability to download the instance logs
  - instance metrics

#### Describe alternatives you've considered

N/A

jpalac:
A portion of this has already been implemented by https://github.com/Particular/ServicePulse/pull/3098.
Documentation available in docs: https://docs.particular.net/servicepulse/platform-health

<img width="1189" height="522" alt="Image" src="https://github.com/user-attachments/assets/2dfa16ca-782f-4bd6-8ce8-9f8be0043603" />

<img width="1134" height="638" alt="Image" src="https://github.com/user-attachments/assets/f2cc6c43-fbb3-4ba0-980a-c6635696ce1f" />

andreasohlund:
Follow up ideas mentioned in this issue has been moved to https://github.com/Particular/ServicePulse/issues/3118. Closing this as done

PR #3098, opened 2026-08-28
Add platform health dashboard

andreasohlund:
Partially addresses https://github.com/Particular/ServicePulse/issues/2317

## Summary
- add the `Platform health` page, support export, and runtime mock controls
- share one frontend platform model across `Platform health` and `Platform Capabilities`
- hide internal platform custom checks by default and let Platform health use them as secondary health signals
- map matching failing internal checks to degraded platform instances
- refresh the Platform health and Custom Checks docs to match the new behavior

## Design Decisions
- Use a shared frontend platform model/store as the source of truth for `Platform health` and `Platform Capabilities`. `PlatformModelStore` owns normalization for primary ServiceControl, remote instances, and Monitoring, while transport stays in the purpose-specific clients.
- Keep `ServicePulse` as a frontend-owned model with only the fields needed for Platform health and update visibility.
- Derive topology behavior from shared instance roles instead of introducing another topology layer. Remote instances remain model-derived rows, and the direct-access rule stays limited to primary ServiceControl and Monitoring.
- Read the primary version from the real `X-Particular-Version` response header and the Monitoring version from its root `version` field. A successful primary root fetch still maps to a `healthy` baseline.
- Use the ServiceControl `internal` flag as the source of truth for platform custom checks. Older responses without the flag are treated as non-internal.
- Keep internal platform custom check inference page-specific in `PlatformHealthStore`. Matching is by `originating_endpoint.name`, and a failing internal check degrades the matching healthy instance.
- Keep `PlatformHealthStore` responsible for page-specific severity, rows, upgrade cues, details, and support export content, while `CustomChecksStore` only controls filtering and pagination for the list view.

## Behavior
- `Show platform custom checks` reveals internal checks in the Custom Checks UI
- Platform health matches internal checks by `originating_endpoint.name`
- matching failing internal checks degrade the associated platform instance

## Verification
- `npm run test:component -- src/stores/CustomChecksStore.spec.ts src/stores/PlatformHealthStore.spec.ts test/mocks/platform-health-state.spec.ts`
- `npm run lint`
- `npm run type-check`


src/Frontend/src/components/platformcapabilities/capabilities/AuditingCapability.ts
```diff
@@ -133,12 +134,7 @@ export function useAuditingCapability(): CapabilityComposable {
       return CapabilityStatus.PartiallyUnavailable;
     }
 
-    // 4. Check if the 'All Messages' feature is not supported OR there are no successful messages
-    if (!isAllMessagesSupported.value || !hasSuccessfulMessages.value) {
-      return CapabilityStatus.EndpointsNotConfigured;
-    }
-
-    // 5. Audit instance is available and there are successful audit messages
+    // 4. All configured audit instances are available.
```
andreasohlund, line 135: Not sure about this one, needs disussion

src/Frontend/src/components/PageFooter.vue
```diff
@@ -53,34 +37,6 @@ const { configuration } = storeToRefs(configurationStore);
               <a :href="newVersions.newSPVersion.newspversionlink" target="_blank">v{{ newVersions.newSPVersion.newspversionnumber }} available</a>)
```
andreasohlund: Needs discussion: should we move this to the health page as well?

src/Frontend/src/components/platformcapabilities/capabilities/AuditingCapability.ts
```diff
@@ -155,17 +149,6 @@ export function useAuditingCapability(): CapabilityComposable {
   const auditIndicators = computed(() => {
     const indicators: StatusIndicator[] = [];
 
-    // Add an indicator for each remote audit instance
```
andreasohlund, line 158: This is now presented on the platform health page

andreasohlund:
@rbev @jpalac this should be ready for a first review pass to catch bigger mistakes, I've tried to capture all important stuff in the PR description

src/Frontend/src/stores/PlatformHealthStore.ts
```diff
@@ -0,0 +1,337 @@
...
+
+    return `${issues.length} issue${issues.length === 1 ? "" : "s"} detected: ${issues.join(", ")}.`;
+  });
+
+  const supportDownloadJson = computed(() => {
+    if (!currentModel.value) {
+      return "";
+    }
+
+    return JSON.stringify(
```
andreasohlund, line 130: Needs discussion, more stuff could go in here in the future

jpalac (approved):

rbev (approved):

PR #3047, opened 2026-06-30
Messages view does not indicate that it is waiting for a response from ServiceControl

ngallegos:
### Who's affected
All users of ServicePulse, but more so those where HTTP requests to ServiceControl from the Messages view take longer to resolve.

### Root cause
The UI only had a loading indicator showing on the first load, not on subsequent requests from auto-refresh, searching, or filtering.

### Confirmed workarounds
None

### Additional Information

When HTTP calls to ServiceControl take longer than expected, the audit messages view does not make it clear that the user is waiting for something to finish. This introduces some UI improvements to help with that:
- The message list now shows the loading spinner with an overlay when waiting on an HTTP call
- The refresh button is linked to HTTP calls instead of becoming ready after a hardcoded 1 second
- The search and filter controls become disabled along with the refresh button while data is being fetched

## Reviewer Checklist

- [x] Components are broken down into sensible and maintainable sub-components.
- [x] Styles are scoped to the component using it. If multiple components need to share CSS, then a .css file is created containing the shared CSS and imported into component scoped style sections.
- [x] Naming is consistent with existing code, and adequately describes the component or function being introduced
- [x] Only functions utilizing Vue state or lifecycle hooks are named as composables (i.e. starting with 'use');
- [x] No module-level state is being introduced. If so, request the PR author to move the state to the corresponding Pinia store.

PR #3105, opened 2026-09-03
All Messages view no longer disabled during slow queries

ramonsmits:
Stacked on:

- #3104

Five fixes that make the All Messages view behave when queries are slow (large audit stores):

- **Actionable feedback on failure:** a failed or timed-out query showed a silent empty list (the error died as an unhandled watcher rejection). It now shows an alert explaining the instance might be too busy, with concrete suggestions. Pairs with the server-side query time limit in Particular/ServiceControl#5848.
- **Controls never lock:** the filter inputs were disabled while a query ran — causing the type-one-character-lose-focus bug and locking users out for the duration of a slow query. Entering a new query is now always possible.
- **New queries supersede in-flight ones:** starting a query aborts the one still running (the abort propagates through ServiceControl and terminates the database-side query) instead of silently dropping the new request; a superseded response can no longer overwrite newer results.
- **Auto-refresh cooperates with slow queries:** a tick landing mid-query no longer disappears and does not cancel the running query — it refreshes as soon as the results are in. A countdown bar shows when the next refresh is due.
- **Typed searches always start:** a search entered while the initial query was still running was silently ignored.
- **Existing results stay visible and usable while a re-fetch runs:** since #3047 every re-fetch covered the list with a white overlay, which flashed on every auto-refresh tick and made the rows unreachable while waiting. That overlay existed because nothing could cancel a running query at the time; with cancel and supersede in place the refresh button is the only activity indicator, and the plain spinner remains only when there is nothing to show yet.
- **Leaving the view stops its activity:** navigating away aborts the in-flight query and releases the auto-refresh registration. It also clears the results: the rows live in the store, which outlives the view, so re-entering it used to show the previous visit's rows under the first-load spinner. A refresh in place keeps stale rows; coming back with possibly different inputs starts from a clean list.

Related:

- Particular/ServiceControl#5848

### DIFF
--- a/src/Frontend/src/components/AutoRefreshIndicator.spec.ts
+++ b/src/Frontend/src/components/AutoRefreshIndicator.spec.ts
@@ -1,0 +1,28 @@
+import { describe, expect, test } from "vitest";
+import { render, screen } from "@testing-library/vue";
+import AutoRefreshIndicator from "@/components/AutoRefreshIndicator.vue";
+
+describe("FEATURE: Auto refresh indicator", () => {
+  test("EXAMPLE: Nothing is rendered while auto refresh is off", () => {
+    render(AutoRefreshIndicator, { props: { nextRefreshAt: null, intervalMs: 5000, refreshing: false } });
+
+    expect(screen.queryByTestId("auto-refresh-indicator")).not.toBeInTheDocument();
+  });
+
+  test("EXAMPLE: A countdown bar is shown while waiting for the next refresh", () => {
+    render(AutoRefreshIndicator, { props: { nextRefreshAt: Date.now() + 2500, intervalMs: 5000, refreshing: false } });
+
+    expect(screen.getByTestId("auto-refresh-indicator")).toBeInTheDocument();
+    const bar = screen.getByTestId("auto-refresh-countdown");
+    const width = parseFloat(bar.style.width);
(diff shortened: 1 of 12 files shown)


johnsimons (changes requested):
@ramonsmits I noticed that there is some extra repainting going on. It seems to redraw the whole screen.
Here is a comparison of the before (left) and after (right), using both Chrome and Safari.

https://github.com/user-attachments/assets/ea96ff7a-7207-4aa5-964f-1b134d349a76

But overall looks good, although I am not sure about the countdown progress bar. I think I would prefer a text, and quite possible that would also be better for screen reader if we do efven support them!

ramonsmits:
> @ramonsmits I noticed that there is some extra repainting going on. It seems to redraw the whole screen. Here is a comparison of the before (left) and after (right), using both Chrome and Safari.

Good find ♥️! Interesting.. that needs a bit of digging :-)


> although I am not sure about the countdown progress bar. I think I would prefer a text, and quite possible that would also be better for screen reader if we do even support them!

True, that bar isn't there anymore. It is now an animation within the refresh button. 

However, due to the stacked PRs that isn't visible here.

johnsimons (approved):
I am approving this becasue I know quite a few other changes have been done in the next PRs

mauroservienti (approved):

PR #3106, opened 2026-09-03
Date/recent time-range picker and a bounded-by-default All Messages view

ramonsmits:
Stacked on:

- #3105

Second PR in the audit query series. Makes the All Messages view **bounded by default** and gives users control over expensive queries. On a ~158M-message store the unbounded default query took 15–17 s; bounded to a recent window it returns in 1–2 s.

### Time range picker (`Sent:` chip)

- Replaces the calendar-only date range control with a compact chip in the query bar labeled for what it filters (`time_sent`)
- Bounds are text expressions: relative (`now-6h`, `now-1d/d`) or tolerant RFC 3339 (`2026-08-31Z`, offsets honored); a pasted ISO 8601 interval splits across both bounds. Timestamps are frequently copied from logs, which is why typing/pasting is first-class next to the Grafana-style presets
- Expressions stay in the URL and resolve at query time, so shared links and auto-refresh windows keep sliding
- Default is the last 6 hours; a different default can be saved per browser, or the range cleared entirely
- A timed-out query offers the next-narrower presets as one-click buttons; a query with no time filter is called out as an unbounded scan

Review feedback applied:

- **Grafana-style editor with a calendar per bound.** The editor is two columns: the absolute range on the left with From and To on their own rows and a calendar button beside each, the quick ranges listed on the right. The calendar is the browser's own date picker (no picker library; date-only, so it closes on the day click), opened on the day the field currently resolves to. The day lands in the field and the time the field already says is kept, a UTC marker included; a field without an absolute time gets the start of the day for From and its end for To. Text stays the source of truth, so relative expressions and pasted timestamps work as before
- **"to" instead of an arrow** between the bounds, on the chip, in the tooltip and on the default line. A range ending `now` is shown as its start only (`now-90m`), since "last N" is what it means
- **Old picker removed.** `DatePickerRange.vue` and `@vuepic/vue-datepicker` had no remaining users and are gone
- **Diagnostics line removed.** The "UTC … · local … · live" line under the editor is gone; the editor only gives feedback when the draft cannot be applied, and the resolved window remains as the chip tooltip

Still open from the review: the popover API for the popup and the presets as links. The wider From field no longer cuts the placeholder off.

The rest of the original scope of this PR moved into separate stacked PRs (see the stack list in the follow-up PRs).


johnsimons (changes requested):
@ramonsmits there is a lot going on here.
Would it be possible to split this PR into more manageable slices.
Example:
- The new date range picker on its own
- The history for the search
- The reorganizing of the UI
-  ....

johnsimons:
I haven't had time to look at the code yet @ramonsmits, but a few things to consider based on the screenshots.

- Could the "history" be built into the text field, so, in other words, when the user's focus is on the search text field, we have a floating panel below it that shows them the history options? I kind of understand why you want to make it a separate dropdown because it is the history for the whole combined filters, but its location at the moment feels associated with the text field.
- The date ranges using `now-2d`, I do like it. It reminds me of Grafana; it may be useful to look at the Grafana UI and get some ideas from it.
    image
    image
- I am not sold on this text, it is hard to figure out what it says
    image
- I think local vs utc times should be app wide not specific per screen
    image

ramonsmits:
> @ramonsmits there is a lot going on here. Would it be possible to split this PR into more manageable slices. Example:
> 
> * The new date range picker on its own
> * The history for the search
> * The reorganizing of the UI
> * ....


@johnsimons I understand, I'll check if I can split these.. thing is.. I already have 3 stacked PRs and this means even more stacked ones many share the same base.

ramonsmits:
> * Could the "history" be built into the text field, so, in other words, when the user's focus is on the search text field, we have a floating panel below it that shows them the history options? I kind of understand why you want to make it a separate dropdown because it is the history for the whole combined filters, but its location at the moment feels associated with the text field.

exactly this, its currently the combination. 


> * The date ranges using `now-2d`, I do like it. It reminds me of Grafana; it may be useful to look at the Grafana UI and get some ideas from it.

Guess what the inspiration was ;-)

The difference was that I wanted a way to be able to paste timestamps as these are frequently copy/pasted from logs etc.

Do you miss anything critical? Already the ability to select options and also ability to enter timestamps with zone offsets or UTC is supported.

I do like the refresh style... maybe I can tweak that a bit more.

> * I am not sold on this text, it is hard to figure out what it says

TBH that is more for diagnostics to validate if the conversion from text input to url output was happening correctly. I'll update it to only show the resolved timezone offset (utc, local or specific offset)

> * I think local vs utc times should be app wide not specific per screen

I partially agree, there might be value in having user (or maybe only local) specific settings but in the primary use case for this screen is diagnostics based on data provided from support inquiries where a timestamp is likely not always fixed. 

My assumption is that a user wants by default to show time based on their system settings on all screens. However, this screen specifically is search related to diagnostics (local) but engineers frequently want to see timestamps in UTC too for comparing as that is what servers usually are set to or in which timestamps are actually stored.

johnsimons (dismissed):
I still think we need to have a calendar picker for dates, same as what Grafana provides 
Image

src/Frontend/src/components/audit/SuperDatePicker.vue
```diff
@@ -0,0 +1,343 @@
...
+}
+
+function onOutsidePointer(event: PointerEvent) {
+  if (open.value && root.value && !root.value.contains(event.target as Node)) open.value = false;
+}
+function onKeydown(event: KeyboardEvent) {
+  if (event.key === "Escape") open.value = false;
+}
+onMounted(() => document.addEventListener("pointerdown", onOutsidePointer));
+onBeforeUnmount(() => document.removeEventListener("pointerdown", onOutsidePointer));
```
johnsimons, line 129: A lot of this code can be removed if you use the [popover api](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API), it has all this build into it 😉

src/Frontend/src/components/audit/SuperDatePicker.vue
```diff
@@ -0,0 +1,343 @@
...
+            class="bound"
+            :class="{ invalid: fromText.trim() !== '' && fromParse.error }"
+            spellcheck="false"
+            autocomplete="off"
+            aria-label="Time range start"
+            placeholder="now-6h or 2026-09-01 08:00Z"
+            @keydown.enter="apply"
+            @paste="onPaste"
+          />
+          <span class="arrow" aria-hidden="true">→</span>
```
johnsimons: This is my preference, but have you considered using the word "to" instead of an arrow?

src/Frontend/src/components/audit/SuperDatePicker.vue
```diff
@@ -0,0 +1,343 @@
...
+          <input v-model="toText" class="bound" :class="{ invalid: toText.trim() !== '' && toParse.error }" spellcheck="false" autocomplete="off" aria-label="Time range end" placeholder="now" @keydown.enter="apply" @paste="onPaste" />
+          <button type="button" class="go" :disabled="!isValid || !isDirty" @click="apply">Apply</button>
+        </div>
+        <div v-if="problem" class="echo bad" role="alert">{{ problem }}</div>
+      </div>
+
+      <div class="quick">
+        <button v-for="preset in rangePresets" :key="preset.label" type="button" class="quick-item" @click="applyPreset(preset.from, preset.to)">
+          {{ preset.label }}
+        </button>
```
johnsimons, line 251: These are styled as links. Have you considered just making these `a` elements instead

src/Frontend/src/components/audit/SuperDatePicker.vue
```diff
@@ -0,0 +1,343 @@
...
+        <div class="bar">
+          <input
+            ref="fromInput"
+            v-model="fromText"
+            class="bound"
+            :class="{ invalid: fromText.trim() !== '' && fromParse.error }"
+            spellcheck="false"
+            autocomplete="off"
+            aria-label="Time range start"
+            placeholder="now-6h or 2026-09-01 08:00Z"
```
johnsimons: This is not visible in its entirety

Image

johnsimons:
You can delete @/components/audit/DatePickerRange.vue" and also remove the @vuepic/vue-datepicker npm package

src/Frontend/src/components/audit/SuperDatePicker.vue
ramonsmits: Good suggestion, I had not used the popover API before. For the picker the dismiss code disappears, but then the popup needs positioning (top layer, centred by default, anchor positioning is not in Firefox ESR) and toggle handlers for aria-expanded, the draft reset and the focus. So about equal in lines, and the gain is the top layer, which nothing on this page needs yet. I suggest we keep the current version and switch when anchor positioning is supported everywhere. For SearchHistory it does not fit: light dismiss closes the panel when you click the search field it belongs to.

src/Frontend/src/components/audit/SuperDatePicker.vue
ramonsmits: Layout updated to fix this

src/Frontend/src/components/audit/SuperDatePicker.vue
ramonsmits: No, but IMHO button is the right type, it isn't a link.

src/Frontend/src/components/audit/SuperDatePicker.vue
ramonsmits: agree, now using TO, but also updated that when TO is NOW, and FROM is also relative to NOW to only show "Last XXX hours"

src/Frontend/src/components/audit/SuperDatePicker.vue
ramonsmits: Also, added a code comment on this decision.

mauroservienti (approved):

PR #3107, opened 2026-09-04
Report in the All Messages view incomplete scatter-gather results and server query timeouts

ramonsmits:
Top of the audit query series, stacked on:

- #3105
- #3106
- #3109
- #3110
- #3111
- #3112
- #3114

ServiceControl (Particular/ServiceControl#5848) now tells clients when a composite result is incomplete: a partial answer stays HTTP 200 with the data that did arrive plus a `X-Particular-Incomplete-Results` header naming every instance that contributed nothing (`instanceId:timeout|unavailable|error`), and a query nobody answered becomes a 504. This PR makes All Messages read both:

- A partial page keeps its data and shows a warning naming each missing instance and why: **Partial results.** No data from audit-2.internal:44444 (timed out), audit-3.internal:44444 (unreachable). The header carries ServiceControl's instance id, which is the base64-encoded API URL; the warning decodes it to host and port and keeps the full URL as a tooltip
- The warning also says what to do: the rows shown come from the instances that did answer, try again in a moment, or check the listed instances on Platform Health (linked); when one timed out it adds that a narrower time range makes the query lighter for a slow instance
- The results line presents the total as a floor while results are partial: `Showing 3 of at least 87,421,337 result(s)`
- A 504 gets a precise "exceeded the ServiceControl query time limit" message (with the existing one-click narrower-range escape hatches) instead of the generic failure text

Partial results warning in All Messages

The header parser is shared (`incompleteResults.ts`) so the conversation and saga views can adopt it next.

### DIFF
--- a/src/Frontend/src/components/audit/AuditList.vue
+++ b/src/Frontend/src/components/audit/AuditList.vue
@@ -1,7 +1,8 @@
 <script setup lang="ts">
 import { FieldNames, useAuditStore } from "@/stores/AuditStore";
 import { storeToRefs } from "pinia";
-import { useRoute, useRouter } from "vue-router";
+import { RouterLink, useRoute, useRouter } from "vue-router";
+import routeLinks from "@/router/routeLinks";
 import ResultsCount from "@/components/ResultsCount.vue";
 import FiltersPanel from "@/components/audit/FiltersPanel.vue";
 import ResultsOptions from "@/components/audit/ResultsOptions.vue";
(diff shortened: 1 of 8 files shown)


src/Frontend/src/components/audit/AuditList.vue
```diff
@@ -212,13 +225,25 @@ watch(autoRefreshValue, (newValue) => {
     </div>
     <WizardDialog v-if="showWizard" title="Getting Started with Auditing" :pages="wizardPages" @close="showWizard = false" />
     <div v-if="queryFailed && !queryInProgress" class="query-error" role="alert" data-testid="query-error">
-      <strong>The query failed or took too long and was stopped.</strong>
+      <strong v-if="queryTimedOut">The query exceeded the ServiceControl query time limit and was stopped.</strong>
```
mauroservienti, line 228: Is there a quick way to see what the configured timeout value is on the backend?

mauroservienti (approved):

PR #3111, opened 2026-09-07
List recent searches under the All Messages search field

ramonsmits:
Stacked on:

- #3105
- #3106
- #3109
- #3110

Per-browser history of audit searches with one-click rerun.

- Any query that used search text or an endpoint is remembered in this browser (most recent first, capped at 10, no duplicates), together with the time-range expressions it ran with. The range is part of a search's identity, and rerunning restores it, so an entry is a true "run this exact query again"
- Review feedback applied: instead of a separate "History" dropdown, the recent searches float **under the search field while it has focus**, the way a browser address bar does. Typing narrows the list to entries whose search text or endpoint match; Escape dismisses it; a click in the field brings it back. An empty history shows nothing

### DIFF
--- a/src/Frontend/src/components/audit/SearchHistory.vue
+++ b/src/Frontend/src/components/audit/SearchHistory.vue
@@ -1,0 +1,209 @@
+<script setup lang="ts">
+import { computed, onBeforeUnmount, onMounted, ref, useTemplateRef } from "vue";
+import { storeToRefs } from "pinia";
+import { useAuditStore } from "@/stores/AuditStore";
+import type { SearchHistoryEntry } from "@/components/audit/searchHistory";
+import TimeSince from "@/components/TimeSince.vue";
+import { describeRangeText } from "@/components/audit/timeRange";
+
+// Wraps the search field (default slot) and floats the recent searches below it while
+// the field has focus, the way a browser's address bar does. Entries carry the whole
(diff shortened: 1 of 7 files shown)


src/Frontend/src/components/audit/SearchHistory.vue
```diff
@@ -0,0 +1,204 @@
...
+}
+
+function onOutsidePointer(event: PointerEvent) {
+  if (open.value && root.value && !root.value.contains(event.target as Node)) open.value = false;
+}
+function onKeydown(event: KeyboardEvent) {
+  if (event.key === "Escape") open.value = false;
+}
+onMounted(() => document.addEventListener("pointerdown", onOutsidePointer));
+onBeforeUnmount(() => document.removeEventListener("pointerdown", onOutsidePointer));
```
johnsimons, line 132: Again, please consider using https://developer.mozilla.org/en-US/docs/Web/API/Popover_API

johnsimons (approved):

src/Frontend/src/components/audit/SearchHistory.vue
ramonsmits: Same reasoning as on the picker thread (#3106), plus the combobox problem: light dismiss would close the panel on a click in the search field.

mauroservienti (approved):

PR #3112, opened 2026-09-07
Use a relative timestamp for the All Messages view Time sent column

ramonsmits:
Stacked on:

- #3105
- #3106
- #3109
- #3110
- #3111

Timestamp presentation in the All Messages list.

- **Age-adaptive Time Sent**: time only for today, `yesterday 14:32`, weekday names within the past week, the regular browser format beyond, with a live relative label (`3 minutes ago`) and the full timestamp in both zones as tooltip
- **`Times: local|UTC` toggle** on the results line flips every timestamp to UTC wall-clock time, persisted per browser. Review note: an app-wide setting was suggested; that is deliberately left for a separate issue, this PR keeps the per-view toggle
- **`ran 3 minutes ago`** on the results line says when the current results were fetched
- **Aligned row columns**: the columns are defined once on the list and shared by every row via subgrid, so nothing wraps mid-label and rows lay out identically

Review feedback applied: the age no longer has a column of its own. It follows the timestamp it belongs to, muted, as the last cell of the metrics line: `Time Sent: 8:57 AM · 4 minutes ago`, left aligned like the id and the metrics next to it, so the age is explained by the label it sits next to. Time Sent stays on the metrics line with the other data cells. The subgrid stays for the alignment; the age is hidden on narrow screens.

### DIFF
--- a/src/Frontend/src/components/audit/AuditList.vue
+++ b/src/Frontend/src/components/audit/AuditList.vue
@@ -313,5 +313,30 @@ watch(autoRefreshValue, (newValue) => {
   margin-bottom: 5rem;
   background-color: #ffffff;
   position: relative;
+  /* The results list is a grid that places nothing itself: it only declares the six
+     columns, and every row (AuditListItem) joins them with `subgrid`. Declaring them here,
+     once, is what keeps rows aligned: a column is measured across ALL rows, which a row
+     laying out its own columns cannot do.
+
+       1.8em                       status icon
+       minmax(0, 1fr)              message id: the one value allowed to shrink and break,
+                                   so a long id never pushes the values off the row
+       minmax(max-content, 1fr)    each value column, read as: never narrower than its
+                                   widest value in the list (so nothing wraps), and once
+                                   every column has that, share the leftover width equally
+                                   (so the columns spread out on a wide screen instead of
+                                   huddling on the left)
+
+     Deliberately NOT a size container: combining container-type with content-sized
+     tracks froze Chrome's layout. */
+  display: grid;
+  grid-template-columns: 1.8em minmax(0, 1fr) repeat(4, minmax(max-content, 1fr));
+  column-gap: 0.375rem;
+  align-content: start;
+}
+
+/* Non-row children (the first-load spinner) span the full width */
+.results-table > :not(.item) {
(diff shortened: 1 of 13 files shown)


src/Frontend/src/components/audit/AuditList.vue
```diff
@@ -277,5 +277,20 @@ watch(autoRefreshValue, (newValue) => {
...
+  /* One column definition for all rows (rows join it via subgrid). Data
+     columns never shrink below their widest value in the list (no wrapping)
+     but share surplus width equally, so they stretch across the row instead
+     of piling up on the right on wide screens. Deliberately NOT a size
+     container: combining container-type with content-sized tracks froze
+     Chrome's layout. */
+  display: grid;
+  grid-template-columns: 1.8em minmax(0, 1fr) minmax(max-content, 1fr) minmax(max-content, 1fr) minmax(max-content, 1fr) minmax(max-content, 1fr) max-content;
+  column-gap: 0.375rem;
+  align-content: start;
```
johnsimons: These changes are quite confusing. I am especially unsure about the `grid-template-columns:` something does not look right.
We also introduce subgrids in the AuditListItem.
Can you explain why we need to do all this?

src/Frontend/src/components/audit/AuditListItem.vue
```diff
@@ -48,12 +50,17 @@ const link = computed(() => {
...
+     metrics and sent time with its age — sits on the bottom line. */
+  grid-column: 1 / -1;
+  grid-template-columns: subgrid;
+  grid-template-rows: auto auto;
   gap: 0.375rem;
   grid-template-areas:
-    "status message-type message-type message-type time-sent"
-    "status message-id processing-time critical-time delivery-time";
+    "status message-type message-type message-type message-type message-type message-type"
+    "status message-id processing-time critical-time delivery-time time-sent age";
```
johnsimons: What is the motivation for changing the order of these? Customers may be used to it!

mauroservienti (approved):

src/Frontend/src/components/audit/AuditListItem.vue
ramonsmits: Reverted, timesent timestamp + age now combined.

PR #3117, opened 2026-09-07
🐛 Stop the refresh button from jumping when a query is in progress

johnsimons:
## Reviewer Checklist


ramonsmits:
This is addressed in:

- https://github.com/Particular/ServicePulse/pull/3110

PR #3122, opened 2026-09-09
Expand one Platform Health row at a time when instances share a name

ramonsmits:
### Symptoms

On Platform Health, with several audit instances that report the same instance name, expanding the health badge of one instance expands all of them at once. The browser console also shows Vue's duplicate-key warning for the rows.

### Who's affected

Anyone running scaled-out audit instances that share an instance name, which default installs commonly do.

### Root cause

The view keyed the expanded row, the `v-for` and the details row `id` on `type-name`. Same-named instances shared one key, so "is this row expanded" was true for all of them.

### Fix

A row now carries the instance's own `id` from the platform model and the view keys on that. A view test with two same-named degraded audit instances expands one and asserts the other stays collapsed.

## Reviewer Checklist

### DIFF
--- a/src/Frontend/src/views/PlatformHealthView.spec.ts
+++ b/src/Frontend/src/views/PlatformHealthView.spec.ts
@@ -172,7 +172,10 @@ describe("PlatformHealthView", () => {
 
     const rows = vi.spyOn(store, "rows", "get").mockReturnValue([
       {
+        id: "row-1",
+
         type: "Error instance",
+
         name: "Particular.ServiceControl",
         version: "6.19.3",
         health: "healthy",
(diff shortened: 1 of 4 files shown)


src/Frontend/src/views/PlatformHealthView.spec.ts
```diff
@@ -172,7 +172,10 @@ describe("PlatformHealthView", () => {
 
     const rows = vi.spyOn(store, "rows", "get").mockReturnValue([
       {
+        id: "row-1",
+
```
andreasohlund: nitpick: why the newlines?

andreasohlund (approved):

src/Frontend/src/views/PlatformHealthView.spec.ts
ramonsmits: no reason, fixed

